PR #4, opened 2022-02-15
Create actions to manually deploy to staging and preview environments

clementprdhomme:
In order to see in Jira which tasks have been deployed and to which environment, we need to create [GitHub deployments](https://github.com/atlassian/github-for-jira/blob/main/docs/deployments.md). The main goal of this PR is then to create such deployments so they can appear on Jira.

Unfortunately, Vercel doesn't create deployments on GitHub. For this reason, the [Vercel Action](https://github.com/marketplace/actions/vercel-action) action is used to manually deploy to Vercel. It will deploy to [our staging](https://heco-invest-frontend.vercel.app/) when commits are pushed to the `develop` branch, and deploy to preview environments when pull requests are created.

In addition to this, the front-end tests are run when changes are made to the `frontend/` folder, and the deployments are not triggered if changes have not been made in this folder.

The `Staging deployment` workflow could also be used for the back-end deployment in the future.

## Testing instructions

Review the workflows.

## Tracking

Not tracked.

### DIFF
--- a/.github/workflows/pull-request-deployment.yml
+++ b/.github/workflows/pull-request-deployment.yml
@@ -1,0 +1,48 @@
+name: Preview deployment
+on:
+  pull_request:
+    paths:
+      - 'front-end/**'
+
+defaults:
+  run:
+    working-directory: frontend
+
+jobs:
+  deploy:
+    name: Deploy to preview
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Create deployment
+      uses: bobheadxi/deployments@v0.6.2
+      id: deployment
+      with:
+        step: start
+        token: ${{ secrets.GITHUB_TOKEN }}
+        env: preview
+
+    - name: Run front-end tests
+      run: yarn test
+
+    - name: Front-end deployment
+      if: vercel-deployment
+      uses: amondnet/vercel-action@v20
+      with:
+        vercel-token: ${{ secrets.VERCEL_TOKEN }}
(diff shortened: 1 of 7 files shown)


.github/workflows/pull-request-deployment.yml
```diff
@@ -0,0 +1,48 @@
+name: Preview deployment
+on:
+  pull_request:
+    paths:
+      - 'front-end/**'
```
tsubik, line 5: typo here, also I would add a path to this workflow file

.github/workflows/staging-deployment.yml
```diff
@@ -0,0 +1,58 @@
+name: Staging deployment
```
tsubik, line 1: I think we can have different worflows for backend and frontend, as with this setup we have to add backend deploy here as well?

tsubik: I don't know if it is still not possible to use output from one workflow in another or maybe something has changed, but if not, maybe it would be easier to just have one workflow `frontend.yml` which will do `tests` as one job and then based on branch, deploy job to either staging or preview. Here is example with backend deploy https://github.com/Vizzuality/tota/blob/main/.github/workflows/backend.yml#L70-L74

clementprdhomme: It seems to be possible: https://docs.github.com/en/actions/using-workflows/reusing-workflows

I'll try to make the tests reusable! Thanks for the feedback.

tsubik: Nice that they added that! :)

PR #6, opened 2022-02-17
Chore/change husky

tsubik:
~Use lint-staged with husky to not invoke frontend pre-commit checks when working on backend and vice-versa. Husky advises to have it installed on the root folder, the same for lint-staged (https://github.com/okonet/lint-staged#how-to-use-lint-staged-in-a-multi-package-monorepo).~

Switched to https://github.com/evilmartians/lefthook as it seems to be easier to configure and extend with different hooks. There is still problem with Typescript type check that cannot easily run on staged files, instead it runs on the whole project which takes a lot of time (and that will grow, maybe it should be pre_push hook, not pre_commit?).

## Testing instructions

Make sure to uninstall previously installed husky config `git config --unset core.hooksPath`. Then run `yarn install` at root directory. Do any dummy change that would invoke errors either on the backend or frontend side (or both) and try to add commit.

## Tracking

Not tracked


tsubik:
I think I found a way better and faster hook manager https://github.com/evilmartians/lefthook I will test it and update PR @clementprdhomme

lefthook.yml
```diff
@@ -0,0 +1,18 @@
...
+    frontend-lint:
+      tags: frontend lint
+      root: "frontend/"
+      glob: "*.{js,jsx,ts,tsx}"
+      run: eslint --fix {staged_files}
+    frontend-types:
+      tags: frontend types
+      root: "frontend/"
+      glob: "**/*.{ts,tsx}"
+      run: echo {staged_files} >/dev/null && yarn check-types
```
tsubik, line 13: lefthook is not without issues, like here {staged_files} must appear otherwise it will always do frontend types check even if there are no staged files matching `root` and `glob`. https://github.com/evilmartians/lefthook/issues/228#issuecomment-924505331

tsubik:
Updated @clementprdhomme

lefthook.yml
tsubik: Other solution is to use https://github.com/gustavopch/tsc-files instead of `tsc`.

tsubik: or moving type check to pre-push

lefthook.yml
```diff
@@ -0,0 +1,18 @@
+pre-commit:
+  parallel: true
+  commands:
+    frontend-lint:
+      tags: frontend lint
+      root: "frontend/"
+      glob: "*.{js,jsx,ts,tsx}"
+      run: eslint --fix {staged_files}
```
clementprdhomme, line 8: Could you use `yarn lint --fix` instead? I believe we can pass the list of files using the `--file` option. You can check the docs by running:
```bash
npx next lint -h
```

lefthook.yml
clementprdhomme: I think that'd be a good idea to move type checking to a pre-push hook instead.

I'm in favour of having every commit type checked but if this takes more than 1 or 2 seconds to execute, maybe we can only check when pushing to remote.

clementprdhomme (commented):
In addition to my comments, I realised you forgot to delete the `.husky` folder in `/frontend`.

tsubik:
`.husky` folder was deleted, at least I don't have it.

PR #11, opened 2022-02-25
[Backend] Setup API base controller

tsubik:
Backend base API controller.

The controller is set up to use cookie authentication. I added the need of CSRF token that will be send to frontend as a cookie and should be sent back as `X-CSRF-TOKEN` (inspired by this https://blog.eq8.eu/article/rails-api-authentication-with-spa-csrf-tokens.html or this https://pragmaticstudio.com/tutorials/rails-session-cookies-for-api-authentication). Also, all requests will be required to be `application/json` requests, from what I know that should prevent any CSRF, but just in case we will also need CSRF token.

## Testing instructions

No need.

## Tracking

https://vizzuality.atlassian.net/browse/LET-119

### DIFF
--- a/backend/Gemfile
+++ b/backend/Gemfile
@@ -14,6 +14,9 @@ gem "sprockets-rails"
 gem "stimulus-rails"
 gem "turbo-rails"
 
+# API
+gem "jsonapi-serializer"
+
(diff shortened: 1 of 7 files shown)


agnessa (approved):
Nice, thanks!

PR #12, opened 2022-02-28
Update the static translation pipeline

clementprdhomme:
This PR updates the static translation pipeline because the previous that was using Transifex Native wouldn't cover our needs for translated server-side rendered content.

In this new pipeline [FormatJS](https://formatjs.io/) is used to gather the source strings, and a git hook and GitHub action are used to push them and pull the translations from Transifex, which is now exclusively used as a database.

The whole pipeline is described in the README file below.

## Testing instructions

- Make sure the application runs locally in dev and production mode.
- Make sure the strings correctly appear in the application.
- Make sure the GitHub action triggered by this PR succeeded

What can only be tested when merged to `develop`:
- The GitHub action pushing the source strings

## Tracking

- [LET-146](https://vizzuality.atlassian.net/browse/LET-146).
- [LET-150](https://vizzuality.atlassian.net/browse/LET-150).


SARodrigues (approved):
Whoa!  

I have a question: what's the best way you found to generate the `id`'s? I noticed ESLint can do it (even on save), is that what you're relying on, or did you find a way for VSCode to insert them as we type?

clementprdhomme:
@SARodrigues Thanks for the review.

So FormatJS comes with [some plugins](https://formatjs.io/docs/guides/bundler-plugins) that should generate the IDs: one for babel and a typescript loader. There's also [one for swc](https://formatjs.io/docs/tooling/swc-plugin), the new compiler in Next.js, but Next.js doesn't support SWC plugins yet…

On the [discussions of Next.js](https://github.com/vercel/next.js/discussions/30174#discussioncomment-1922431), I found that we could go around this by telling ESLint to generate the IDs for us. That's what I did in this PR.

But now that you've asked, I've realised I haven't tested the TS loader 🤭 and the FormatJS' documentation recommends not writing IDs directly in the code. Let me test with the loader, and maybe I can remove the ESLint rule 😅.

SARodrigues:
@clementprdhomme Thanks for explaining!  

> the FormatJS' documentation recommends not writing IDs directly in the code

I think we have to figure out this better, because it may be confusing for new people joining the project. Due to the ESlint rule, VSCode will automatically show an error due to the missing `id` which may prompt someone to "fix it". Even if they know, I believe some people are using the "Auto fix on save" feature, that'd auto insert the `id`s on file save.

clementprdhomme:
@SARodrigues Unfortunately, TS loader doesn't play nice with Next.js' bundling process.

I agree it's confusing if the developer doesn't have “auto fix on save”. In no way that would be a solution, but I can update the “Quick start” section of the README file to specifically recommend toggling it on.

You may not be aware of this but ESLint is run automatically to fix issues when you create a commit (look at `/lefthook.yml`). If the dev forgets to add the ids, they will be added automatically before committing.

If you think that's still not good enough, we could [temporarily re-enable Babel](https://nextjs.org/docs/advanced-features/compiler#unsupported-features) on all the `.tsx?` files. I'd be ok with this as well.

SARodrigues:
@clementprdhomme I think a mention in the README would suffice, as I'd imagine (or hope) developers do read it.

PR #15, opened 2022-03-03
Add `react-intl` support to Storybook

SARodrigues:
This PR adds [react-intl](https://www.npmjs.com/package/react-intl) support to Storybook by means of the [storybook-react-intl](https://storybook.js.org/addons/storybook-react-intl) addon, as well as its configuration. 

In addition, I've added support for [next/image](https://nextjs.org/docs/api-reference/next/image), by means of the [storybook-addon-next](https://storybook.js.org/addons/storybook-addon-next) addon. 

## Testing instructions

- Make sure that Storybook runs correctly  

## Tracking

[LET-156](https://vizzuality.atlassian.net/browse/LET-156)

### DIFF
--- a/frontend/.storybook/main.js
+++ b/frontend/.storybook/main.js
@@ -10,7 +10,7 @@ module.exports = {
     '../containers/**/*.stories.@(js|jsx|ts|tsx)',
   ],
   staticDirs: ['../public'],
-  addons: ['@storybook/addon-links', '@storybook/addon-essentials', {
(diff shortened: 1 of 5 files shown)


clementprdhomme (approved):
Works well 👍️!

I found two issues unrelated to your changes:
- Storybook crashes if locales are missing in `lang/compiled` (should be fixed in #14)
- Storybook wouldn't compile without adding this dependency: `"@deck.gl/extensions": "^8.4.16"`. I have the same issue in `develop` as well. Would you mind adding it?

PR #16, opened 2022-03-03
[Backend] Open calls endpoint

tsubik:
Open calls endpoints

- api/v1/open_calls/:slug
- api/v1/open_calls/:uuid
- api/v1/open_calls

To create open call slug, I'm using account name and open call name. The slug is done from open call original name (base language).

```
{
  "data": {
    "id": "0436adc8-9ec9-422c-9486-a44ce021eedf",
    "type": "open_call",
    "attributes": {
      "name": "Open call 1",
      "slug": "kutch-spencer-open-call-1",
      "description": "Voluptatem blanditiis autem. Iste voluptates rerum. Animi tempore iure. Inventore facere ad.",
      "ticket_size": "scaling",
      "instrument_type": "loan",
      "sdgs": [
        1,
        4,
        5
      ],
      "money_distribution": "Et dolorem assumenda. Praesentium magni totam. Dolorem minima et. Incidunt omnis error.",
      "impact_description": "Voluptas molestiae est. Omnis asperiores quia. Dolorem autem corrupti. Totam nemo laborum.",
      "closing_at": "2023-01-03T15:10:20.792Z"
    },
    "relationships": {
      "investor": {
        "data": {
          "id": "6618f883-0d83-4b35-a476-9ca3dcc669bb",
          "type": "investor"
        }
      }
    }
  }
}
```

## Testing instructions

Test endpoints. Also, the full spec suite has been done, please review it.

## Tracking

https://vizzuality.atlassian.net/browse/LET-110
https://vizzuality.atlassian.net/browse/LET-117

### DIFF
--- a/backend/app/controllers/api/v1/open_calls_controller.rb
+++ b/backend/app/controllers/api/v1/open_calls_controller.rb
@@ -3,6 +3,8 @@ module API
     class OpenCallsController < BaseController
       include API::Pagination
 
+      before_action :fetch_open_call, only: [:show]
+
       def index
         open_calls = OpenCall.all.includes(:investor)
         pagy_object, open_calls = pagy(open_calls, page: current_page, items: per_page)
@@ -14,15 +16,13 @@ module API
       end
 
       def show
-        open_call = fetch_open_call
-
-        render json: OpenCallSerializer.new(open_call).serializable_hash
+        render json: OpenCallSerializer.new(@open_call).serializable_hash
       end
 
       private
 
       def fetch_open_call
-        OpenCall.friendly.find(params[:id])
+        @open_call = OpenCall.friendly.find(params[:id])
       end
     end
   end
--- a/backend/app/controllers/api/v1/open_calls_controller.rb
+++ b/backend/app/controllers/api/v1/open_calls_controller.rb
@@ -1,0 +1,29 @@
+module API
+  module V1
+    class OpenCallsController < BaseController
+      include API::Pagination
+
+      def index
+        open_calls = OpenCall.all.includes(:investor)
+        pagy_object, open_calls = pagy(open_calls, page: current_page, items: per_page)
+        render json: OpenCallSerializer.new(
+          open_calls,
+          links: pagination_links(:api_v1_open_calls_path, pagy_object),
+          meta: pagination_meta(pagy_object)
+        ).serializable_hash
+      end
+
+      def show
+        open_call = fetch_open_call
+
+        render json: OpenCallSerializer.new(open_call).serializable_hash
+      end
+
+      private
+
+      def fetch_open_call
+        OpenCall.friendly.find(params[:id])
+      end
+    end
+  end
+end
--- a/backend/app/models/open_call.rb
+++ b/backend/app/models/open_call.rb
@@ -4,11 +4,12 @@ class OpenCall < ApplicationRecord
 
   belongs_to :investor
 
-  validates :instrument_type, inclusion: {in: InstrumentType::TYPES, allow_blank: true}, presence: true
-  validates :ticket_size, inclusion: {in: TicketSize::TYPES, allow_blank: true}, presence: true
+  validates :instrument_type, inclusion: {in: InstrumentType::TYPES}
+  validates :ticket_size, inclusion: {in: TicketSize::TYPES}
   validates :sdgs, array_inclusion: {in: Sdg::TYPES}
+  validates :language, inclusion: {in: I18n.available_locales.map(&:to_s)}
 
-  validates_presence_of :closing_at, :language
+  validates_presence_of :closing_at
 
   translates :name, :description, :money_distribution, :impact_description
 
@@ -18,5 +19,8 @@ class OpenCall < ApplicationRecord
 
   def original_name
     I18n.with_locale(language) { name }
+    # to prevent crashing validations because of setting slug, fallback just to current locale name
+  rescue I18n::InvalidLocale
+    name
   end
 end
--- a/backend/app/models/open_call.rb
+++ b/backend/app/models/open_call.rb
@@ -1,0 +1,22 @@
+class OpenCall < ApplicationRecord
+  extend FriendlyId
+  friendly_id :investor_prefixed_name, use: :slugged
+
+  belongs_to :investor
+
+  validates :instrument_type, inclusion: {in: InstrumentType::TYPES, allow_blank: true}, presence: true
+  validates :ticket_size, inclusion: {in: TicketSize::TYPES, allow_blank: true}, presence: true
+  validates :sdgs, array_inclusion: {in: Sdg::TYPES}
+
+  validates_presence_of :closing_at, :language
+
+  translates :name, :description, :money_distribution, :impact_description
+
+  def investor_prefixed_name
+    "#{investor&.name} #{original_name}"
+  end
+
+  def original_name
+    I18n.with_locale(language) { name }
+  end
+end
--- a/backend/db/migrate/20220303132723_create_open_calls.rb
+++ b/backend/db/migrate/20220303132723_create_open_calls.rb
@@ -1,0 +1,38 @@
+class CreateOpenCalls < ActiveRecord::Migration[7.0]
+  def change
+    create_table :open_calls, id: :uuid do |t|
+      t.belongs_to :investor, foreign_key: {on_delete: :cascade}, type: :uuid, null: false, index: true
+
+      t.text :name_en
+      t.text :name_es
+      t.text :name_pt
+      t.text :slug
+      t.text :description_en
+      t.text :description_es
+      t.text :description_pt
+      t.text :money_distribution_en
+      t.text :money_distribution_es
+      t.text :money_distribution_pt
+      t.text :impact_description_en
+      t.text :impact_description_es
+      t.text :impact_description_pt
+
+      t.boolean :trusted, null: false, default: false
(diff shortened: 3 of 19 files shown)


backend/app/controllers/api/v1/open_calls_controller.rb
```diff
@@ -0,0 +1,29 @@
+module API
+  module V1
+    class OpenCallsController < BaseController
+      include API::Pagination
+
+      def index
+        open_calls = OpenCall.all.includes(:investor)
```
martintomas, line 7: I see you don't use any gem for authorization -- recommend to check for example cancancan. Super easy to add at beginning of project almost seamlessly (no extra work) and it is super easy to define rights later on if you ever need them

backend/app/controllers/api/v1/open_calls_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+        pagy_object, open_calls = pagy(open_calls, page: current_page, items: per_page)
+        render json: OpenCallSerializer.new(
+          open_calls,
+          links: pagination_links(:api_v1_open_calls_path, pagy_object),
+          meta: pagination_meta(pagy_object)
+        ).serializable_hash
+      end
+
+      def show
+        open_call = fetch_open_call
```
martintomas, line 17: I would prefer `before_action` maybe here

backend/app/models/open_call.rb
```diff
@@ -0,0 +1,22 @@
+class OpenCall < ApplicationRecord
+  extend FriendlyId
+  friendly_id :investor_prefixed_name, use: :slugged
+
+  belongs_to :investor
+
+  validates :instrument_type, inclusion: {in: InstrumentType::TYPES, allow_blank: true}, presence: true
```
martintomas, line 7: Yeah, to keep integrity with `InstrumentType::TYPES` is gonna be slight headache -- no big deal, but if you would have table for `InstrumentType::TYPES`, you would get this for free. Sry, had to mention it :D

backend/app/models/open_call.rb
```diff
@@ -0,0 +1,22 @@
...
+  validates :ticket_size, inclusion: {in: TicketSize::TYPES, allow_blank: true}, presence: true
+  validates :sdgs, array_inclusion: {in: Sdg::TYPES}
+
+  validates_presence_of :closing_at, :language
+
+  translates :name, :description, :money_distribution, :impact_description
```
martintomas, line 13: First time to use Traco and even though I like implementation part of gem, to have column per language at same table is kind of violation of data normalization.

backend/app/models/open_call.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def investor_prefixed_name
+    "#{investor&.name} #{original_name}"
```
martintomas, line 16: And Friendly ID is also kind of new to me :D. Do I understand it correctly that `investor_prefixed_name ` needs to be uniq? If so, It feels a bit tricky to have uniq (slug) for `original_name ` which is localised. Even more, language is just string without any validation so you can put there whatever you want :)

backend/db/migrate/20220303132723_create_open_calls.rb
```diff
@@ -0,0 +1,38 @@
+class CreateOpenCalls < ActiveRecord::Migration[7.0]
+  def change
+    create_table :open_calls, id: :uuid do |t|
+      t.belongs_to :investor, foreign_key: {on_delete: :cascade}, type: :uuid, null: false, index: true
```
martintomas, line 4: Hmm, used to slightly different approach:
 - at db, you just define `foreign_key: true`
 - what should happens with related record is defined on Rails level `has_many :open_calls, dependent: :destroy`
 
Same applies for `null: false` <-- this is by default checked by Rails via `belongs_to :investor, optional: ?`
And I believe that `index: true` is default, no need to write it

backend/db/migrate/20220303132723_create_open_calls.rb
```diff
@@ -0,0 +1,38 @@
...
+      t.text :slug
+      t.text :description_en
+      t.text :description_es
+      t.text :description_pt
+      t.text :money_distribution_en
+      t.text :money_distribution_es
+      t.text :money_distribution_pt
+      t.text :impact_description_en
+      t.text :impact_description_es
+      t.text :impact_description_pt
```
martintomas, line 18: yeah, this is scary, all these columns with language mutations :D

backend/db/migrate/20220303132723_create_open_calls.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      t.boolean :trusted, null: false, default: false
+
+      t.string :ticket_size, null: false
+      t.string :instrument_type, null: false
```
martintomas, line 23: :) super simple to break integrity of these records when you start manipulating with TYPES -- but again, I know that you don't except anybody to change TYPES, so you can definitely ignore this ;)

backend/app/controllers/api/v1/open_calls_controller.rb
```diff
@@ -0,0 +1,29 @@
+module API
+  module V1
+    class OpenCallsController < BaseController
+      include API::Pagination
```
martintomas, line 4: by the way, forgot to mention, but I love you pagination concern, really cool

martintomas (approved):
Had some spare time so really briefly checked PR :). It is really good, just a few notes -- basically I am getting familiar with traco or friendly_id so maybe some of my comments originates from there.

backend/app/controllers/api/v1/open_calls_controller.rb
tsubik: Not a fan on cancancan, more a pundit person ;] but that's a decision for later.

backend/app/controllers/api/v1/open_calls_controller.rb
tsubik: If there are more actions then yes.

backend/db/migrate/20220303132723_create_open_calls.rb
tsubik: All of those are closed lists which will have descriptions etc. that will be pushed to transifex at some point. We need to easy way to push/pull translations for that and Rails locale files seem like a good choice.

backend/app/models/open_call.rb
tsubik: Everything has pros and cons.

backend/app/controllers/api/v1/open_calls_controller.rb
martintomas: Haha :), more cancancan person than pundit -- you have one simple file for all rights even for bigger projects with cancancan, but you have soo many files with rights with pundit

backend/db/migrate/20220303132723_create_open_calls.rb
agnessa: Never depend on application level validations only. The database has performant and standardised mechanisms to protect integrity of your data. Databases tend to live longer than applications.

backend/app/models/open_call.rb
tsubik: We used globalize for example in other projects. Huge pain in the ass with all of those joins and complex queries.

backend/app/controllers/api/v1/open_calls_controller.rb
martintomas: sure, no problem. From my experience use case is often that you start with API controller which has 1 or 2 actions and it often grows later on, so it is better to take into account that it will grow later on then in middle to switch to `before_action` and refactor already existing code ;)

backend/app/models/open_call.rb
tsubik: Well, I guess I just forgot to add validation for language. I will do.

backend/app/controllers/api/v1/open_calls_controller.rb
tsubik: Well, yeah, I know we will have create/update actions as well, so I can move it. I just was too lazy to do it now.

backend/app/models/open_call.rb
martintomas: Yup, I see your point ;). Liked flexibility of joins (super easy to add new language without migrations) and had a few (probably just two) simple methods which heavy lifted all complexity. On other hand, I understand that it will be probably slightly slower -- would love to see some comparison, because foreigner keys are indexed and tables are small.
 - with joins, you get one row with all data per language at db --> and do filtering at db which language you are interested in
 - without joins, you load all data for all languages to app and do filtering there

Whatever, your points are absolutely valid ;)

tsubik:
Thanks for the feedback @martintomas I'm going to post a couple updates based on it.

backend/db/migrate/20220303132723_create_open_calls.rb
martintomas: Valid point @agnessa , thanks ;). I would mention that Rails defines often a lot of hooks on app level, therefore if you delete cascade relations on db level, you avoid these hooks completely. But you are right with your points

backend/db/migrate/20220303132723_create_open_calls.rb
martintomas: Thanks @tsubik , I know that you have already mentioned this at email. Will not mentioned this again, don't worry :)

agnessa (approved):

PR #17, opened 2022-03-04
[Backend] API sparse fieldset

tsubik:
Allow for sparse fieldset (very basic for now)

For example `api/v1/investors?fields[investor]=instagram,facebook`

We can use this for lists, map endpoints for which we don't need all the attributes.

## Testing instructions

Try API investors endpoint with different fields

## Tracking

No tracking

### DIFF
--- a/backend/app/controllers/concerns/api/sparse_fieldset.rb
+++ b/backend/app/controllers/concerns/api/sparse_fieldset.rb
@@ -1,0 +1,7 @@
+module API
+  module SparseFieldset
+    def sparse_fieldset
+      (params[:fields]&.to_unsafe_h || {}).transform_values { |v| v.split(",") }
+    end
+  end
+end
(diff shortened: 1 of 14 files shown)


martintomas (approved):
:) looks great

backend/app/controllers/concerns/api/sparse_fieldset.rb
```diff
@@ -0,0 +1,7 @@
+module API
+  module SparseFieldset
+    def sparse_fieldset
+      (params[:fields]&.to_unsafe_h || {}).transform_values { |v| v.split(",") }
```
martintomas, line 4: surprisingly rubocop does not complain when using `""` instead of `''` so all good :)

tsubik: We are not using `''`. Just use `""` everywhere. For this project, we are trying to use https://github.com/testdouble/standard

martintomas: Aha, perfect, thanks @tsubik , will check it ;)

agnessa:
I applied this to open calls & project developers so it's all symmetrical

agnessa (approved):

PR #22, opened 2022-03-09
Dependabot configuration

agnessa:
Hey, is this what the dependabot config is supposed to be? I don't have a working example

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

## Testing instructions

really no clue

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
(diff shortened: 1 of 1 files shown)


tsubik:
Yeah, me neither. I thought it is always configurable from Security tab in repo. I don't see it here, but probably I'm not just owner.

image

Frontend does not have it configured?

agnessa:
Yes, we have it enabled in the security tab. But because I haven't seen any backend alerts so far, I started thinking maybe we're missing this config to point to the right directory. I'm keen to merge this and see what happens!

tsubik:
Ok. Yeah, basically, so far I was trying to do `bundle update` from time to time, and if there was a security issue then backend security task was catching that quickly. As we work on the project daily, then Dependabot didn't have a chance to shine :D

tsubik (approved):

PR #23, opened 2022-03-11
Feature/backend transifex

tsubik:
Transifex configuration for the backend. We follow the same path as frontend with Zulu lang trick (take a look here https://github.com/Vizzuality/heco-invest/tree/develop/frontend#translations).

I believe when translations are done, we can just pull all of them into the repo and get rid of Zulu.

I wouldn't keep Transifex indefinitely as a database for those translations, especially if that costs money. Periodic manual pull to the repo would be good. 

For deployments, we can add auto pull into Dockerfile to always have fresh translations there. I think frontend will be doing the same in Dockerfile.

Beware of pulling only translated content, `--mode onlytranslated`, otherwise you end up with `yml` file fill up with empty strings which is weird.

I will update all those enums with proper descriptions from design in different pull request.

## Testing instructions

Dunno, `tx push -s`, `tx pull --mode onlytranslated`

## Tracking

https://vizzuality.atlassian.net/browse/LET-145


clementprdhomme:
Hey,

I'll let Agnieszka review the code. I only have two comments:

- At the moment, Transifex is free. It's free for open source projects so I don't think the cost will be an issue.
- I renamed the [resource on Transifex](https://www.transifex.com/heco-invest/heco-invest/content/) as “Back-end application”. I believe that would be easier to understand for the client.

clementprdhomme (approved):

backend/.tx/config
```diff
@@ -0,0 +1,8 @@
+[main]
+host = https://www.transifex.com
```
martintomas, line 2: cool, really nice how simple this is ;)

backend/app/models/investor.rb
```diff
@@ -10,7 +10,7 @@ class Investor < ApplicationRecord
   validates :impacts, array_inclusion: {in: Impact::TYPES}
   validates :sdgs, array_inclusion: {in: Sdg::TYPES}
   validates :investor_type, inclusion: {in: InvestorType::TYPES}
-  validates :language, inclusion: {in: I18n.available_locales.map(&:to_s)}
+  validates :language, inclusion: {in: (I18n.available_locales - [:zu]).map(&:to_s)}
```
martintomas, line 13: Maybe it would be worthy to make`I18n.available_locales - [:zu]` more dry so it does not repeat all the time here. It should be possible to use `StaticModel` approach for this --> `Language::TYPES`, where
```
class Language
  TYPES = I18n.available_locales - [:zu]).map(&:to_s)
end
```
then it is super easy to add some extra stuff to language later on

martintomas (approved):
Really like simplicity of `transifex ` ;)

agnessa (commented):
Can you add a line or to to the README about how to update translations?

tsubik:
I will update this PR later today, so please don't merge.

PR #25, opened 2022-03-14
Health Check endpoint

martintomas:
Simple endpoint for health checking

## Testing instructions

`GET /health_check`

## Tracking

https://vizzuality.atlassian.net/browse/LET-154


tsubik (approved):
I like the simplicity. We can start with this and improve eventually if there will be need to check connections to other services. (like redis)

agnessa (approved):

PR #26, opened 2022-03-15
Use Rswag

martintomas:
I have added Rswag gem and rewritten Investors API tests to actually use it. I have tried to be compatible with `match_snapshot` approach as much as possible and used rswag dsl as wrapper for all existing tests. Please @tsubik or @agnessa check it and if you are fine with this approach, I will refactor rest of the API tests ;)

**Note**:  This PR also adds `backend/swagger/v1/swagger.yaml` which is dynamically generated file that contains API documentation. Better approach is to generate this file during pipeline run so you are sure that documentation is always up to date. Please take into account that to actually generate this documentation from testing files, rails app needs to have access to testing database.

Screenshot 2022-03-15 at 12 13 47

## Testing instructions

To generate swagger documentation, please use `rake rswag:specs:swaggerize`. Afterwards, you can found it at `/api-docs`.

## Tracking

https://vizzuality.atlassian.net/browse/LET-153

### DIFF
--- a/backend/Gemfile
+++ b/backend/Gemfile
@@ -7,6 +7,7 @@ gem "rails", "~> 7.0.1"
 gem "pg", "~> 1.1"
 gem "puma", "~> 5.0"
 gem "redis", "~> 4.0"
+gem "rswag"
 
 gem "cssbundling-rails"
 gem "jsbundling-rails"
--- a/backend/Gemfile.lock
+++ b/backend/Gemfile.lock
@@ -66,6 +66,8 @@ GEM
       i18n (>= 1.6, < 2)
       minitest (>= 5.1)
       tzinfo (~> 2.0)
+    addressable (2.8.0)
+      public_suffix (>= 2.0.2, < 5.0)
     annotate (3.2.0)
       activerecord (>= 3.2, < 8.0)
       rake (>= 10.4, < 14.0)
@@ -109,6 +111,8 @@ GEM
       reline (>= 0.3.0)
     jsbundling-rails (1.0.2)
       railties (>= 6.0.0)
+    json-schema (2.8.1)
+      addressable (>= 2.4)
     jsonapi-serializer (2.2.0)
       activesupport (>= 4.2)
     loofah (2.14.0)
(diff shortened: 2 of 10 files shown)


tsubik (commented):
That looks great to me! Just one question, is it possible to have example value from response/factories. Right now, I see type names instead of some example values. We probably don't want to have docs on production (or we can add some base auth layer, but that is for later actually), only staging/development, thoughts @agnessa?

tsubik:
I think there is also option to convert swagger.json to Postman collections. I wonder what would be better for frontend guys, but this is something they can always do on their own.

tsubik:
Maybe something like this https://github.com/rswag/rswag#enable-auto-generation-examples-from-responses works.

agnessa:
> That looks great to me! Just one question, is it possible to have example value from response/factories. Right now, I see type names instead of some example values. We probably don't want to have docs on production (or we can add some base auth layer, but that is for later actually), only staging/development, thoughts @agnessa?

I agree, perhaps let's suppress in production. As for sample values, agree it would be nice if not too crazy!

backend/app/controllers/api/v1/base_controller.rb
```diff
@@ -20,7 +20,7 @@ def sparse_fieldset
       end
 
       def included_relationships
-        params[:include]&.split(",")
+        params[:includes]&.split(",")
```
martintomas, line 23: I had to rename `include` param, because `rswag` complained when I used `include` key. I guess that it is not issue at this part of project, but let me know if you thing otherwise :)

backend/spec/requests/api/v1/enums_spec.rb
```diff
@@ -1,12 +1,22 @@
...
-      expect(response).to have_http_status(:ok)
-      expect(response.body).to match_snapshot("api/v1/enums")
+      response "200", :success do
+        schema type: :object, properties: {
+          data: {type: :array, items: {"$ref" => "#/components/schemas/enum"}}
+        }
+
+        run_test!
+
+        it "matches snapshot", generate_swagger_example: true do
```
martintomas, line 16: to reuse examples generated by factories, I have added flag `generate_swagger_example ` -- it identifies test which response should be used as example at documentation

agnessa (approved):

PR #28, opened 2022-03-16
Implement the input component

clementprdhomme:
This PR adds a simple input that can be used as a text, number, email or password input.

## Testing instructions

- Respects the [visual design](https://www.figma.com/file/3epSYh4KQZBCyh4Y5OzmHw/HeCo---Design-System?node-id=2313%3A39246)
- Can be set as text, number, url or password input type
- Has visual error state when input validation fails ([see error state example](https://www.figma.com/file/3epSYh4KQZBCyh4Y5OzmHw/HeCo---Design-System?node-id=2377%3A3138))
- Compatible with React Hook Form (see [example of integration](https://react-hook-form.com/get-started/#Integratinganexistingform))
- Accessible by [WCAG 2.1 standards](https://www.w3.org/TR/WCAG21/)
- Focused state for keyboard users
- Requires an 'id' (to link to a label externally) and 'type' props
- Available in Storybook

## Tracking

[LET-163](https://vizzuality.atlassian.net/browse/LET-163).


clementprdhomme:
I pushed a minor change so that the component appears in the correct folder in Storybook and so that the `register` prop cannot be changed (in Storybook as well).

PR #29, opened 2022-03-17
Improve the appearance of Storybook

clementprdhomme:
This PR makes two changes to Storybook:

1. The preview now has a light background colour which is the same used on the platform
2. The description and types of the props are shown in the “Controls” panel by default

The preview has a light background and the descriptions and types of the props are shown in the controls panel

## Testing instructions

Make sure:
- the background of the preview window uses the “background-light” colour
- the rest of the UI does not use this colour as background
- the description and types of the props are shown by default in the “Controls” panel

## Tracking

Not tracked.


frontend/.storybook/preview.js
```diff
@@ -1,12 +1,12 @@
 import React from 'react';
 
-import { themes } from '@storybook/theming';
 import { OverlayProvider } from '@react-aria/overlays';
 import { SSRProvider } from '@react-aria/ssr';
 import { reactIntl, localesNames } from './react-intl.js';
 import '../styles/globals.css';
 
 export const parameters = {
+  controls: { expanded: true },
```
SARodrigues, line 9: So that's the one we were missing! 👍

SARodrigues (approved):
Looks much better now!

PR #30, opened 2022-03-17
feat: Location model and importer (with Colombia Wiki data)

martintomas:
First iteration of importing Colombia location data. I have fetched data from wiki where department names and list of municipalities can be found. 

Crawler also downloaded coordinates for every municipality from wiki --> "almost" all locations had coordinates there ;). I plan to use these municipality coordinates during next iteration when I add regions -- need to compute intersections between regions and municipalities. Because of complexity of these two tasks I have decided to split this to two separate PRs - one for administrative regions (which are covered by this PR) and second for geographic regions.

## Testing instructions

Just run `rake db:seed` and check that locations data were correctly imported ;)

## Tracking

https://vizzuality.atlassian.net/browse/LET-130

### DIFF
--- a/backend/db/migrate/20220317074221_create_locations.rb
+++ b/backend/db/migrate/20220317074221_create_locations.rb
@@ -1,0 +1,22 @@
+class CreateLocations < ActiveRecord::Migration[7.0]
+  def change
+    create_table :locations, id: :uuid do |t|
+      t.belongs_to :parent, foreign_key: {on_delete: :cascade, to_table: :locations}, type: :uuid
+
+      t.string :location_type, null: false
+
+      t.text :name_en
+      t.text :name_es
+      t.text :name_pt
+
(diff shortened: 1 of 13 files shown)


backend/db/migrate/20220317074221_create_locations.rb
```diff
@@ -0,0 +1,22 @@
...
+    create_table :locations, id: :uuid do |t|
+      t.belongs_to :parent, foreign_key: {on_delete: :cascade, to_table: :locations}, type: :uuid
+
+      t.string :location_type, null: false
+
+      t.text :name_en
+      t.text :name_es
+      t.text :name_pt
+
+      t.index [:location_type, :parent_id, :name_en], unique: true, where: "parent_id IS NOT NULL", name: "uniq_name_en_with_parent_id"
```
agnessa, line 12: nice! 💪

agnessa (approved):

PR #32, opened 2022-03-17
Implement the layout of the project developer public page

SARodrigues:
This PR adds the project developer public page. 

**In this PR**:  
- New `Carousel` component  
  _Very simple Carousel component, following the example from [W3.org](https://www.w3.org/WAI/tutorials/carousels/)._  
  _Added Storybook examples with it being used with images, regular elements, and elements/slides of different heights_

- New `ProjectCard` component 
  _Which can both be passed an `href` to link to (eg: project's URL), or a callback to be executed on click/press (eg: to trigger an info card/modal when used in the project search screens). Built based on [[Inclusive Components](https://inclusive-components.design/)](https://inclusive-components.design/cards/)'s card recommendations. It can be navigated to by keyboard, and its contents read by screen readers._ 
   _Added to Storybook_

- New `usePropagablePointerEventHandler` hook
    _[`usePress`](https://react-spectrum.adobe.com/react-aria/usePress.html) (and [`useButton`](https://react-spectrum.adobe.com/react-aria/useButton.html)) from `react-spectrum` stops event propagation. This can be a problem, especially when use carousels (like the one above) that rely on pointer events._
    _This hook allows us to clone native pointer events and re-emit them, restoring event propagation (only where needed), which was the case for making the `ProjectCard`s work inside the `Carousel`'s slides._

- New `project-developer` page  
  _Which makes use of the shared `ProfileHeader` and `TagsGrid` components, and also can display `ProjectCard`s inside a `Carousel` as per designs, with functional keyboard navigation._
  _Note that actual pagination/infinite scroll has not been added that, although it can most likely be a simple wrapper around the added Carousel (with a couple extra props needed), but IMHO it's easier to implement when we actually connect to the API._

## Tracking

[LET-141](https://vizzuality.atlassian.net/browse/LET-141)

## Screenshot

screencapture-localhost-3000-project-developer-id-2022-03-18-11_46_59

## Video

https://user-images.githubusercontent.com/6273795/159024484-4555dd9c-65e1-4925-990f-bcc50241344b.mp4


frontend/components/carousel/component.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+import { debounce } from 'lodash-es';
+
+import AriaLive from './aria-live';
+import Arrows from './arrows';
+import Paging from './paging';
+import { SlideProps } from './slide';
+import type { CarouselProps } from './types';
+
+export const Carousel: FC<CarouselProps> = ({ className, children }: CarouselProps) => {
```
clementprdhomme, line 25: The arrows are not visible on narrow screens and the dots (pagination) looks a bit cramped.

clementprdhomme: On iOS 12.5.5 (my iPhone), the swipe doesn't work on the slides though it works on newer iOS versions. I guess the problem is with the pointer events.

If you find an easy way to fix this, that'd be great (either by changing the events or having a  polyfill such as https://cdn.polyfill.io/v3/). If that's too much effort, we can forget about this issue since we have the arrows and dots.

clementprdhomme (commented):
Great job @SARodrigues 👏, especially on the accessibility and the carousel!

frontend/components/carousel/component.tsx
```diff
@@ -0,0 +1,171 @@
...
+      window.addEventListener('resize', handleResize);
+    };
+  }, [handlePointerDown, handlePointerUp, handlePointerMove, handleResize]);
+
+  useEffect(() => {
+    slidesWrapperContainerRef.current.scrollLeft = 0;
+    setSlideHeight(slidesWrapperRef?.current?.offsetHeight || '100%');
+  }, [currentSlide, numSlides]);
+
+  useLayoutEffect(() => {
```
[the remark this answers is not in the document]
SARodrigues: Not sure why I went with `useLayoutEffect` in the first place... I'm going to blame this one on VSCode autocomplete 😁

frontend/components/carousel/component.tsx
SARodrigues: I hid the arrows for small screens. I'll give it a go at debugging it, but if it takes to long I can re-enable them (maybe move them to the bottom next to the dots.

SARodrigues: I couldn't find a great solution, so I've changed it so that the arrows are shown on small screens, next to the dots. 

I've tried a few polyfills (and made the component noSSR, but the behavior of the slider was sketchy to say the least. 

Screenshot 2022-03-22 at 15 38 22

clementprdhomme: That's good enough. It looks very nice 👏!

clementprdhomme (approved):

PR #34, opened 2022-03-18
Frontend Dockerfile - first iteration

tiagojsag:

- You ran the tests
    
---

Description of the PR

## Testing instructions

What to test? How to do it?

## Tracking

Link to the task(s), if any

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -28,6 +28,8 @@ ENV NEXT_TELEMETRY_DISABLED 1
 ARG TRANSIFEX_TOKEN
 ENV TX_TOKEN $TRANSIFEX_TOKEN
 
+RUN apk add curl bash
+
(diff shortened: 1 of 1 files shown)


clementprdhomme (approved):
Just a quick note that when deploying to production, we'll have to pull the translated strings from Transifex, as we do on staging: https://github.com/Vizzuality/heco-invest/blob/develop/.github/workflows/frontend-deployment.yml#L67-L84.

I can help you understand how we're deploying to development/staging at the moment.

tiagojsag:
thanks for the feedback and the clarification call. Based on it, I made a few additional changes:
- I pinned the transifex CLI version, as opposed to loading from "master", which is evil
- I did away with the file that holds the transifex API key, and instead am using an env var, which is documented, and overall a saner approach.

PR #35, opened 2022-03-18
Adding regions to Colombia locations

martintomas:
Second part of feature focused on import of location data. This one is about Colombia regions (mosaics) ;)

I have added new model `LocationMember` which allows to have `have_many` relation between other locations (regions). Reason for this is that administrative location (state, municipality, etc.) can be at multiple geographic locations (regions). Locations importer now supports import of regions and also assigns these regions to correct municipalities.

About regions data -- I have used shapefile provided via Jira ticket to get regions and downloaded coordinates for every municipality from wikipedia. Then I have computed intersections and decided which municipalities are at which regions. Unfortunately, this approach is super inaccurate, because whole municipality is represented just by one point and even though point itself is not at region, part of the municipality still can be. Unfortunately, I don't have exact geometry for every municipality so this is as far as I can go with wikipedia data.
 
colombia_regions

## Testing instructions

Same as first iteration of location import feature. Run `rake db:seed` and check that LocationMember table contains data ;)

## Tracking

https://vizzuality.atlassian.net/browse/LET-130

### DIFF
--- a/backend/app/controllers/api/v1/enums_controller.rb
+++ b/backend/app/controllers/api/v1/enums_controller.rb
@@ -8,7 +8,8 @@ module API
           TicketSize.all,
           Impact.all,
           # ProjectDeveloperType.all, wait for better definition
-          InvestorType.all
+          InvestorType.all,
+          LocationType.all
         ].flatten
 
         serialized_enums = data.map do |d|
--- a/backend/app/controllers/api/v1/locations_controller.rb
+++ b/backend/app/controllers/api/v1/locations_controller.rb
@@ -1,0 +1,36 @@
+module API
+  module V1
+    class LocationsController < BaseController
+      before_action :fetch_location, only: :show
+
+      def index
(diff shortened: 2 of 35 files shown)


martintomas:
Okay, I have used data from https://docs.google.com/spreadsheets/d/1QKYxL7nuZXJJ9Q2MJD_MN31s0ObllK8b/edit?usp=sharing&ouid=118214021291914641487&rtpof=true&sd=true for departments and municipalities, which has some pros and cons at this point:
 - departments and municipalities from trusted source, no wiki data ;)
 - we had at least one coordinate from wiki per every municipality which is no more true --> till we got proper geometry for every municipality from that list, we cannot really import location members (intersections between regions and municipalities). 

Whatever, Importer changes which are part of this MR are still valid and can be merged to dev. After we will have correct geometries, we can finish import of `location_members`.

agnessa (approved):

PR #36, opened 2022-03-21
implement checkbox component

barbara-chaves:
Add a checkbox component

## Testing instructions

* Respects the [visual design](https://www.figma.com/file/3epSYh4KQZBCyh4Y5OzmHw/HeCo---Design-System?node-id=2313%3A39279)
* Has visual error state when input validation fails ([see text input error example](https://www.figma.com/file/3epSYh4KQZBCyh4Y5OzmHw/HeCo---Design-System?node-id=2377%3A3138))

    * Focused state for keyboard users

    * Requires a label

* Available in Storybook

## Tracking

[LET-164](https://vizzuality.atlassian.net/browse/LET-164?atlOrigin=eyJpIjoiODQ3MDY5ODEyY2Q1NDJjOGEyYmI5NzcwN2RhN2FiYjUiLCJwIjoiaiJ9)


frontend/components/forms/checkbox/component.stories.tsx
```diff
@@ -1,43 +1,134 @@
...
+  );
+};
+
+export const ErrorState: Story<CheckboxProps<FormValues>> = TemplateWithForm.bind({});
+ErrorState.args = {
+  id: 'form-accept',
+  name: 'accept',
+  registerOptions: {
+    disabled: false,
+    required: 'Accept term and Privacy Policy is required.',
```
clementprdhomme, line 132:
```suggestion
    required: 'Accept Terms and Privacy Policy is required.',
```

frontend/components/forms/checkbox/types.d.ts
```diff
@@ -1,7 +1,21 @@
-import { InputHTMLAttributes } from 'react';
+import React, { InputHTMLAttributes } from 'react';
```
clementprdhomme, line 1: Can you rename the file `types.ts` instead of `types.d.ts`?

frontend/components/forms/checkbox/component.tsx
```diff
@@ -1,34 +1,52 @@
...
-}: CheckboxProps) => {
-  const st = useStatus({ meta, disabled });
+  const [invalid, setInvalid] = useState(false);
+  const state = useToggleState(props);
+  const { isSelected, toggle } = state;
+  const { isFocusVisible, focusProps } = useFocusRing(rest);
 
+  const ref = useRef();
+  const { inputProps } = useCheckbox(rest, state, ref);
+  console.log(isFocusVisible);
```
clementprdhomme, line 24: You don't need any of this since you're using a native checkbox 😉.
```suggestion
```

frontend/components/forms/checkbox/component.tsx
```diff
@@ -1,34 +1,52 @@
...
-      {...input}
-      {...props}
-      type="checkbox"
-      disabled={disabled}
-      className={cx({
-        'form-checkbox': true,
-        [THEME[theme].base]: true,
-        [THEME[theme].status[st]]: true,
-        [className]: !!className,
+      className={cx(
```
clementprdhomme, line 27: You can greatly simplify the classes (have a look at the other form inputs). For example, you don't need `isSelected` or `isFocusVisible`. Use `:checked` or `:focus` instead. You also don't need to specify the duration of the animations.

clementprdhomme: Instead of `hover:transition-all` I would prefer `transition-all` so that we always have animations. If you can also use `transition` instead of `transition-all`, it's even better.

clementprdhomme: I don't think `bg-[url("/images/checkbox-checked.svg")]` is correct. It should be `checked:bg-[url("/images/checkbox-checked.svg")]` instead, no?

frontend/components/forms/checkbox/component.tsx
```diff
@@ -1,34 +1,52 @@
...
+  const [invalid, setInvalid] = useState(false);
+  const state = useToggleState(props);
+  const { isSelected, toggle } = state;
+  const { isFocusVisible, focusProps } = useFocusRing(rest);
 
+  const ref = useRef();
+  const { inputProps } = useCheckbox(rest, state, ref);
+  console.log(isFocusVisible);
   return (
     <input
```
clementprdhomme, line 26: Please specify the type of the input: `type="checkbox"`. It's very important for accessibility.

frontend/components/forms/checkbox/component.tsx
clementprdhomme: Don't forget to add `outline-none` in the list of classes otherwise we'll keep seeing the default focus ring of the browser.

frontend/components/forms/checkbox/component.stories.tsx
```diff
@@ -1,43 +1,134 @@
...
 Default.args = {
-  theme: 'dark',
+  id: 'form-accept',
+  name: 'accept',
+  registerOptions: {
+    disabled: false,
+  },
+};
+
+const TemplateWithLabel: Story<CheckboxProps<FormValues>> = (args: CheckboxProps<FormValues>) => {
```
clementprdhomme, line 49: This story is not accessible. When you click on the label, the checkbox should get the focus and toggle. If you try in Storybook, you'll see it's not the case. Below I will detail why it's not working.

frontend/components/forms/checkbox/component.stories.tsx
```diff
@@ -1,43 +1,134 @@
...
+  const { register } = useForm<FormValues>();
+
+  return (
+    <div className="p-3">
+      <label htmlFor="check">
```
clementprdhomme, line 54: In order to link the label with the input, `htmlFor` must contain the ID of the checkbox. Here you can see the ID it references is `'check'` while the ID of the checkbox is `'checkbox'`.

frontend/components/forms/checkbox/component.stories.tsx
```diff
@@ -1,43 +1,134 @@
...
+        <Checkbox
+          id="checkbox"
```
clementprdhomme, line 56: There's also another issue here. This ID will be overwritten by the ID passed from the args. On line 70 you can see you pass another ID: `'form-accept'`. Here is my suggestion: 
```suggestion
          id={args.id}
```

Of course, if the ID of the checkbox is now `args.id`, you should also update the label's `htmlFor` attribute.

clementprdhomme (changes requested):

frontend/components/forms/checkbox/component.tsx
barbara-chaves: You are right! I am going to remove it.

frontend/components/forms/checkbox/types.ts
```diff
@@ -0,0 +1,23 @@
...
+  /** Name of the input */
+  name: Path<FormValues>;
+  /** React Hook Form's `register` function */
+  register: UseFormRegister<FormValues>;
+  /** Options for React Hook Form's `register` function */
+  registerOptions?: RegisterOptions<FormValues, FieldPath<FormValues>>;
+  /** class to change checkbox element style */
+  className?: string;
+
+  disabled?: boolean;
```
clementprdhomme, line 19: This prop is unnecessary because `RegisterOptions<FormValues, FieldPath<FormValues>>` already has a `disabled` attribute.

frontend/components/forms/checkbox/component.tsx
```diff
@@ -1,34 +1,34 @@
...
-      disabled={disabled}
-      className={cx({
-        'form-checkbox': true,
-        [THEME[theme].base]: true,
-        [THEME[theme].status[st]]: true,
-        [className]: !!className,
+      className={cx(
+        'appearance-none inline-block w-4 h-4 mr-2 mt-0.5 px-0.5 py-[3px] border border-beige rounded hover:border hover:border-green-dark outline-none focus-visible:ring-green-dark focus-visible:ring-2 focus-visible:ring-offset-2 focus-visible:ring-offset-white  checked:border-green-dark checked:bg-green-dark checked:bg-[url("/images/checkbox-checked.svg")] bg-no-repeat bg-center transition',
+        {
+          'border-red': invalid,
```
clementprdhomme, line 21: This class should only be applied when the input is invalid. `onInvalid` is not re-executed if the input becomes valid.
```suggestion
          'invalid:border-red': invalid,
```

frontend/components/forms/checkbox/component.tsx
```diff
@@ -1,34 +1,34 @@
...
-      {...props}
       type="checkbox"
-      disabled={disabled}
-      className={cx({
-        'form-checkbox': true,
-        [THEME[theme].base]: true,
-        [THEME[theme].status[st]]: true,
-        [className]: !!className,
+      className={cx(
+        'appearance-none inline-block w-4 h-4 mr-2 mt-0.5 px-0.5 py-[3px] border border-beige rounded hover:border hover:border-green-dark outline-none focus-visible:ring-green-dark focus-visible:ring-2 focus-visible:ring-offset-2 focus-visible:ring-offset-white  checked:border-green-dark checked:bg-green-dark checked:bg-[url("/images/checkbox-checked.svg")] bg-no-repeat bg-center transition',
```
clementprdhomme, line 19: Could you change the opacity when the input is disabled? You can use `disabled:opacity-60` (it's used on the other inputs).

SARodrigues (approved):

PR #37, opened 2022-03-21
API endpoint for locations

martintomas:
I am adding two new API endpoints for location --> one for `index` and second for `show/detail`.

Only other thing which is probably worthy to mention is that `index` locations API endpoint allows to filter locations based on:
 - location_type
 - parent_id

I have also added LocationType values to Enum ;).

## Testing instructions

Please check swagger documentation, there are two new api endpoints for locations:
Screenshot 2022-03-21 at 11 42 21

```
{
  "data": {
    "id": "4debca7f-9251-403f-b893-ff0e78bd6af4",
    "type": "location",
    "attributes": {
      "name": "Hane, Lehner and Goyette",
      "location_type": "municipality"
    },
    "relationships": {
      "parent": {
        "data": {
          "id": "5331a4ba-ef75-4b7f-8cdd-8ec39e692bbd",
          "type": "location"
        }
      },
      "regions": {
        "data": [
          {
            "id": "b5debe16-a4ab-4811-ac48-3b480af1fc56",
            "type": "location"
          }
        ]
      }
    }
  }
}
```

## Tracking

https://vizzuality.atlassian.net/browse/LET-129

### DIFF
--- a/backend/app/models/location_type.rb
+++ b/backend/app/models/location_type.rb
@@ -2,7 +2,7 @@ class LocationType
   include EnumModel
 
   TYPES = %w[
-    state
+    country
     department
     municipality
     region
--- a/backend/app/controllers/api/v1/enums_controller.rb
+++ b/backend/app/controllers/api/v1/enums_controller.rb
@@ -8,7 +8,8 @@ module API
           TicketSize.all,
           Impact.all,
           # ProjectDeveloperType.all, wait for better definition
-          InvestorType.all
+          InvestorType.all,
+          LocationType.all
         ].flatten
 
         serialized_enums = data.map do |d|
--- a/backend/app/controllers/api/v1/locations_controller.rb
+++ b/backend/app/controllers/api/v1/locations_controller.rb
@@ -1,0 +1,36 @@
+module API
+  module V1
+    class LocationsController < BaseController
+      before_action :fetch_location, only: :show
+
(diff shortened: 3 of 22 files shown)


backend/app/models/location_type.rb
```diff
@@ -2,7 +2,7 @@ class LocationType
   include EnumModel
 
   TYPES = %w[
-    state
+    country
```
martintomas, line 5: I am renaming `state` to `country` so this keeps naming convention used at design

tsubik (approved):
Nice!

backend/app/models/location_type.rb
```diff
@@ -2,7 +2,7 @@ class LocationType
...
     department
```
agnessa, line 6: do we then need to rename department to state?

martintomas: :D should I? I am fine with renaming `state` to `country` because it still makes sense for me, but in this case `department` still makes more sense to me then `state`. Also wiki and Santiago's data speak about these entities as departments. But if you wanna go for `state`, I will do it, no problem -- just please let me know your final word and I will tweak it eventually before merging this 🙏

agnessa: you're right, in the file it is called department, in the design state. Let's stick with department for now then.

PR #38, opened 2022-03-21
Implement a `Select` component

clementprdhomme:
This PR adds a `<Select />` component to be used in forms with [React Hook Form](https://react-hook-form.com/).

A few notes on the changes:

- The component doesn't receive the usual `register` and `registerOptions` props but instead `control` and `controlOptions`. This is because with [React Aria's implementation](https://react-spectrum.adobe.com/react-aria/useSelect.html) of a `<select />` component, we don't get access to a real input to which we can assign arbitrary props. The real input would be needed to check, for example, the validation state. Instead, `control` is used and it allows to retrieve React Hook Form's validation state.
- As a consequence of what's written above, there is no native validation of the input (only React Hook Form's). Screen reader users can still be made aware of the state of the component using external errors linked to the component via `aria-describedby` (there is an example on Storybook).
- I've removed the `fill-current` class of the `<Icon />` component because it didn't pay nice with the icons coming from `react-feather` (some use stroked instead of fill colours). I reviewed the code and I don't think it has broken any icon.

## Testing instructions

- Respects the [visual design](https://www.figma.com/file/3epSYh4KQZBCyh4Y5OzmHw/HeCo---Design-System?node-id=2313%3A39262)
- Has visual error state when input validation fails ([see text input error example](https://www.figma.com/file/3epSYh4KQZBCyh4Y5OzmHw/HeCo---Design-System?node-id=2377%3A3138))
- Compatible with React Hook Form (see [example of integration](https://react-hook-form.com/get-started/#Integratinganexistingform))
- Accessible by [WCAG 2.1 standards](https://www.w3.org/TR/WCAG21/)
- Focused state for keyboard users
- Requires an 'id' prop to link to a label externally
- Available in Storybook

## Tracking

[LET-171](https://vizzuality.atlassian.net/browse/LET-171).


SARodrigues (approved):
> I've removed the fill-current class of the <Icon /> component because it didn't pay nice with the icons coming from react-feather (some use stroked instead of fill colours). I reviewed the code and I don't think it has broken any icon.

I don't think it breaks anything either; had the same issue in another branch.

SARodrigues:
Rebased to fix a conflict with `develop`, will merge after tests pass. :)

PR #39, opened 2022-03-22
Feature/backend auth

tsubik:
WIP

## Testing instructions

What to test? How to do it?

## Tracking

Link to the task(s), if any


backend/app/controllers/api/v1/csrf_controller.rb
```diff
@@ -0,0 +1,9 @@
+module API
+  module V1
+    class CsrfController < BaseController
```
martintomas, line 3: I guess this controller is just for needs of tests, maybe it is even not needed 🤔

backend/app/controllers/api/v1/sessions_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+  module V1
+    class SessionsController < BaseController
+      before_action :authenticate_user!, only: [:destroy]
+
+      def create
+        user = User.find_by_email(params[:email])
+
+        if user&.valid_password?(params[:password])
+          # warden checks devise options like confirmable
+          # if not confirmed then throws an error that is handled by failure_app
```
martintomas, line 11: I personally believe that this warden check is done even before it comes here --> it somehow detects that session exists and tries to load user. Whatever, quite mystery for me :D

backend/app/controllers/api/v1/sessions_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+          sign_in user
+          render json: UserSerializer.new(user).serializable_hash
+        else
+          raise API::UnprocessableEntityError, "Invalid email or password"
+        end
+      end
+
+      def destroy
+        sign_out current_user
+        render json: {data: nil}
```
martintomas, line 21: maybe return of `head :ok` could be enough

backend/app/models/user.rb
```diff
@@ -0,0 +1,22 @@
...
+  enum role: {light: 0, investor: 1, project_developer: 2}, _default: :light
+
+  validates :password,
+    length: {minimum: 12, message: "must be at least 12 characters long"},
+    allow_nil: true
+  validate :password_complexity
+  validates :ui_language, inclusion: {in: Language::TYPES}
+  validates_presence_of :first_name, :last_name
+
+  def password_complexity
```
martintomas, line 16: should be `private`

backend/app/models/user.rb
```diff
@@ -0,0 +1,22 @@
...
+    return if password.blank?
+    return if password.match?(/^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)./)
+
+    errors.add :password, "must include at least one lowercase letter, one uppercase letter, and one digit"
```
martintomas, line 20: and probably localized :)

backend/app/views/rswag/ui/home/index.html.erb
```diff
@@ -0,0 +1,113 @@
...
+    // Apply mandatory parameters
+    configObject.dom_id = "#swagger-ui";
+    configObject.presets = [SwaggerUIBundle.presets.apis, SwaggerUIStandalonePreset];
+    configObject.layout = "StandaloneLayout";
+
+    // If oauth2RedirectUrl isn't specified, use the built-in default
+    if (!configObject.hasOwnProperty("oauth2RedirectUrl"))
+      configObject.oauth2RedirectUrl = window.location.href.replace("index.html", "oauth2-redirect.html");
+
+    configObject.requestInterceptor = function (req) {
```
martintomas, line 100: haha, never seen this, cool ;). Tried it and I actually needed to click on Authorize button to make it work, but then It worked perfectly.

Screenshot 2022-03-22 at 18 18 09

martintomas (approved):
Looks really nice!

backend/app/controllers/api/v1/sessions_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+      def create
+        user = User.find_by_email(params[:email])
+
+        if user&.valid_password?(params[:password])
+          # warden checks devise options like confirmable
+          # if not confirmed then throws an error that is handled by failure_app
+          sign_in user
+          render json: UserSerializer.new(user).serializable_hash
+        else
+          raise API::UnprocessableEntityError, "Invalid email or password"
```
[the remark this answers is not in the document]
tsubik: Yeah, I will add that, will use devise string for that.

backend/app/controllers/api/v1/sessions_controller.rb
tsubik: Hmm, yeah, I will be simpler. We can do it for all similar endpoints that perform some action but no response is needed.

backend/app/models/user.rb
```diff
@@ -0,0 +1,22 @@
+class User < ApplicationRecord
+  belongs_to :account, optional: true
+
+  devise :database_authenticatable, :confirmable, :registerable,
+    :recoverable, :rememberable, :validatable
+
+  enum role: {light: 0, investor: 1, project_developer: 2}, _default: :light
```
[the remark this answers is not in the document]
tsubik: What do you mean? You mean, that instead of `has_one` could be `has_many` that is true from db schema point of view, also account investor/project_developer relation is not polymorphic.

That role here is just some cache. I did it differently before like:

```
 def role
    return nil if account.nil?
    return "investor" if account.investor.present?
    return "project_developer" if account.project_developer.present?

    nil
  end
``` 

I wonder if to use strings instead of numbers for all rails enums (for sake of consistency with not-rails enums), like `light: "light"` etc. What do you think?

backend/app/models/user.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  validates :password,
+    length: {minimum: 12, message: "must be at least 12 characters long"},
```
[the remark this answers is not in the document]
tsubik: true story

backend/app/views/rswag/ui/home/index.html.erb
tsubik: Hmm, or just at least one request to API must be done to initiate session. I will try. I probably clicked everything :D

backend/spec/requests/api/v1/session_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+    @user = create(:user, email: "user@example.com", password: "SuperSecret1234")
+  end
+
+  path "/api/v1/session" do
+    post "Creates User Session/Logs In" do
+      tags "Session"
+      consumes "application/json"
+      produces "application/json"
+      security [csrf: []]
+      parameter name: :user_params, in: :body, type: :object,
```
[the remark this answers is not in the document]
tsubik: I tried it, it didn't work. I tried like `parameter name: :email, in: :body, type: :string` or smth like this. In examples it is also done with json object. I don't know why there is no example body, I think there should be something like 

```
{
   "email": "string",
   "password: "string"
}
```

backend/spec/swagger_helper.rb
```diff
@@ -20,8 +20,31 @@
       },
       paths: {},
       components: {
-        securitySchemes: {},
+        securitySchemes: {
```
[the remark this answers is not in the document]
tsubik: In this case it will be just informative :)

backend/app/views/rswag/ui/home/index.html.erb
tsubik: I tried it once again. You don't have to click on Authorize. Just if you don't have session, you need to do first request. First request to `POST /api/v1/session` and I have authenticity token error, but second request I have wrong email or password, so it works as expected.

backend/spec/requests/api/v1/session_spec.rb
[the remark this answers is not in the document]
tsubik: Ahh, I misread your message and see that I'm missing schema now :) maybe I worked too much today already

backend/app/models/user.rb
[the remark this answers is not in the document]
martintomas: Ohh, this is completely fine, I just though that you can be investor and project developer at same time, because of
```
has_one :investor
has_one :project_developer
```
so kind of expected that user will have `roles` instead of `role`. Whatever, I am fine with this, no big deal

PR #45, opened 2022-03-24
API endpoint for creating project developer

martintomas:
I have read through required fields for project developer formular and added missing ones:
 - picture (active storage)
 - entity_legal_registration_number
 - locations (has many relation)

Then I have added new api endpoint for creating project developer with all required fields -- it creates account, updates user role and creates project developer record. Error is returned in case user already have account.

There are a few tricky parts which I was not sure how to handle and which I will comment at code at this PR. I have also discovered two rswag bugs, which made testing a bit difficult.

## Testing instructions

Rebuild database (drop and create if from scratch) and check swagger api docs. There should be new API endpoint for creating project developers `POST /api/v1/project_developers`

Screenshot 2022-03-24 at 16 16 00

## Tracking

https://vizzuality.atlassian.net/browse/LET-187


backend/app/controllers/api/v1/project_developers_controller.rb
```diff
@@ -3,6 +3,9 @@ module V1
     class ProjectDevelopersController < BaseController
       include API::Pagination
 
+      skip_before_action :require_json!, only: :create
```
martintomas, line 6: I had to upload picture file for project developer so I have decided to go with `multipart/form-data` content type instead.

backend/app/controllers/api/v1/project_developers_controller.rb
```diff
@@ -23,6 +26,17 @@ def show
         ).serializable_hash
       end
 
+      def create
+        current_user.with_lock do
+          raise API::UnprocessableEntityError, I18n.t("errors.messages.user_multiple_accounts") if current_user.account_id.present?
```
martintomas, line 31: I wasn't really sure what should happen when user already have account -- I decided to raise error :)

backend/app/controllers/api/v1/project_developers_controller.rb
```diff
@@ -31,6 +45,17 @@ def fetch_project_developer
         account = Account.friendly.find(params[:id])
         ProjectDeveloper.find_by!(account_id: account.id)
       end
+
+      def account_params
+        params.fetch(:project_developer_params, params)
```
martintomas, line 50: this is one of the rswag bugs which is fixed by this not super nice hack (https://github.com/rswag/rswag/issues/348). Unfortunately when using `formData`, rswag automatically adds name of the variable to param which in this case is `project_developer_params `

backend/config/application.rb
```diff
@@ -11,6 +11,8 @@ class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 7.0
 
+    Rails.application.routes.default_url_options[:host] = ENV.fetch("BACKEND_URL", "http://localhost:4000")
```
martintomas, line 14: we will probably need to tweak this later on so it uses correct env variable

backend/spec/requests/api/v1/project_developers_spec.rb
```diff
@@ -38,6 +38,91 @@
...
+          about: {type: :string},
+          website: {type: :string},
+          linkedin: {type: :string},
+          facebook: {type: :string},
+          twitter: {type: :string},
+          instagram: {type: :string},
+          project_developer_type: {type: :string, enum: ProjectDeveloperType::TYPES},
+          entity_legal_registration_number: {type: :string},
+          mission: {type: :string},
+          "categories[]": {type: :array, items: {type: :string, enum: Category::TYPES}, collectionFormat: :multi},
```
martintomas, line 62: another rswag bug is here with `collectionFormat ` when used together with `formData `. It allows you to use multiple items from selectbox, but it sends it as `["value-1, value-2"]` <-- one string inside array (https://github.com/rswag/rswag/issues/498). Unfortunately I was not able to fix this, therefore only one item can be picked via rswag during testing of this.

backend/spec/requests/api/v1/project_developers_spec.rb
```diff
@@ -38,6 +38,91 @@
         end
       end
     end
+
+    post "Create new Project Developer" do
+      tags "Project Developers"
+      consumes "multipart/form-data"
```
tsubik, line 44: Hmm, the plan was to use DirectUpload to not use multipart/form-data. Of course, that will be more work on the frontend. Maybe we could do it in another iteration to not focus on details now. We have to set file size limits, and also validate if for example images are actually images (I think active storage analyze using imagemagick or vips is doing that).

backend/app/controllers/api/v1/project_developers_controller.rb
tsubik: Great choice :+1:

backend/spec/requests/api/v1/project_developers_spec.rb
martintomas: I have not thought about direct upload approach, but it is good option. For now I will keep `multipart/form-data`, but I have also added `direct-upload` to rswag documentation. There are not extra changes required if we use direct-upload or multipart/form-data approach now. Later on, we can get rid of multipart/form-data and switch to pure json again if required.

About validation issue, I have created https://vizzuality.atlassian.net/browse/LET-226 to address this, because this is another quite complex topic so lets keep it independent.

backend/app/controllers/api/v1/users/project_developers_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+            current_user.update! account: account, role: :project_developer
+            project_developer = ProjectDeveloper.create! project_developer_params.merge account: account
+            render json: ProjectDeveloperSerializer.new(project_developer).serializable_hash
+          end
+        end
+
+        private
+
+        def account_params
+          params.fetch(:project_developer_params, params)
```
tsubik, line 22: so the frontend will need to add root `project_developer_params`? Would that change things https://api.rubyonrails.org/classes/ActionController/ParamsWrapper.html, in base controller wrap is done for all :json requests

martintomas: Nope, that is basically optional, I have put it here just because of rswag tests, you can use:
```
{ project_developer_params: { DATA } }
or
{ DATA }
```
my guess is that FE will be using just DATA (second option)

martintomas: hmm, but maybe if i rename `project_developer_params ` to `project_developer`, I will not need this hack, will check it :)

martintomas: Soo it could theoretically work but based on documentation
```
On Active Record models with no :include or :exclude option set, it will only wrap the parameters returned by the class method attribute_names.
```
which means that I would need to manually add all translation and active storage attributes, which does not look really nice:
```
wrap_parameters include: Account.attribute_names + %i[mission location_ids about picture]
```
so I don't think it is worthy. Still, FE can use this endpoint without any root element, no need to add `project_developer_params`

tsubik: ok

backend/spec/requests/api/v1/project_developers_spec.rb
agnessa: yes, let's keep validation separate 👍

martintomas:
Anybody brave enough to give me approve @agnessa @tsubik ? :) If you have some change requirements, please don't hesitate to share -- or we can have a call if your feedback is more complex ;)

backend/app/controllers/api/v1/users/project_developers_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+    module Users
+      class ProjectDevelopersController < BaseController
+        skip_before_action :require_json!, only: :create
+        before_action :authenticate_user!
+
+        def create
+          current_user.with_lock do
+            raise API::UnprocessableEntityError, I18n.t("errors.messages.user_multiple_accounts") if current_user.account_id.present?
+
+            account = Account.create! account_params
```
tsubik, line 12: missing set `owner_id` to current_user.id. Account model is missing `has_one :owner`

backend/app/serializers/api/v1/project_developer_serializer.rb
```diff
@@ -4,7 +4,10 @@ class ProjectDeveloperSerializer
       include JSONAPI::Serializer
 
       attributes :name, :slug, :about, :website, :instagram, :facebook, :linkedin, :twitter,
-        :mission, :project_developer_type, :categories, :impacts, :language
+        :mission, :project_developer_type, :categories, :impacts, :language, :picture_url,
+        :entity_legal_registration_number
```
tsubik, line 8: We need `review_status` as well. Public endpoints will return only `approved`

backend/app/controllers/api/v1/users/project_developers_controller.rb
martintomas: Added missing owner, thanks @tsubik ;)

backend/app/serializers/api/v1/project_developer_serializer.rb
martintomas: added review_status to response ;)

tsubik (approved):
LGTM :+1:

agnessa (approved):

PR #53, opened 2022-03-28
Account type selection page

SARodrigues:
This PR adds an account type selection screen, which will be shown to the user after they sign up and verify their email. Responsiveness for mobile devices was implemented as well. 

## Testing instructions

- Respects the [visual design](https://www.figma.com/file/KC49uMR9t1YARmFR0Sr6aN/HeCo---Work-in-progress?node-id=788%3A14933)
- Lets the user choose between the creation of a Project Developer or Investor account
- Accessible by [WCAG 2.1 standards](https://www.w3.org/TR/WCAG21/)
- Available for translation in Spanish, Portuguese and English
- Page works well on desktop devices

## Tracking

[LET-172](https://vizzuality.atlassian.net/browse/LET-172)

## Screenshot
Screenshot 2022-03-28 at 16 01 54


frontend/containers/account-type-selector/account-type-item/component.tsx
```diff
@@ -0,0 +1,68 @@
+import { FC, useState } from 'react';
+
+import { useIntl } from 'react-intl';
+
+import cx from 'classnames';
+
+import Image from 'next/image';
+
+import { usePress, useFocusWithin } from '@react-aria/interactions';
+import noop from 'lodash-es';
```
clementprdhomme, line 10: Are you not importing the whole library when you do that?
```suggestion
import { noop } from 'lodash-es';
```

frontend/containers/account-type-selector/component.tsx
```diff
@@ -0,0 +1,68 @@
+import { FC, useMemo } from 'react';
+
+import { useIntl } from 'react-intl';
+
+import cx from 'classnames';
+
+import noop from 'lodash-es';
```
clementprdhomme, line 7:
```suggestion
import { noop } from 'lodash-es';
```

frontend/containers/account-type-selector/component.tsx
```diff
@@ -0,0 +1,68 @@
...
+      {
+        accountTypeId: 'investor',
+        name: intl.formatMessage({ defaultMessage: 'Investor / Funder', id: 'OChccW' }),
+        description: intl.formatMessage({
+          defaultMessage: 'I want to find opportunities to invest / donate',
+          id: 'nhAnF3',
+        }),
+        imageSrc: '/images/account-types/investor.svg',
+        imageTitle: intl.formatMessage({
+          defaultMessage: 'Drawing of a person looking at a list of card',
```
clementprdhomme, line 29: You can ignore this if you don't agree: 
```suggestion
          defaultMessage: 'Illustration of a person looking at a list of cards',
```

frontend/containers/account-type-selector/component.tsx
```diff
@@ -0,0 +1,68 @@
...
+      {
+        accountTypeId: 'project-developer',
+        name: intl.formatMessage({ defaultMessage: 'Project developer', id: 'yF82he' }),
+        description: intl.formatMessage({
+          defaultMessage: 'I want to find the right investment for my projects',
+          id: 'Ow9MS8',
+        }),
+        imageSrc: '/images/account-types/project-developer.svg',
+        imageTitle: intl.formatMessage({
+          defaultMessage: 'Drawing of a person looking at tree',
```
clementprdhomme, line 42:
```suggestion
          defaultMessage: 'Illustration of a person looking at a tree',
```

frontend/containers/account-type-selector/account-type-item/component.tsx
```diff
@@ -0,0 +1,68 @@
...
+  });
+
+  return (
+    <div
+      className={cx({
+        'flex flex-col items-center gap-3 px-12 py-8 text-center bg-white border rounded-xl': true,
+        'cursor-pointer hover:ring-1 hover:ring-green-dark transition-all': true,
+        'ring-2 ring-green-dark': isFocusWithin,
+      })}
+      role="group"
```
clementprdhomme, line 41: Is that role useful? If so, I'd be curious to read more about it.

frontend/containers/account-type-selector/account-type-item/component.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const { focusWithinProps } = useFocusWithin({
+    onFocusWithinChange: setIsFocusWithin,
+  });
+
+  return (
+    <div
+      className={cx({
+        'flex flex-col items-center gap-3 px-12 py-8 text-center bg-white border rounded-xl': true,
+        'cursor-pointer hover:ring-1 hover:ring-green-dark transition-all': true,
+        'ring-2 ring-green-dark': isFocusWithin,
```
clementprdhomme, line 39: I don't think you need `useFocusWithin`:
```suggestion
        'focus-within:ring-2 ring-green-dark': true,
```

frontend/containers/account-type-selector/account-type-item/component.tsx
```diff
@@ -0,0 +1,68 @@
...
+import cx from 'classnames';
+
+import Image from 'next/image';
+
+import { usePress, useFocusWithin } from '@react-aria/interactions';
+import noop from 'lodash-es';
+
+import { AccountTypeItemProps } from './types';
+
+export const AccountTypeItem: FC<AccountTypeItemProps> = ({
```
clementprdhomme, line 14: I like the way you implemented this component (from the accessibility point of view). Did you use the Inclusive Components article for the card?

frontend/containers/account-type-selector/account-type-item/component.tsx
SARodrigues: When navigating with VoiceOver, it makes it so that the elements inside are considered part of a group. In this case, the goal was so that someone navigating using a screen reader would be aware that the illustration, button and description inside this element are all part of the same "group" of elements (which in this case a card). 

However, I tested it again with VoiceOver and it turns out it's not needed. By setting the `aria-label` VoiceOver does it automatically (does announce the card's items as a group), so I removed the role.

frontend/containers/account-type-selector/account-type-item/component.tsx
SARodrigues: I did, yes!  🙂

frontend/containers/account-type-selector/account-type-item/component.tsx
clementprdhomme: Thanks for the explanation!

barbara-chaves (approved):

frontend/containers/account-type-selector/account-type-item/component.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  const intl = useIntl();
+
+  const { pressProps } = usePress({
+    onPress: () => onClick(accountTypeId),
+  });
+
+  const { focusWithinProps } = useFocusWithin({
+    onFocusWithinChange: setIsFocusWithin,
+  });
```
clementprdhomme, line 32: You don't need this anymore neither `isFocusWithin`.

clementprdhomme (approved):

PR #55, opened 2022-03-29
Backend reset passwords controller

tsubik:
Reset passwords controller.

```
POST /api/v1/reset_password  (param email)
```

Sends email to user with reset password token. Currently, I'm just showing token in email, and not sending any link, because there are no frontend pages to display form to change password.

```
PATCH/PUT /api/v1/reset_password  (params password, token)
```

Route to change password, requires token and new password.

## Testing instructions

Use API docs.

- post reset password for your user
- get token from email
- put the token with new password to update reset password route

## Tracking

https://vizzuality.atlassian.net/browse/LET-189
https://vizzuality.atlassian.net/browse/LET-191

### DIFF
--- a/backend/spec/requests/api/v1/reset_password_spec.rb
+++ b/backend/spec/requests/api/v1/reset_password_spec.rb
@@ -23,6 +23,7 @@ RSpec.describe "API V1 Reset Password", type: :request do
 
         run_test!
 
+        # head ok so no generate example
         it "sends email" do
           mail = ActionMailer::Base.deliveries.last
           expect(mail.subject).to eq("Reset password instructions")
@@ -82,7 +83,7 @@ RSpec.describe "API V1 Reset Password", type: :request do
 
         run_test!
 
-        it "changes user password and signs in user" do
+        it "changes user password and signs in user", generate_swagger_example: true do
           user.reload
           expect(user.valid_password?("NewPassword1234")).to eq(true)
           expect(user.reset_password_token).to be_nil
@@ -104,7 +105,7 @@ RSpec.describe "API V1 Reset Password", type: :request do
         context "no token" do
           let(:params) { {} }
 
-          it "returns correct error", generate_swagger_example: true do
+          it "returns correct error" do
             expect(response_json["errors"][0]["title"]).to eq("Reset password token can't be blank")
           end
         end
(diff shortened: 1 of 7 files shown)


backend/spec/requests/api/v1/reset_password_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+        },
+        required: ["email", "reset_password_token"]
+      }
+
+      response "200", :success do
+        schema type: :object, properties: {
+          data: {"$ref" => "#/components/schemas/user"}
+        }
+        let(:token) {
+          user.send(:set_reset_password_token) # this is protected method
```
martintomas, line 78: maybe there is some magic happening inside this method, but is it not more readable when use set token manually?
```
create :user, reset_password_token: token, reset_password_sent_at: Time.current
```

backend/spec/requests/api/v1/reset_password_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+          end
+        end
+
+        context "invalid password" do
+          let(:token) {
+            user.send(:set_reset_password_token) # this is protected method
+          }
+          let(:params) { {password: "secret", reset_password_token: token} }
+
+          it "returns correct error", generate_swagger_example: true do
```
martintomas, line 131: In reality `generate_swagger_example: true` is used only once per response type so even when you use it for multiple tests, it will pick just one example and save if for docs

backend/spec/requests/api/v1/reset_password_spec.rb
tsubik: Not sure if that will work, reset_password_token is different than confirmation_token (dunno why, some devise inconsistency). It is encrypted in db, but I can check if it works.

backend/spec/requests/api/v1/reset_password_spec.rb
tsubik: ahh, upps

backend/spec/requests/api/v1/reset_password_spec.rb
tsubik: I think I will leave that protected method. Otherwise I have to generate by hand like this https://mokacoding.com/blog/how-to-manually-generate-devise-reset-password-link/ and save hashed in db, and send raw as param.

martintomas (approved):
Looks good to me ;)

PR #56, opened 2022-03-29
feat: API endpoint for updating project developer

martintomas:
New API endpoint which allows to update Project Developer information.

Screenshot 2022-03-29 at 11 06 59

## Testing instructions

Open Rswag documentation and try to update project developer of user which is already project developer. Don't forget to check out `Send empty value` for fields which you don't wanna update ;).

## Tracking

https://vizzuality.atlassian.net/browse/LET-232


backend/app/controllers/api/v1/accounts/project_developers_controller.rb
```diff
@@ -16,11 +15,23 @@ def create
           end
         end
 
+        def update
+          if current_user.role.to_sym != :project_developer
+            raise API::UnprocessableEntityError, I18n.t("errors.messages.user.no_project_developer")
+          end
+
+          current_user.account.update! account_params.except(:language)
```
martintomas, line 23: Language can be update only via Admin so I have move this out here

backend/app/controllers/api/v1/accounts/project_developers_controller.rb
```diff
@@ -16,11 +15,23 @@ def create
...
+          render json: ProjectDeveloperSerializer.new(current_user.account.project_developer).serializable_hash
+        end
+
         private
 
         def account_params
-          params.fetch(:project_developer_params, params)
+          data = params.fetch(:project_developer_params, params)
             .permit :language, :picture, :name, :website, :linkedin, :facebook, :twitter, :instagram, :about
+          data = data.except(:picture) if data[:picture].blank?
```
martintomas, line 33: This is here to just prevent `ActiveSupport::MessageVerifier::InvalidSignature` in case that you send empty string via rswag

backend/app/controllers/api/v1/accounts/base_controller.rb
```diff
@@ -0,0 +1,9 @@
+module API
+  module V1
+    module Accounts
+      class BaseController < API::V1::BaseController
```
tsubik, line 4: :+1:

backend/app/controllers/api/v1/accounts/project_developers_controller.rb
tsubik: you can add `before_filter :ensure_project_developer!` or `require_project_developer!` or maybe we should add authorization layer.

```
def ensure_project_developer!
  return if current_user.project_developer?   

  raise API::UnprocessableEntityError, I18n.t("errors.messages.user.no_project_developer")
end
```

backend/app/controllers/api/v1/accounts/project_developers_controller.rb
tsubik: Hmm, maybe a comment to remove it if we remove multipart form? Normally if no one wants to update picture it won't be present in body.

backend/spec/requests/api/v1/accounts/project_developers_spec.rb
```diff
@@ -82,7 +82,105 @@
...
+      let(:project_developer_params) do
+        {
+          picture: fixture_file_upload("picture.jpg"),
+          name: "Name",
+          about: "About",
+          website: "http://website.com",
+          linkedin: "http://linkedin.com",
+          facebook: "http://facebook.com",
+          twitter: "http://twitter.com",
+          instagram: "http://instagram.com",
```
tsubik, line 127: I wanted to say to add language param here and check that it was not updated, but that will land in example, so not good.

martintomas: I will test it at different context, so that should be fine, good point ;)

backend/app/controllers/api/v1/accounts/project_developers_controller.rb
martintomas: If we will remember to checkout `Send empty values` for `picture` all the time when using rswag, I can remove this, np

backend/app/controllers/api/v1/accounts/project_developers_controller.rb
martintomas: Like it, will do it ;)

backend/app/controllers/api/v1/accounts/project_developers_controller.rb
tsubik: Yeah, then maybe remove it, as this is not code related, but more rswag and api docs related.

backend/app/controllers/api/v1/accounts/project_developers_controller.rb
```diff
@@ -2,12 +2,12 @@ module API
   module V1
     module Accounts
       class ProjectDevelopersController < BaseController
-        skip_before_action :require_json!, only: :create
-        before_action :authenticate_user!
+        skip_before_action :require_json!, only: %i[create update]
+        before_action :require_project_developer!, except: :create
```
tsubik, line 6: I would whitelist, especially if there is only one action to whitelist.

tsubik (approved):
All good :+1:

backend/app/controllers/api/v1/accounts/project_developers_controller.rb
tsubik: But that is just opinion ;]

tsubik: Or maybe, no, you are right, better to require by default here :+1: and use `except`

agnessa (approved):

PR #58, opened 2022-03-29
feat: Validate images with help of mini_magick

martintomas:
Making sure that uploaded pictures are truly images. API returns url of images only if
 - image was analyzed with AnalyzeJob
 - image contains dimensions --> therefore it is image
 
Till we are sure that image is truly image, we return `null` value.

## Testing instructions
Testing is a bit tricky till there is no way how to run background jobs, but basically it can be done with:
- create or update picture of project developer via swagger doc
- check that null is returned because image is not analyzed yet
- open console and manually run  `ActiveStorage:: AnalyzeJob.perform_now(YOUR_BLOB)`
- check that project developer info returned by API contains picture urls now

## Tracking

https://vizzuality.atlassian.net/browse/LET-226


tsubik:
Hey @martintomas I was thinking that maybe we don't need to add anything extra and use default or add another analyzer. https://edgeapi.rubyonrails.org/classes/ActiveStorage/Blob/Analyzable.html

tsubik:
We could return placeholder until blob is `analyzed?`

tsubik:
I think when we have no dimensions after rails image analysis, then the image could be incorrect, or just not supported. https://github.com/rails/rails/blob/de53ba56cab69fb9707785a397a59ac4aaee9d6f/activestorage/lib/active_storage/analyzer/image_analyzer/image_magick.rb#L12

martintomas:
Hmm, that is actually good idea. I know that you have mentioned similar approach, but I have not connected all dots till now :). I will try to use your proposal --> correct image should have dimensions after analyzer is done. Looks better then have extra job for checking this. Could you be a bit more specific about `We could return placeholder until blob` @tsubik ? I think there are several approaches how to return placeholder:
 - return null and FE deals with the placeholder --> this way FE know that picture needs to be replaced by placeholder and can do some other stuff (like not make image clickable, or whatever)
 - save image at public folder
 - save image at Rails assets --> we will have just one `variant` of placeholder image
 - attach placeholder image as ActiveStorage blob --> allows us to have blob variants, but not sure how to do it :)

tsubik:
Ok. I think return `null` as you suggested is the easiest option.

martintomas:
I have refactored code based on your feedback @tsubik and I am using `ActiveStorage::AnalyzeJob` now. Looks so simple now ;)

tsubik:
Ahh, one thing that is missing. Please add `imagemagick \` in Dockerfile

backend/app/services/blobs/get_link.rb
```diff
@@ -18,5 +18,9 @@ def call
     def modify(original_blob)
       options.present? ? original_blob.variant(options) : original_blob
     end
+
+    def validated?
+      original_blob.analyzed? && original_blob.metadata[:width] && original_blob.metadata[:height]
```
tsubik, line 23: Yeah, it looks simple :)

tsubik (approved):
:rocket: 

You could also update initial PR description for future reference. What do you think?

PR #59, opened 2022-03-29
refactoring: minor tweaking of email confirmation tests

martintomas:
I have just removed logic from `let` variables inside email confirmation tests

### DIFF
--- a/backend/spec/requests/api/v1/email_confirmation_spec.rb
+++ b/backend/spec/requests/api/v1/email_confirmation_spec.rb
@@ -80,37 +80,43 @@ RSpec.describe "API V1 Email Confirmation", type: :request do
       end
 
       response "422", "Invalid Token" do
-        let(:confirmation_token) { "invalid_token" }
+        context "when confirmation token is invalid" do
+          let(:confirmation_token) { "invalid_token" }
 
-        run_test!
+          run_test!
 
-        it "returns correct error", generate_swagger_example: true do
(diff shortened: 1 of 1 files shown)


backend/spec/requests/api/v1/email_confirmation_spec.rb
```diff
@@ -80,37 +80,43 @@
...
 
         context "already confirmed" do
-          let(:confirmation_token) do
-            unconfirmed_user.confirm # not sure why it does not work in before block
-            unconfirmed_user.confirmation_token
-          end
+          let(:confirmation_token) { unconfirmed_user.confirmation_token }
 
-          it "returns correct error", generate_swagger_example: true do
+          before { unconfirmed_user.confirm }
```
tsubik, line 98: and now it works?

tsubik: So, we should always, run_test in separate contexts?

tsubik (approved):

backend/spec/requests/api/v1/email_confirmation_spec.rb
tsubik: or just only when we have problems with `before`.

tsubik: Yeah, but going back to previous version and adding `before` and things not working ;] haha

PR #60, opened 2022-03-30
fix review status and project status enum values

tsubik:
Fixing wrong enum values, should start from `0` as this is default value in DB as well.

As we don't care about the current data. All is good.

## Testing instructions

What to test? How to do it?

## Tracking

No tracking


tsubik:
Also, updating all backend dependencies.

martintomas (approved):

PR #61, opened 2022-03-30
Backend Configure rails app to deploy to subdirectory

tsubik:
I believe that we have to configure rails app to run in subdirectory

https://guides.rubyonrails.org/configuring.html#config-action-controller-relative-url-root
https://gist.github.com/ebeigarts/5450422

I would leave setting env variables in infrastructure to @tiagojsag 

## Testing instructions

Deploy and check if it works

## Tracking

No tracking


tiagojsag:
you can add it here: https://github.com/Vizzuality/heco-invest/blob/9458713178e19b99555283d22f31406225506278/infrastructure/base/modules/env/main.tf#L126

Let me know once it's in place so I can deploy the change

tsubik:
Ok. I think that will be needed in Dockerfile as well, as we have `assets:precompile` after adding backoffice.

tsubik:
But maybe that should be added there when adding backoffice.

martintomas (approved):
Tried, works for me on my local machine.

agnessa (approved):

tsubik:
@tiagojsag I think that is ready to deploy. Didn't change Dockerfile as we don't do assets:precompile for now.

PR #64, opened 2022-03-30
feat: Adding /backend url to Swagger doc

martintomas:
Just simple tweak which ensures that `/backend` url endpoint is first at list of available servers at rswag doc so you don't need to manually specify url yourself.

On local machine, you will need to switch to localhost on your own or start your rails server with `RAILS_RELATIVE_URL_ROOT=/backend`

Screenshot 2022-03-30 at 14 45 39


tsubik (approved):
LGTM :+1:

PR #66, opened 2022-03-30
Let 174 project developer form

barbara-chaves:
### Add the project developer setup page
Create page with form using multi-page layout
Add validations
Add service

## Testing instructions

* Respects the [visual design](https://www.figma.com/file/KC49uMR9t1YARmFR0Sr6aN/HeCo---Work-in-progress?node-id=70%3A14667)

* Contains all the fields defined in the “Forms” tab of the [requirements document](https://docs.google.com/spreadsheets/d/1qtP-qkm8iS6OlluiJwRvQoi6rCns9iKhRYoNzc9qkGg/edit#gid=1884416048)

* The fields are validated:

  * required fields are not empty

* Errors are displayed if:

  * the validation does not pass

  * the API doesn't respond with success

* The fields are part of a form

* Input and form errors are correctly presented to assistive technologies (AT) ([see this guide on form validation](https://webaim.org/techniques/formvalidation/))

* The correct locale is set on the inputs ([see documentation of the lang attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang)) if it’s different from the page’s

* Available for translation in Spanish, Portuguese and English

* Page works well on desktop devices

## Tracking

[LET-174](https://vizzuality.atlassian.net/browse/LET-174)


SARodrigues (changes requested):

SARodrigues:
One thing I noticed about the Figma designs CSS code comments and the actually tailwind classes we're using, is that the red values don't match. The inputs/errors colors described as `red-600` in Figma, visually seem to be `red-700`. I think we should use the latter as it's the one that matches the designs visually. Not sure how that discrepancy came to be though, as for other colors they seem to match.

frontend/pages/project-developer/new.tsx
```diff
@@ -0,0 +1,575 @@
...
+                      defaultMessage: 'insert the profile name',
+                      id: '0WHWA/',
+                    })}
+                    aria-describedby="profile-error"
+                  />
+                </Label>
+                <ErrorMessage id="profile-error" errorText={errors?.profile?.message} />
+              </div>
+              <div className="md:w-1/2">
+                <Label htmlFor="project-developer-type" id="project-developer-type-label">
```
SARodrigues, line 339: Screenshot 2022-03-31 at 13 07 16

frontend/pages/project-developer/new.tsx
```diff
@@ -0,0 +1,575 @@
+import { ChangeEvent, useState, useCallback } from 'react';
```
SARodrigues, line 1: Should this file be under `project-developers/new` instead, following the `investors/new` pattern instead?  

The problem I see with the current slug is that (although unlikely) if a project developer's name is `New` and they end up with the `new` slug, either their profile will be inaccessible or this form will.

frontend/lang/compiled/zu.json
```diff
@@ -1 +1,160 @@
-{}
```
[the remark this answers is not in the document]
clementprdhomme: You're right: https://github.com/Vizzuality/heco-invest/tree/develop/frontend#pipeline. Maybe we could update `.gitignore` to avoid this in the future? I know I also pushed the `compiled` folder unintentionally in the past.

frontend/pages/project-developers/new.tsx
```diff
@@ -0,0 +1,502 @@
...
+              </h1>
+              <p className="font-sans text-base text-gray-600">
+                <FormattedMessage defaultMessage="Tell us about your work." id="Y6xIpg" />
+              </p>
+            </div>
+            {interests.map(({ name, title, items, infoText, required }) => (
+              <div key={name} className="mb-7">
+                <p className="font-sans font-semibold text-sm text-gray-800 mb-4.5 mr-2.5 inline-block">
+                  {title}
+                </p>
```
clementprdhomme, line 445: This the the label of the checkboxes. The current input is not accessible.

frontend/pages/project-developers/pending.tsx
```diff
@@ -0,0 +1,74 @@
...
+          aria-hidden="true"
+          src="/images/pending-approval.svg"
+          alt=""
+        />
+        <h1 className="mt-6 font-serif text-3xl font-semibold text-green-dark">
+          <FormattedMessage defaultMessage="Pending approval" id="/CaREm" />
+        </h1>
+        <p className="my-6">
+          <FormattedMessage
+            defaultMessage="Your account was successfully created, but you steel need to be approved by our platform administrators."
```
clementprdhomme, line 43:
```suggestion
            defaultMessage="Your account was successfully created, but you still need to be approved by our platform administrators."
```

frontend/pages/project-developers/pending.tsx
```diff
@@ -0,0 +1,74 @@
...
+        </p>
+        <p>
+          <FormattedMessage
+            defaultMessage="Until approval you can continue exploring our database."
+            id="adPUXO"
+          />
+        </p>
+        <Button className="mt-6 mb-12">
+          <FormattedMessage defaultMessage="Explore" id="7JlauX" />
+        </Button>
```
clementprdhomme, line 55: Can you transform this into a link that goes to `/discover`?

frontend/services/api.ts
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * API service require to be authenticated.
+ * Then, by default all the request are sending the authorization header.
+ */
+
+const dataFormatter = new Jsona();
+
+const defaultConfig = {
+  baseURL: `${process.env.NEXT_PUBLIC_API_URL || process.env.STORYBOOK_API_URL}/api/v1/`,
```
clementprdhomme, line 13: `NEXT_PUBLIC_API_URL` is already defined as `/backend/api/v1` in `.env.sample` so I don't think you should add `/api/v1/` here at the end.

clementprdhomme: This will need to change once we merge #76 because of the new environment variables.

clementprdhomme (approved):

clementprdhomme:
@barbara-chaves Let's merge #76 first and then rebase this PR.

Great job 👏!

PR #71, opened 2022-03-31
LET-183 SMTP config

agnessa:
Adds smtp config to the Rails app and to the tf sorcery

## Testing instructions

In development I left the letter_opener as mail delivery method, but I did test with the sendgrid account. It is a temporary account which sends emails from my address. Hopefully at some point next week we'll just need to update the password to use the target one.

## Tracking

https://vizzuality.atlassian.net/browse/LET-183?atlOrigin=eyJpIjoiYmFiN2ZmMTFlYWQ3NGJkMmE3YjZiZmE2NmUzYWIwZTkiLCJwIjoiaiJ9

### DIFF
--- a/infrastructure/base/modules/env/main.tf
+++ b/infrastructure/base/modules/env/main.tf
@@ -41,6 +41,14 @@ module "transifex_api_key" {
   use_random_value = false
 }
 
+module "sendgrid_api_key" {
+  source           = "../secret_value"
+  region           = var.gcp_region
+  key              = "${var.project_name}_sendgrid_api_key"
+  value            = var.sendgrid_api_key
+  use_random_value = false
+}
+
 locals {
   frontend_docker_build_args = {
     TRANSIFEX_TOKEN = var.transifex_token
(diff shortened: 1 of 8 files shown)


infrastructure/base/modules/env/main.tf
```diff
@@ -161,6 +172,26 @@ module "backend_cloudrun" {
...
+      name  = "SMTP_PORT"
+      value = 587
+    },
+    {
+      name  = "MAILER_DEFAULT_HOST"
+      value = var.domain
+    },
+    {
+      name  = "MAILER_DEFAULT_FROM"
+      value = "agnieszka.figiel@vizzuality.com"
```
tiagojsag, line 194: I'd really encourage you to not do this. This seems like the perfect way for this to be forgotten and left like this forever, so I'd suggest you go for something like "noreply@${var.domain}"

agnessa: This needs to be a verified email address (confirmed), and I already verified mine, not keen on spending time sorting out any other email. I will chase the client to replace with actual sender email next week.

PR #73, opened 2022-04-01
LET-249 - Mitigate cold starts on production CloudRun

tiagojsag:
This has been applied, so you can validate how it looks like live


agnessa (approved):

agnessa:
awesome, thanks!

PR #76, opened 2022-04-04
Let 211 connect the sign up form with the api

barbara-chaves:
This PR adds the connection between the sign-up form and the API

## Testing instructions

* The sign up data is sent and registered in the API

* Errors are displayed if:

  * the API responds with validation errors
  * the API does not respond

* After a sign-up success, the user is directed to the account creation flow

* The errors are presented assisted technologies
## Tracking

[LET-211](https://vizzuality.atlassian.net/browse/LET-211)


clementprdhomme:
@barbara-chaves If that's ok with you, I renamed the environment variables (and updated their descriptions) so it's a bit more coherent.

@tiagojsag We've changed the environment variables. You can see the list of them on the [README file](https://github.com/Vizzuality/heco-invest/blob/8ff878ffadc52b94d09c7a601326104cbb9c722c/frontend/README.md#environment-variables). What do we need to update for the staging and production deployment? Their values are stored on LastPass.

frontend/.env.sample
```diff
@@ -1,5 +1,5 @@
-NEXT_PUBLIC_API_URL=""
-NEXTAUTH_URL=http://localhost:3000
+NEXT_PUBLIC_DOMAIN=https://staging.heco.vizzuality.com
+NEXT_PUBLIC_BACKEND_DOMAIN=https://staging.heco.vizzuality.com
+NEXT_PUBLIC_API_PATH=/backend/api/v1
```
tiagojsag, line 3: I would strongly encourage you to refactor your code, so that this value has `/backend` in it, and everything else is part of the endpoint subpath, not the baseUrl:
- /api is part of the method signature, not the URL of the API. The backend can choose to implement something at https://staging.heco.vizzuality.com/backend/auth (for example) which is valid and would break your implementation
- /v1 is evil because if the api decides to add a /v2 set of endpoints, you now have to add yet another variable.

frontend/next.config.js
```diff
@@ -43,4 +43,16 @@ module.exports = {
...
     return config;
   },
+  async rewrites() {
+    if (process.env.PROXY_BACKEND !== 'true') return [];
+
+    const backend = process.env.NEXT_PUBLIC_BACKEND_DOMAIN;
+
+    return [
+      {
+        source: '/backend/:path*',
```
tiagojsag, line 53: evil hardcoded value that should be a constant
also: evil redirect that the FE should not know nor care about

frontend/services/api.ts
```diff
@@ -0,0 +1,41 @@
+import axios from 'axios';
+import Jsona from 'jsona';
+
+const dataFormatter = new Jsona();
+
+const isServer = typeof window === 'undefined';
+const baseUrl = isServer
```
tiagojsag, line 7: evil!!!!!
Use a single var, and change it according to the env you're in

frontend/.env.sample
tiagojsag, line 3:
```suggestion
NEXT_PUBLIC_FRONTEND_URL=https://staging.heco.vizzuality.com
NEXT_PUBLIC_BACKEND_URL=https://staging.heco.vizzuality.com/backend
```

frontend/services/api.ts
```diff
@@ -0,0 +1,41 @@
...
+  ? process.env.NEXT_PUBLIC_BACKEND_DOMAIN + process.env.NEXT_PUBLIC_API_PATH
+  : process.env.NEXT_PUBLIC_API_PATH;
```
tiagojsag, line 9:
```suggestion
const baseUrl = process.env.NEXT_PUBLIC_BACKEND_URL;
```

frontend/services/authentication/authService.ts
```diff
@@ -0,0 +1,37 @@
...
+import { SignInDto } from 'types/signin';
+
+import API from '../api';
+
+export function signIn() {}
+
+export async function signOut() {
+  await API.request({
+    method: 'DELETE',
+    url: '/session',
```
tiagojsag, line 16:
```suggestion
    url: '/api/v1/session',
```

tiagojsag: Also: check what usage axios makes of uri vs url, when combined with a basepath. IIRC there is some difference, but I'm, not sure

tiagojsag (changes requested):
some suggestions

clementprdhomme:
@barbara-chaves When testing the sign up, I get this error:
>Can't verify CSRF token authenticity.

Am I missing something?

tsubik:
@clementprdhomme This is because you didn't have a csrf cookie. There must be any even GET request to the server setting that cookie before doing non-GET request (like sign up). I believe sign-up page will be eventually accessible only when there is no logged in user (otherwise redirect somewhere else). So there will be check for logged in user before (GET /user), then it should work. For now, you can just add `const { user } = useMe()` on the frontend to mitigate this.

clementprdhomme:
@tiagojsag We're ready to merge. Do you want to make the environment variables changes? You can find them on Lastpass.

clementprdhomme (approved):

PR #79, opened 2022-04-05
Enable log to stdout on prod

tsubik:
I don't see any error logs details, so I suspect this must be configured. Is that correct @tiagojsag ? Those logs I don't see in GCP https://github.com/Vizzuality/heco-invest/blob/develop/backend/app/controllers/concerns/api/errors.rb#L43

## Testing instructions

What to test? How to do it?

## Tracking

Link to the task(s), if any


tsubik:
Or maybe without that env variable. Just enable this by default.

agnessa:
I guess this might help, although I have no understanding how log streaming is configured.

tsubik:
Ok. So we need this. I will remove env variable and enable that by default.

PR #81, opened 2022-04-05
[Frontend] Add missing docker args

tsubik:
Missing build args.

## Testing instructions

What to test? How to do it?

## Tracking

No tracking


infrastructure/base/modules/env/main.tf
```diff
@@ -56,7 +56,6 @@ locals {
     NEXT_PUBLIC_FRONTEND_URL = "https://${var.domain}"
     NEXT_PUBLIC_BACKEND_URL = "https://${var.domain}/backend"
     NEXT_PUBLIC_GOOGLE_ANALYTICS = var.google_analytics_key
-    NEXT_PUBLIC_PROXY_BACKEND = "false"
```
tsubik, line 59: I think this could be removed, as will never be needed. Maybe that env variable is not needed at all. We could enable proxy whenever frontend_url domain is different from backend_url domain.

tiagojsag: that kind of magic logic hidden away in the code sounds like a great time sink for us-in-6-months. keep the var, it makes the behavior explicit

PR #87, opened 2022-04-12
LET-288 contact email & phone

agnessa:
https://vizzuality.atlassian.net/browse/LET-288?atlOrigin=eyJpIjoiZmJlMTY4OGJhM2M5NDEyOWFkOTcyNDUyOWI0ZmZlMDkiLCJwIjoiaiJ9

### DIFF
--- a/backend/app/models/account.rb
+++ b/backend/app/models/account.rb
@@ -23,6 +23,7 @@ class Account < ApplicationRecord
   validates :language, inclusion: {in: Language::TYPES}
   validates :picture, attached: true, content_type: /\Aimage\/.*\z/
   validates :contact_email, presence: true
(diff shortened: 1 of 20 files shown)


backend/app/models/account.rb
```diff
@@ -22,6 +22,7 @@ class Account < ApplicationRecord
   validates :instagram, url: true
   validates :language, inclusion: {in: Language::TYPES}
   validates :picture, attached: true, content_type: /\Aimage\/.*\z/
+  validates :contact_email, presence: true
```
tsubik, line 25: Do we need email validation? I think devise is doing that with regex, but I'm not sure.

tsubik: `Devise.email_regexp`

agnessa: sure, added!

tsubik (approved):
:+1:

PR #91, opened 2022-04-14
Remove the native validation from the forms

clementprdhomme:
React Hook Form is used with a resolver (Yup) to validate the form. Unfortunately, RHF is not able to validate individual fields when using one. The whole schema needs to be validated to check the validity of just one field.

As long as the user is the initiator of the change to a field's value, validation will be triggered on this field only. Nevertheless, if a field's value is updated programmatically and it needs to be validated, then the whole form will be validated and display errors.

This is the case of the `<TagGroup />` component: the “Select All” button needs to update the field's value and we also need to trigger validation on that field to remove the red ring (border). The problem is that if the user clicks the “Select All” button, then the whole form will be validated and errors will be displayed next to fields the user hasn't even had a look at yet. You can see this in action when replicating [Susana's steps](https://vizzuality.atlassian.net/browse/LET-276?focusedCommentId=15782).

The solution proposed by this PR is imperfect but I believe it is the simplest to fix the issue and to keep the fields' validation updated (add/remove the red ring): remove the native validation.

A better solution would be to pass `control` and `controlOptions` props to each of the inputs (like for the `<Select />` component) so that we would have a standardised way to deal with inputs and they could read their validation status by themselves instead of relying on the parent to pass it.

## Testing instructions

1. Go to the PD form
2. On the 3rd step, click the “Select All” button of the first field

Make sure errors are not triggered elsewhere in the form.

## Tracking

Based on an issue reported on the user story [LET-276](https://vizzuality.atlassian.net/browse/LET-276). This PR blocks the acceptance of the story.

### DIFF
--- a/frontend/components/forms/checkbox/component.stories.tsx
+++ b/frontend/components/forms/checkbox/component.stories.tsx
@@ -16,6 +16,7 @@ export default {
   title: 'Components/Forms/Checkbox',
   argTypes: {
     register: { control: { disable: true } },
+    invalid: { control: { disable: true } },
   },
 } as Meta;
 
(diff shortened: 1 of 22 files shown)


SARodrigues (changes requested):
Please see: https://vizzuality.atlassian.net/browse/LET-276?focusedCommentId=15786

clementprdhomme:
Replaced by #99.

PR #92, opened 2022-04-14
Feature/project creation endpoint

tsubik:
Project creation endpoint

`POST /backend/api/v1/account/projects`

Like project developer I use `multipart/form-data` here as there will be gallery and Shapefile upload add later. Of course, when frontend switches to DirectUpload we will switch to use only `json`.

## Testing instructions

Test with API docs

## Tracking

https://vizzuality.atlassian.net/browse/LET-280

### DIFF
--- a/backend/db/seeds.rb
+++ b/backend/db/seeds.rb
@@ -7,11 +7,8 @@ if Rails.env.development?
   User.delete_all
 
   5.times do
-    user = FactoryBot.create(:user)
-    investor_account = FactoryBot.create(:account, owner: user)
-    user.update!(account: investor_account, role: "investor")
-
-    investor_account.update!(owner: user)
+    investor_account = FactoryBot.create(:account)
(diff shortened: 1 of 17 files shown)


backend/db/seeds.rb
```diff
@@ -7,8 +7,11 @@
   User.delete_all
 
   5.times do
-    investor_account = FactoryBot.create(:account)
-    FactoryBot.create(:user, account: investor_account)
+    user = FactoryBot.create(:user)
+    investor_account = FactoryBot.create(:account, owner: user)
+    user.update!(account: investor_account, role: "investor")
+
+    investor_account.update!(owner: user)
```
agnessa, line 14: do we need this if the account was created with this owner?

agnessa (approved):

backend/db/seeds.rb
tsubik: Right, we don't need this.

PR #94, opened 2022-04-14
Backend/Add missing category

tsubik:
Adding missing category. Not sure if we have a description somewhere @agnessa ?

Test snapshot was weird, but ids are normally IGNORED and not compared, so it worked ;] 

## Testing instructions

api docs

## Tracking

No tracking


agnessa:
we don't have the description yet

tsubik:
We can add description in another PR.

martintomas (approved):

PR #99, opened 2022-04-20
Fixing Tag/TagGroup validation errors & keyboard focus ring

SARodrigues:
**Description:**  

Fix for [LET-170](https://vizzuality.atlassian.net/browse/LET-170)'s current issues with Tag/TagGroup error display as well as keyboard focus rings. 

## Testing instructions

Verify that the issues described [here](https://vizzuality.atlassian.net/browse/LET-276?focusedCommentId=15782) no longer occur:  

1. Choosing tags works
2. The "Select All" button works
3. Clicking "Select All" on the categories tags section will not trigger validation on the impacts tags section
4. Clicking "Submit" does not trigger a missing email error  

Also:  

1. Rings appear around the tags when navigating via keyboard "Tab"  
2. Rings do not appear around the tags when clicking on them 

## Tracking

[LET-170](https://vizzuality.atlassian.net/browse/LET-170)

### DIFF
--- a/frontend/components/forms/tag-group/component.stories.tsx
+++ b/frontend/components/forms/tag-group/component.stories.tsx
@@ -27,6 +27,7 @@ const Template: Story<TagProps<FormValues>> = (args: TagProps<FormValues>) => {
     register,
     setValue,
     clearErrors,
+    watch,
     formState: { errors },
   } = useForm<FormValues>();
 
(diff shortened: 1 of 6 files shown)


SARodrigues:
@clementprdhomme Although it _mostly_ works on Storybook as well, it's not perfect. Try selecting and deselecting any tag that's not the first one; a native validation will pop up "Please tick this box", although the form can still submit and the error borders are shown correctly. 

Not sure what I'm missing here though. The behaviour does not occur when the component is used; I suspect it may have something to do because we're not using `yup` on storybook.

clementprdhomme (approved):

PR #102, opened 2022-04-21
feat: allow to switch between Jobs and API/Backoffice routes

martintomas:
I am adding ability to enable/disable appropriate set of routes based on ENV variables. Current routes groups are:
 - IS_API_INSTANCE --> enables API and Backoffice routes
 - IS_JOBS_INSTANCE --> enables JOBS routes

Please note that route `post :test, to: ->(_env) { [200, {}, [""]] }` is here just for testing purposes, because I wanted to be able to test that everything works. This one will be removed after we have some logic at `jobs` controllers.

## Testing instructions

There is already test for this, but basically set `IS_JOBS_INSTANCE = true` and check that `/jobs/test` returns `success`

## Tracking

https://vizzuality.atlassian.net/browse/LET-259


backend/config/routes.rb
```diff
@@ -1,30 +1,10 @@
...
-
-      resources :enums, only: [:index]
-
-      scope "account", module: "accounts" do
-        resource :project_developer, only: [:create, :update, :show]
-        resources :projects, only: [:create]
-      end
-    end
+  if ENV["JOBS_INSTANCE"]
+    draw :jobs
```
tsubik, line 5: I didn't know it could be split in such a elegant way.

tiagojsag (commented):
I would prefer something like a "INSTANCE_TYPE" with a string value, that could be either "API" or "JOBS". Makes it easier to extend and understand (these are var name/value examples, feel free to modify them)

Also: is there a use case (for example local dev) where we would want the same app instance to support both things? in which case, 2 separate IS_API_INSTANCE and IS_JOBS_INSTANCE boolean env vars would perhaps make more sense

martintomas:
> I would prefer something like a "INSTANCE_TYPE" with a string value, that could be either "API" or "JOBS". Makes it easier to extend and understand (these are var name/value examples, feel free to modify them)
> 
> Also: is there a use case (for example local dev) where we would want the same app instance to support both things? in which case, 2 separate IS_API_INSTANCE and IS_JOBS_INSTANCE boolean env vars would perhaps make more sense

Thanks @tiagojsag , that is good point. I find it quite useful to enable API and JOBS routes at same time for develop or test environments  so I have picked your second example and added `IS_API_INSTANCE` and `IS_JOBS_INSTANCE` variables so we can freely control which routes are available where.

backend/config/routes.rb
```diff
@@ -1,30 +1,7 @@
...
-
-      scope "account", module: "accounts" do
-        resource :project_developer, only: [:create, :update, :show]
-        resources :projects, only: [:create]
-      end
-    end
-  end
+  draw :backoffice if ENV["IS_API_INSTANCE"]
+  draw :api if ENV["IS_API_INSTANCE"]
+  draw :jobs if ENV["IS_JOBS_INSTANCE"]
```
agnessa, line 6: this is beautiful

PR #108, opened 2022-04-22
feat: require authorized user for direct upload

martintomas:
Making sure that `/rails/active_storage/direct_uploads` route can be accessed only by authenticated users

## Testing instructions

Can be tested same way as other API endpoints via Rswag documentation

## Tracking

https://vizzuality.atlassian.net/browse/LET-336


tsubik:
Ahh, right, when requiring CSRF token, probably there will be problems with using npm library for activestorage on frontend side.

tsubik:
Let's make sure there will be no problems on the frontend, if there are maybe we can just one security gate like require json and not require CSRF token. Then you can inherit from ActiveStorage Base controller.

tsubik:
Lol, that library is pretty rails specific https://github.com/rails/rails/blob/main/activestorage/app/assets/javascripts/activestorage.js#L520

tsubik:
Maybe then frontend can save csrf-token in meta attribute for activestorage to pick it up.

martintomas:
Tried couple examples with `ActiveStorage::DirectUploadsController` as parent for `DirectUploadsController ` and was not able to make `Devise` work properly, therefore I will keep this reimplementation version of code for now. Definitely feel free to tell me if you have some idea how make this better :)

martintomas:
Oki douky, moved authorization directly to routes which made it work and does not require to patch or reimplement `ActiveStorage::DirectUploadsController`

agnessa (approved):

backend/config/routes.rb
```diff
@@ -1,6 +1,10 @@
 Rails.application.routes.draw do
   get :health_check, to: ->(_env) { [204, {}, [""]] }
 
+  authenticate :user do
```
tsubik, line 4: Nice!

tsubik (approved):

PR #116, opened 2022-04-26
feat: Allow to mark Projects and Project Developers as favourite

martintomas:
I am adding two new models `FavouriteProject ` and `FavouriteProjectDeveloper` which allow to mark appropriate entities as `current_user` favourite ones. Both models come with appropriate API endpoints which allow to create and destroy favouritism.

I am also adding `favourite` attribute to Project and Project Developer serializers so it is possible for FE to decide which entities are user favourites. This unfortunately requires `current_user ` param to be passed into serializer so I have slightly modified other controllers to reflect this change.

## Testing instructions

There are automatic tests which ensure that everything works and both API endpoints can be tested via Rswag doc

## Tracking

https://vizzuality.atlassian.net/browse/LET-323 & https://vizzuality.atlassian.net/browse/LET-319


backend/db/migrate/20220426073907_create_favourite_projects.rb
```diff
@@ -0,0 +1,12 @@
+class CreateFavouriteProjects < ActiveRecord::Migration[7.0]
+  def change
+    create_table :favourite_projects, id: :uuid do |t|
+      t.belongs_to :user, foreign_key: {on_delete: :cascade}, type: :uuid, null: false
+      t.belongs_to :project, foreign_key: {on_delete: :cascade}, type: :uuid, null: false
+
+      t.index [:user_id, :project_id], unique: true
```
agnessa, line 7: do we have anything to prevent hitting pg uniqueness error in case we get request to create a duplicate favourite? what response would FE get in such case?

martintomas: Favourite record is created via `@project.favourite_projects.create! user: current_user` command. What do you think about changing `create!` to `create` so when creation of record fails because of duplicity, it will not crash, but return valid response, which is in this case serialized Project object (`render json: ProjectSerializer.new(@project, params: {current_user: current_user}).serializable_hash`)?

martintomas: Ofc I can also add validation to model, just to be on save side. When duplicity occurs, we either return validation error or fails silently (project is already favourited so we return it, no need to return error)

agnessa: Thanks!

backend/app/controllers/api/v1/favourite_projects_controller.rb
```diff
@@ -0,0 +1,24 @@
+module API
+  module V1
+    class FavouriteProjectsController < BaseController
+      before_action :authenticate_user!
```
tsubik, line 4: We need a story to review authorization in the backend @agnessa. This endpoint and many others, should be only for users that belong to an approved account. I see that `review_status` is for project_developer and investor, not sure why that is not for account. Maybe that will be a chance to implement this with either pundit or cancancan, or if not needed then just simpler.

tsubik (approved):

tsubik:
Just some really minor naming issues :) and questions.

backend/app/serializers/api/v1/project_developer_serializer.rb
```diff
@@ -16,6 +16,12 @@ class ProjectDeveloperSerializer
       attribute :picture do |object|
         image_links_for object.picture
       end
+
+      attribute :favourite do |object, params|
+        next false if params[:current_user].blank?
```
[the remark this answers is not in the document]
martintomas: Yp, good point, will change it ;)

backend/app/controllers/api/v1/favourite_projects_controller.rb
martintomas: I would personally go for cancancan for this kind of restrictions, but we can also pick up pundit or wrap appropriate controller action with some custom method -- basically this should not be difficult after we decide which tech approach to use and which endpoints are restricted.

PR #118, opened 2022-04-26
Add forms/ProjectGallery component

SARodrigues:
This PR adds a small "gallery" to be used on the Project form. It is meant to display the uploaded project photos and allow the user to select one of them to be used as the project cover photo. 

We assume a max of 6 images (which the component doesn't restrict); if not enough images have been added yet then placeholders for them will be displayed, as per designs. 

Behaviour and implementation is similar to (read copied from) the [form/SDGs](https://github.com/Vizzuality/heco-invest/pull/107) component. 

Sizing is adaptive; height the full parent's height (to keep it consistent with the uploader this component will be displayed next to), width is responsive. 

## Testing instructions

Verify that: 
- Image placeholders are added when less than 6 images have been passed  
- An image can be selected as the cover image  
- Error states are shown as expected, and they go away when an image is selected as cover  

## Tracking

[LET-346](https://vizzuality.atlassian.net/browse/LET-346)

## Screenshots

Screenshot 2022-04-26 at 12 02 16
Screenshot 2022-04-26 at 12 03 02

### DIFF
--- a/frontend/containers/forms/project-gallery/project-gallery-image/types.ts
+++ b/frontend/containers/forms/project-gallery/project-gallery-image/types.ts
@@ -1,0 +1,30 @@
+import { UseFormRegisterReturn, RegisterOptions, FieldPath } from 'react-hook-form';
+
+export type ProjectGalleryImageType = {
+  /** Image ID */
+  id: string;
+  /** Title of the image */
+  title: string;
+  /** Source for the image */
+  src: string;
+};
+
+export type ProjectGalleryImageProps<FormValues> = {
+  /** Classes to apply to the container */
+  className?: string;
+  /** Image to display */
+  image: ProjectGalleryImageType;
+  /** Whether the image is the default checked one in the group. Defaults to `false` */
+  defaultSelected?: boolean;
+  /**
+   * Name of the image group (will be assigned to each individual `<Tag />`)
+   * Defaults to inferring it from the first Tag's name.
+   **/
+  name?: string;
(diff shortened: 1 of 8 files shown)


frontend/containers/forms/project-gallery/project-gallery-image/types.ts
```diff
@@ -0,0 +1,30 @@
+import { UseFormRegisterReturn, RegisterOptions, FieldPath } from 'react-hook-form';
+
+export type ProjectGalleryImageType = {
+  /** Image ID */
+  id: string;
+  /** Title of the image */
+  title: string;
```
clementprdhomme, line 7: Since the images will be uploaded by the user, we won't get any titles, right? Should we make it optional?

clementprdhomme (approved):
An improvement you could make is that the screen reader tells the user why an image can be selected. Nothing tells the user it's to make it a cover image.

frontend/containers/forms/project-gallery/project-gallery-image/types.ts
SARodrigues: We should have some auto generated titles IMHO, even if just "Image 1, Image 2" (`image {index + 1}` perhaps), so it's more obvious what the user is selecting. As it is it'll be the responsibility of the wrapper to set it. 

One thing I've left in the open, is that there may be value into having a HOC wrapping both the (future) uploader and this gallery, so that HOC could have the responsibility to decide image titles / tell the SR what's being selected.

PR #121, opened 2022-04-27
Fix/frontend/api deserialization

SARodrigues:
While working on [LET-355](https://vizzuality.atlassian.net/browse/LET-355), it came to my attention that the API service's deserialisation is not working _at all_. 

This PR aims to fix that. 

Major visible changes after deserialisation is fixed:  

- We won't have to dig into `attributes` anymore
  _(it was the biggest hint that deserialisation was not working, but I missed it)_  
- Simplified types 
  _we no longer need to specify `attributes`, `relationships`_  
- `includes` will now work 

**Notes:**  

The one commit that fixes serialization is this one: https://github.com/Vizzuality/heco-invest/pull/121/commits/983628b7ac7ddf9f99c247e25dfaeabd21ab9d28  

The remaining changes are to make everything compatible with serialised data. I may have missed a type or two here and there, but it's a lot of files to touch. 

## Testing instructions

Test that the application works normally

## Tracking

N/A

## Deserialised data example: 
Query: `https://staging.heco.vizzuality.com/backend/api/v1/project_developers/sarodrigues?includes=projects
`
_(If there was pagination in this endpoint, there'd be a `meta` property next to `data` containing the pagination info)_

Screenshot 2022-04-27 at 13 02 33

### DIFF
--- a/frontend/containers/project-form-pages/pages/general-information.tsx
+++ b/frontend/containers/project-form-pages/pages/general-information.tsx
@@ -42,8 +42,6 @@ const GeneralInformation = ({
     setShowInvolvedProjectDevelopers(!!Number(e.target.value));
   };
 
-  console.log(locations);
-
   return (
     <div>
       <h1 className="font-serif text-3xl font-semibold mb-2.5">
--- a/frontend/containers/project-form-pages/pages/general-information.tsx
+++ b/frontend/containers/project-form-pages/pages/general-information.tsx
@@ -42,6 +42,8 @@ const GeneralInformation = ({
     setShowInvolvedProjectDevelopers(!!Number(e.target.value));
   };
 
+  console.log(locations);
+
   return (
     <div>
       <h1 className="font-serif text-3xl font-semibold mb-2.5">
@@ -123,7 +125,7 @@ const GeneralInformation = ({
                 className="mt-2.5"
                 placeholder={formatMessage({ defaultMessage: 'select', id: 'J4SQjQ' })}
               >
-                {locations?.country?.map(({ id, attributes: { name } }) => (
+                {locations?.country?.map(({ id, name }) => (
                   <Option key={id}>{name}</Option>
                 ))}
               </Combobox>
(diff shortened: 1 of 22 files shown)


frontend/services/api.ts
```diff
@@ -21,12 +21,16 @@ const API = axios.create({
 
 const onResponseSuccess = (response) => {
   try {
-    const parsedData = JSON.parse(response);
     return {
-      data: dataFormatter.deserialize(parsedData),
-      meta: parsedData.meta,
+      data: {
+        // Deserialized data
+        data: !!response.data && dataFormatter.deserialize(response.data),
```
clementprdhomme, line 27: `data`'s type would be `boolean | whateverDeserializeReturns`. Shouldn't it be something like this?
```suggestion
        data: response.data ? dataFormatter.deserialize(response.data) : null,
```

frontend/services/api.ts
```diff
@@ -21,12 +21,16 @@ const API = axios.create({
...
+        // Metadata (pagination) if any
+        ...(!!response.data.meta && { meta: response.data.meta }),
```
clementprdhomme, line 29: I think this may as well crash if trying to destructure `false`:
```suggestion
        ...(response.data.meta ? { meta: response.data.meta } : {}),
```

frontend/containers/project-form-pages/pages/general-information.tsx
```diff
@@ -42,6 +42,8 @@ const GeneralInformation = ({
     setShowInvolvedProjectDevelopers(!!Number(e.target.value));
   };
 
+  console.log(locations);
```
clementprdhomme, line 45: 😉

```suggestion
```

clementprdhomme (commented):
Much cleaner!

frontend/containers/project-form-pages/pages/general-information.tsx
SARodrigues: 😅

PR #123, opened 2022-04-27
feat: Save geometry (geojson) of Project

martintomas:
I am adding `geometry` attribute to Project which allows to store GeoJson provided by FE. For now, I have decided to make this attribute `jsonb`, but feel free to share in comments if you would prefered to go for postgis and make this attribute proper `geometry` or `polygon` (http://www.postgis.net/docs/ST_AsGeoJSON.html). 

There are also new Project validations which ensure that `geometry` is valid json and that it can be processed by `RGeo::GeoJSON` lib <-- is valid geojson.

## Testing instructions

Everything is again covered by tests, but it can be also verified manually via Rswag documentation.

## Tracking

https://vizzuality.atlassian.net/browse/LET-322

### DIFF
--- a/backend/app/controllers/api/v1/accounts/projects_controller.rb
+++ b/backend/app/controllers/api/v1/accounts/projects_controller.rb
@@ -47,6 +47,8 @@ module API
               :progress_impact_tracking,
               :description,
               :relevant_links,
+              :geometry,
+              geometry: {},
               involved_project_developer_ids: [],
               target_groups: [],
               impact_areas: [],
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -14,6 +14,7 @@ RUN apk update && \
       tzdata \
       postgresql-dev \
       imagemagick \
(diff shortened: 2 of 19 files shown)


backend/app/controllers/api/v1/accounts/projects_controller.rb
```diff
@@ -47,6 +47,8 @@ def project_params
               :progress_impact_tracking,
               :description,
               :relevant_links,
+              :geometry,
```
martintomas, line 50: I have expected that FE will send geometry as JSON (for example `{type: "Point", coordinates: [1, 2]}`), but Rswag sends similar piece of geometry as `'{type: "Point", coordinates: [1, 2]}'` (String), therefore I am allowing this type of param at controller. Rest of the code is able to deal with this without any extra modification.

agnessa (approved):
thank you, I tested it 👍 sorry it took me so long

PR #127, opened 2022-04-28
Profile header styling change + barebones project page

SARodrigues:
This PR is meant to keep other tasks unblocked while work on [LET-353 (FE - Update the header and contact section of the PD public page)](https://vizzuality.atlassian.net/browse/LET-353) is ongoing. That way work on `LET-353` can be entirely self contained, without causing conflicts with other ongoing tasks. 

**In this PR:**  
- Design changes to the `ProfileHeader` component  
- Added a barebones `project/{id}` page with placeholders for content, as well as the base service to fetch a single project    
  _Virtually the same implementation done in `project-developer/{id}`_
  _It already fetches the `project` from the API_

## Testing instructions

N/A

## Tracking

Part of the ongoing [LET-353 ](https://vizzuality.atlassian.net/browse/LET-353)

## Screenshots 

**Developer profile header**  
Screenshot 2022-04-28 at 11 23 09

**Barebones project page with placeholders** 
Screenshot 2022-04-28 at 11 24 14

### DIFF
--- a/frontend/containers/profile-header/component.tsx
+++ b/frontend/containers/profile-header/component.tsx
@@ -41,7 +41,7 @@ export const ProfileHeader: FC<ProfileHeaderProps> = ({
 
   return (
     <div className={className}>
-      <div className="pt-24 pb-12 bg-center bg-cover sm:pt-40 md:pt-56 bg-radial-green-dark bg-green-dark">
+      <div className="py-10 mx-4 bg-center bg-cover lg:mx-0 lg:px-4 lg:py-18 rounded-2xl bg-radial-green-dark bg-green-dark">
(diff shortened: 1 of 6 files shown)


clementprdhomme (approved):
Just so you keep it in mind, there's a specific viewport size range where the profile image doesn't look as nice as it should be:
The profile image has a white margin at the bottom

SARodrigues:
>  Just so you keep it in mind, there's a specific viewport size range where the profile image doesn't look as nice as it should be

Thank you @clementprdhomme I'm keeping a spreadsheet with "known issues to work on" as per Santi's suggestion (separate one), I'll add it to it. May be a quick fix tomorrow. Thank you for pointing it out.

PR #128, opened 2022-04-28
LET-257 Integrate cloudtasker & create "something" to test Cloud Tasks

agnessa:
This change integrates the cloudtasker gem, which will allow us to use Cloud Tasks as a job processing queue.

The gem comes with a local redis-based server which we can use in development.

I added what I think is going to be needed in terms of environment variables to configure the actual Cloud Tasks integration in staging / production.

Please note: I had trouble with gem dependencies here, I found I needed to:
- upgrade `google-cloud-storage`
- downgrade `faraday`

## Testing instructions

There's a TEMPORARY way to test jobs using endpoints exposed in the swagger doc. There are 2 of them, one works synchronously, the other asynchronously, should be doing the same thing. They will be removed once integration is operational.

Please note this comment for infrastructure setup - I haven't done this:
```
  # Specify the namespace for your Cloud Task queues.
  #
  # The gem assumes that a least a default queue named '${gcp_queue_prefix}-default'
  # exists in Cloud Tasks. You can create this default queue using the
  # gcloud SDK or via the `rake cloudtasker:setup_queue` task if you use Rails.
```

Also please note this placeholder - I don't even know how to do it, the idea is we have a CloudRun jobs instance and this is its inter-platform url, we'll have no https and no auth on it

```
    {
      name  = "CLOUDTASKER_PROCESSOR_HOST"
      value = "TODO: THIS AUTOMAGICKED INTERNAL JOBS INSTANCE URL"
    },
```

## Tracking

[Link to the task(s), if any](https://vizzuality.atlassian.net/browse/LET-257?atlOrigin=eyJpIjoiY2Q3N2EwMDFjM2IzNGEzMDliYjlmZmY1NGEzNWFmNmYiLCJwIjoiaiJ9)


tsubik (approved):
Works locally :+1: tested both sync, async

tiagojsag (commented):
a couple of things: 
- the gcp gem has auto data loading when running inside GCP - see [this](https://github.com/googleapis/google-cloud-ruby/tree/main/google-cloud-storage#authentication). It would be great to know if cloudtasked has the same, so we could remove those redundant env vars altogether
- The current TF changes rely on outputs that do not exist, so this will dead-lock the TF project until we implement the queues.

agnessa:
> 

OK, so it's possible you're right about magical authentication, since cloudtasker builds on google-cloud-tasks gem, which has similar magical project id discovery. I haven't tried without providing these settings to cloudtasker yet.

As for the tf stuff 🤣 , in that case I'll remove it.

agnessa:
I extracted the tf changes here https://github.com/Vizzuality/heco-invest/pull/139 so I can merge this (we can use it locally for now)

martintomas (approved):
Worked for me on local machine ;)

PR #129, opened 2022-04-28
Geometry upload input (no map)

clementprdhomme:
This PR contains part of [LET-348](https://vizzuality.atlassian.net/browse/LET-348).

It adds a “geometry upload input” which:
- accepts the following file types (`.kml`, `.kmz`, `.shp`, `.prj`, `.shx`, `.dbf`, `.cfg`)
- allows the user to picks several at once (as they might be needed to correctly project the geometry)
- converts the files to a single GeoJSON object
- displays error when the files can't be parsed

What's missing from LET-348:
- Displaying a preview of the geometry
- Allowing the user to draw on a map instead of uploading a file
- Allowing the user to search for a specific place on the map
- Allowing the user to make the map bigger

## Testing instructions

Test uploading some geometries. You can be found examples in the code of a previous project in `web/src/testdata/converter`.

## Tracking
Part of [LET-348](https://vizzuality.atlassian.net/browse/LET-348).


SARodrigues (approved):
Tested with `.kml`, `.kmz`, `.shp` and invalid files.

PR #130, opened 2022-04-28
feat: adding filters to index API endpoints

martintomas:
Adding possibility of filtering to index API endpoint for `Project, Project Developer, Investor and Open Call`.  Filtering is done via `API::Filterer` service which is shared across all endpoints. Other features of this service are:
 - it allows to use multiple filters --> you can for example filter by `category` and by `ticket_size` at same time
 - it allows to filter by multiple values per one attribute --> you can for example filter by `category` where `category` is array of values (if you will filter by `category=[forestry-and-agroforestry, non-timber-forest-production]` at `ProjectDeveloper` query, result will contain all records where every record have at least one of these categories)
 - service automatically pluralize searching params, therefore it works for example with `category/categories` attribute at `Project/ProjectDeveloper` model
 - it automatically checks column names of appropriate model and ignores params which does not exist at table --> for example `sdg` filter will be ignored if you use it for `ProjectDeveloper`
 
Reason why I have tried to cover all these use cases is that FE can use same params for filtering and does not need to care which entity it is filtering for. It also means that same service can be used across all controllers without any additional tweaking.

Rswag documentation currently defines filtering params as `string/integer for sdg`, but code supports also variant with `array`. I basically had to pick one option for doc, so decided to go for simpler one mentioned at jira task, but this can be quite easily change to `array`. This is just for needs of doc, code supports both options out of box.

## Testing instructions

There are automatic tests for all modifications, but filtering can be also tested via Rswag.

## Tracking

https://vizzuality.atlassian.net/browse/LET-369


backend/app/services/api/filterer.rb
```diff
@@ -0,0 +1,44 @@
...
+      pluralize(filters.slice(*ENUM_FILTERS)).slice(*column_names).each do |filter_key, filter_value|
+        sql = ActiveRecord::Base.sanitize_sql ["ARRAY[#{filter_key}]::text[] && ARRAY[?]::text[]", Array.wrap(filter_value)]
+        self.query = query.where sql
+      end
+    end
+
+    def filter_by_only_verified
+      return unless filters[:only_verified] && column_names.include?("review_status")
+
+      self.query = query.where review_status: :approved
```
agnessa, line 30: I believe this is not what it needs to do. The only_verified filter applies to projects and open calls and it should work on the "trusted" column (we had a few iterations on the name... 😅 ).

But, as a matter of fact this brings up something else, i.e. in case of project developers and investors we have the `review_status` flag and when showing search results, we should only return those who are `approved`. I imagine this endpoint will be used for public search only, so perhaps we could add a `filter_by_review_status` with condition hardcoded in it, i.e. not controlled by parameters?

martintomas: Aha :), thanks for clarification, will fix it according to your recommendation.

backend/app/controllers/api/v1/investors_controller.rb
```diff
@@ -32,6 +33,12 @@ def fetch_investor
         account = Account.friendly.find(params[:id])
         Investor.find_by!(account_id: account.id)
       end
+
+      def filter_params
+        params.fetch(:filter, {})
```
tsubik, line 38: Other endpoints were not using that kind of form, but I like it :+1: . We should also change in other story the rest of endpoints like /locations to have filters in this way `filter[location_type]`

backend/app/services/api/filterer.rb
tsubik: Yes, approved scope should be applied in controller before filters.

backend/app/services/api/filterer.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    def call
+      filter_by_enums
+      filter_by_only_verified
+      query
+    end
+
+    private
+
+    def filter_by_enums
```
tsubik, line 20: Hmm, clever.

backend/app/controllers/api/v1/investors_controller.rb
```diff
@@ -32,6 +33,12 @@ def fetch_investor
...
+          .permit :category, :impact, :sdg, :instrument_type, :ticket_size, :only_verified,
+            category: [], impact: [], sdg: [], instrument_type: [], ticket_size: []
```
tsubik, line 40: So to filter by multiple request like this is needed `http://localhost:4000/backend/api/v1/investors?filter[category][]=tourism-and-recreation&filter[category][]=forestry-and-agroforestry`, I would probably just try to do something like `filter[category]=cat1,cat2,cat3` but maybe that is ok, what do you think @agnessa ?

tsubik: Maybe that is just a detail, and the current implementation at least does not have to do anything special with params and just feeding the Filterer service.

backend/app/controllers/api/v1/investors_controller.rb
martintomas: Hmm, will try, but I am afraid that FE already uses some of these params 😨

backend/app/controllers/api/v1/investors_controller.rb
martintomas: Actually, that is good point. I normally prefer array option, even though url looks more ugly, but we already have some params which use `,` to split multiple options and it will solve rswag doc issue, so why not ;)

backend/app/controllers/api/v1/investors_controller.rb
agnessa: something for the future then :)

backend/app/services/api/filterer.rb
```diff
@@ -0,0 +1,44 @@
...
+      pluralize(filters.slice(*ENUM_FILTERS)).slice(*column_names).each do |filter_key, filter_value|
+        sql = ActiveRecord::Base.sanitize_sql ["ARRAY[#{filter_key}]::text[] && ARRAY[?]::text[]", filter_value.split(",")]
```
agnessa, line 22: love how you avoided `IN`

backend/app/controllers/api/v1/investors_controller.rb
martintomas: :) in the end, I have changed this with this PR, will see if it blows up something :D

agnessa: oh, but you changed that already? I'd probably separate that in order not to break FE, or maybe check with them how hard it is to change this?

PR #139, opened 2022-05-02
LET-244 Ad-hoc async tasks using Cloud Tasks

agnessa:

- You ran the tests
    
---

Description of the PR

## Testing instructions

What to test? How to do it?

## Tracking

Link to the task(s), if any


backend/.env.sample
```diff
@@ -6,6 +6,10 @@ IS_JOBS_INSTANCE=true
 BACKEND_URL=localhost:4000/backend
 RAILS_RELATIVE_URL_ROOT=/backend
 
+GCP_REGION=
+GCP_PROJECT_ID=
```
tsubik, line 10: Hmm, I see we need to set that anyway for cloudtasker?

agnessa: Tiago thinks we don't, that was my first attempt 😊

tsubik: that depends probably on this gem, should it also use some default region etc.

tsubik: and looks like project_id needs to be set in  cloudtasker initializer. But it could probably be taken from `Google::Auth.get_application_default`

agnessa:
Tiago implemented it in a separate PR, closing

PR #143, opened 2022-05-02
LET-327 Asynchronous translation of the project

agnessa:
Integrated Google Translation API to automatically translate columns when project is created. It is handled by startig a background job in the controller, not via model callback.

I haven't  been able to figure out how to parametrise this API without having an env var with GCP project id available.

When running terraform plan I see that the new env var will be added and the new service "created". I honestly don't know if that is the equivalent of clicking the "enable" button.

## Testing instructions

Locally:
- set the project id env var in .env
- start cloudtasker
- create a project using rswag docs
- check in server log that translation succeeded (no idea how to change lng using rswag docs)

## Tracking

https://vizzuality.atlassian.net/browse/LET-327?atlOrigin=eyJpIjoiMTU4YzI1NjFkNGFjNGUyOGFiM2I2YjljYTY4NzljNGUiLCJwIjoiaiJ9


infrastructure/base/modules/env/main.tf
```diff
@@ -147,6 +147,10 @@ module "backend_cloudrun" {
       name  = "DATABASE_HOST"
       value = module.database.database_host
     },
+    {
```
tiagojsag, line 150: this should not be needed: https://cloud.google.com/ruby/docs/reference/google-cloud-translate/latest/AUTHENTICATION

infrastructure/base/modules/translation/main.tf
```diff
@@ -0,0 +1,4 @@
+resource "google_project_service" "translation_api" {
```
tiagojsag, line 1: :+1:

tiagojsag (changes requested):

backend/app/controllers/api/v1/accounts/projects_controller.rb
```diff
@@ -12,6 +12,8 @@ def create
               language: current_user.account.language
             )
           )
+          Translations::TranslateProjectJob.perform_later(project.id)
```
martintomas, line 15: I believe it makes probably more sense to add this as after action on model -- at least if we plan to run translation job everytime when translatable column changes. If this approach would be too aggressive, it could be moved to some controller concern and again as after action. What I am trying to avoid is situation where you have to explicitly define translation across all endpoints <-- much easier and maintainable is to have it at one place for all translatable models

backend/app/jobs/translations/translate_project_job.rb
```diff
@@ -0,0 +1,16 @@
+module Translations
+  class TranslateProjectJob < ApplicationJob
+    queue_as :default
+
+    rescue_from(ActiveRecord::RecordNotFound) do |exception|
```
martintomas, line 5: maybe should be moved to some base class for Translations or even to ApplicationJob

backend/app/jobs/translations/translate_project_job.rb
```diff
@@ -0,0 +1,16 @@
...
+      Rails.logger.error(exception)
+      Rails.logger.error(Rails.backtrace_cleaner.clean(exception.backtrace).join("\n"))
+      # TODO: handle exception tracking
+    end
+
+    def perform(project_id)
```
martintomas, line 11: I am just wondering, AR models should come with its own serialisers, therefore you can ` Translations::TranslateProjectJob.perform_later(project)` and here can be `perform(project)` <-- no need to pass `id` and then have `project = Project.find(project_id)` everywhere

backend/app/jobs/translations/translate_project_job.rb
```diff
@@ -0,0 +1,16 @@
...
+      project = Project.find(project_id)
+      Translations::TranslateProject.new(project).call
```
martintomas, line 13: no big deal, but I believe this will be always called via this job, so maybe logic could be moved here and no need to have special service for it

backend/app/services/translations/translate_content.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    def client
+      @client = Google::Cloud::Translate.translation_service
+    end
+
+    def call(target_language_code)
+      response = client.translate_text contents: @translatable_content,
+        source_language_code: @source_language_code,
+        target_language_code: target_language_code,
+        parent: "projects/#{ENV["GCP_PROJECT_ID"]}/locations/global",
```
martintomas, line 16: looks like this depends on `projects` entity, even though it looks like same base class for content translations.

backend/app/services/translations/translate_project.rb
```diff
@@ -0,0 +1,50 @@
...
+        end
+      end
+      @project.save!
+    end
+
+    def source_language_code
+      @project.language
+    end
+
+    def self.translatable_columns
```
martintomas: Hmm, I am not really convinced that we need special class for Project translation (TranslateProject). It would be great to have one class which translates every resource `Translations::Translate.new(resource).call`. Information about `translatable_columns` can be obtained from via Traco methods (I think method is called `translatable_attributes`). Rest of the code looks like it can be made more abstract

backend/spec/rails_helper.rb
```diff
@@ -43,4 +43,13 @@
       ActiveJob::Base.queue_adapter.performed_jobs.clear
     end
   end
+
+  config.before do
+    # this is so we don't hit the Translation API in tests, don't know of a civilised way to disable it
+    @translate_content_double = instance_double(Translations::TranslateContent)
```
martintomas, line 49: I think more elegant solution is either use webmocks (and then stub request) or use `vcr` which makes it even simpler. Also this stubbing should be turned on only at tests which it apply for --> in this case, it will be done for all tests, which I think is kind of overkill

backend/spec/services/translations/translate_project_spec.rb
```diff
@@ -0,0 +1,23 @@
+require "rails_helper"
+
+RSpec.describe Translations::TranslateProject do
+  let(:project) {
+    translatable_columns = Translations::TranslateProject.translatable_columns
+    attrs = translatable_columns.map { |col| ["#{col}_en", "test"] }.to_h
+    create(:project, attrs.merge(name_en: "Project name")) # for simplicity just one translatable attribute present
```
martintomas, line 6: Yup, I think it is fine :), I don't think we need to run translation for every translatable column and every language

martintomas (commented):
Looks good, but I would really give it though if it is:
 - possible to move translations to some action hook so you don't need to specify it manually across whole app
 - possible to make `TranslateProjectJob` more abstract so it can be used for any translatable resource

backend/app/controllers/api/v1/accounts/projects_controller.rb
agnessa: OK, that's what I did first, but then found specs irritating to update and ended up with this :)

backend/app/jobs/translations/translate_project_job.rb
agnessa: yes, I would go for a superclass for translation jobs

backend/app/jobs/translations/translate_project_job.rb
agnessa: I prefer skinny jobs, as it is then easier to use the same service in the console or rake task or somewhere else entirely, and also - writing specs for a service is easier than writing specs for a job I think :)

backend/spec/rails_helper.rb
agnessa: OK, good point

backend/spec/services/translations/translate_spec.rb
```diff
@@ -0,0 +1,23 @@
+require "rails_helper"
+
+RSpec.describe Translations::Translate do
+  subject { described_class.new resource }
+
+  describe "#call" do
+    [Project, ProjectDeveloper, OpenCall, Investor, Account].each do |klass|
```
agnessa, line 7: :)

PR #144, opened 2022-05-03
feat: locations selectors

barbara-chaves:
This PR adds interdependent filters to the locations filters (Project form)

## Testing instructions

* The filters options should be filtered by their parent input:
   * country -> state -> municipality

* The selectors should work without filters

* The dependent fields should be reseted when their parent changes

## Tracking

[LET-347](https://vizzuality.atlassian.net/browse/LET-347)

### DIFF
--- a/frontend/containers/project-form-pages/pages/general-information.tsx
+++ b/frontend/containers/project-form-pages/pages/general-information.tsx
@@ -30,7 +30,7 @@ const GeneralInformation = ({
 }: ProjectFormPagesProps<ProjectForm>) => {
   const [showInvolvedProjectDevelopers, setShowInvolvedProjectDevelopers] = useState(false);
   const [locationsFilter, setLocationsFilter] =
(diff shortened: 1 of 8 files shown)


frontend/containers/project-form-pages/pages/general-information.tsx
```diff
@@ -25,11 +26,14 @@ const GeneralInformation = ({
   errors,
   control,
   getValues,
+  resetField,
 }: ProjectFormPagesProps<ProjectForm>) => {
   const [showInvolvedProjectDevelopers, setShowInvolvedProjectDevelopers] = useState(false);
+  const [locationsFilter, setLocationsFilter] =
+    useState<{ country?: string; department?: string }>();
```
clementprdhomme, line 33: You could avoid possible mistakes by doing this:
```suggestion
    useState<{ country: string; department: string }>({ country: undefined, department: undefined });
```

clementprdhomme (approved):

PR #145, opened 2022-05-03
feat: Restricting visibility of contact attributes to approved accounts

martintomas:
Making sure that contact information attributes for PD endpoints are restricted and only user with approved account can see them.

## Testing instructions

You can trust tests or rswag ui :)

## Tracking

https://vizzuality.atlassian.net/browse/LET-318


backend/app/models/account.rb
```diff
@@ -29,4 +29,8 @@ def slug_preview
     set_slug unless slug.present?
     slug
   end
+
+  def approved?
+    investor&.approved? || project_developer&.approved?
```
tsubik, line 34: I'm not sure why we have approved on investor/project_developer and not directly on account, maybe that is just a mistake? @agnessa

tsubik (commented):
Are there any API specs for logged-in user? I see all snapshots changing those attributes to `null` so those are for public users.

martintomas:
> Are there any API specs for logged-in user? I see all snapshots changing those attributes to `null` so those are for public users.

You should be able to see filled values for example at https://github.com/Vizzuality/heco-invest/blob/feature/restrict-visibility-contacts/backend/spec/fixtures/snapshots/api/v1/accounts-project-developer.json#L27 where response is returned for user with approved account.

backend/app/models/account.rb
martintomas: Yeah, hmm, I agree with you, if `Account` model is used to save data which are same for Investor and Project Developer, then `approved` attributes should be moved to it. If you all of us agree, I can refactor code this way :)

agnessa: agree

agnessa (approved):

backend/app/models/account.rb
tsubik: Yes, better to move it. Seems more logical to have account.approved directly there.

tsubik:
@martintomas but this is a different controller, the private/account one. I'm talking about some one extra spec for public controller to for logged-in user. To make sure contact info is returned.

martintomas:
> @martintomas but this is a different controller, the private/account one. I'm talking about some one extra spec for public controller for logged-in user. To make sure contact info is returned.

Aha, okay, I guess there is no one. I though that private/account one is enough, because serializer is still same, but I can add new test to public endpoint too, np

PR #148, opened 2022-05-04
feat: Adding sorting to index API endpoints (Project, PD, Investor, OpenCalls)

martintomas:
I am adding `sorting` argument to index API endpoints for Project, Project Developer, Investor and Open Calls. There is new service called `API::Sorter` responsible for modifying existing query with correct `order` option.

Tricky part was sorting by `name` which is saved at `Account` table for Project Developer and Investor and it is not translated (`name` attribute of Project or Open Call is translated). In the end I decided to define simple structure stored at `SORTING_OPTIONS ` which tells me if localization should be used or not and if sorting attribute is taken from different table.

## Testing instructions

New param `sorting` is available via Rswag or can be tested manually.

Available values of sorting param are: `sorting=name%20desc`, `sorting=name%20asc`, `sorting=created_at%20desc`, `sorting=created_at%20asc`

## Tracking

https://vizzuality.atlassian.net/browse/LET-326

### DIFF
--- a/backend/app/services/api/sorter.rb
+++ b/backend/app/services/api/sorter.rb
@@ -1,0 +1,61 @@
+module API
+  class Sorter
+    attr_accessor :query, :sorting_query_param, :sorting_direction_param, :language
+
+    SORTING_DIRECTIONS = %i[asc desc]
+    SORTING_OPTIONS = {
+      name: {
+        localized: true,
+        specific: {ProjectDeveloper => {relation: :account, localized: false},
+                   Investor => {relation: :account, localized: false}}
+      },
+      created_at: {}
+    }
+
+    def initialize(query, sorting_by:, language: I18n.locale)
+      @query = query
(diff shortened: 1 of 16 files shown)


backend/app/services/api/sorter.rb
```diff
@@ -0,0 +1,61 @@
+module API
+  class Sorter
+    attr_accessor :query, :sorting_query_param, :sorting_direction_param, :language
+
+    SORTING_DIRECTIONS = %i[asc desc]
+    SORTING_OPTIONS = {
+      name: {
+        localized: true,
+        specific: {ProjectDeveloper => {relation: :account, localized: false},
```
tsubik, line 9: We can check which columns have translations and configure filters accordingly, for example `Investor.locale_columns`

tsubik: I would prefer to do it automatically to not forget about it.

agnessa (approved):

PR #149, opened 2022-05-04
Add breadcrumbs component to Project/Project Developer/Investor pages

SARodrigues:
This PR adds a "simple" breadcrumbs component, to be use in the PD, Project, Investor, Open Call public pages. 

It's very project specific but although the breadcrumbs are generated almost automatically (just needs the name of the PD/Project/etc passed via prop), it still retains enough flexibility to be used in other ways, if required. Examples can be found in Storybook. 

They're also accessible, following the description/examples found [here](https://www.aditus.io/patterns/breadcrumbs/).

**In this PR:**  
- Added a `containers/Breadcrumbs` component  
- Set up the new component in the Project Developer, Project, Investor pages. 

**Note:**  
The breadcrumbs suggest that, for instance when viewing a Project page, that there is a link to an index of Project pages. I hardcoded links to a `/search/projects` page and so on, just for example/testing purposes. We will need to adjust these links as the respective search pages are created. 

Fallbacks for when the substitution/name for a dynamic url param hasn't been passed have been implemented to avoid development errors; the component can infer the breadcrumb name based on the route path segments. 

## Testing instructions

Open _a_ Project Developer, Project, Investor page and verify that the breadcrumbs show correctly. 

## Tracking

[LET-341](https://vizzuality.atlassian.net/browse/LET-341)

## Screenshot
Screenshot 2022-05-04 at 16 56 11

### DIFF
--- a/frontend/containers/breadcrumbs/component.tsx
+++ b/frontend/containers/breadcrumbs/component.tsx
@@ -2,6 +2,8 @@ import { FC } from 'react';
 
 import { useIntl } from 'react-intl';
 
+import cx from 'classnames';
+
 import Link from 'next/link';
 import { useRouter } from 'next/router';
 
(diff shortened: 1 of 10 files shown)


barbara-chaves (approved):

frontend/containers/breadcrumbs/component.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+export const Breadcrumbs: FC<BreadcrumbsProps> = ({
+  className,
+  substitutions: propSubstitutions = {},
+}: BreadcrumbsProps) => {
+  const intl = useIntl();
+  const { route, asPath } = useRouter();
+
+  // Predefined substitutions for this project. Ideally this would live in a constants file,
+  // but due to the need to use `react-intl` they're kept here.
```
clementprdhomme, line 22: I found the same issue on another part of the code. I still moved the contants to a different files (`helpers.ts`) and passed `intl` as a prop. But it's ok to keep it here.

frontend/containers/breadcrumbs/component.tsx
```diff
@@ -0,0 +1,128 @@
...
+  };
+
+  const routeSegments: string[] = route.split('/').slice(1);
+  const pathSegments: string[] = asPath.split('/').slice(1);
+
+  const nameFromPathname = (text: string): string =>
+    text
+      .replace(/(-|_)/, ' ') // Change Dashes and Underscores into spaces
+      .split(' ') // Split into an array of words
+      .map((word) => word.charAt(0).toUpperCase() + word.slice(1)) // Uppercase words' first letters
```
clementprdhomme, line 49: Be careful with the `chartAt` function. I believe it won't work as you expect for certain characters (such as emojis) which are actually composed of several characters. In this case however, I don't think any changes are needed.

clementprdhomme (approved):
Good job!

PR #150, opened 2022-05-05
refactoring: Removing multipart support from API endpoints

martintomas:
FE uses direct upload now, therefore I am removing multipart support from API endpoints ;)

## Testing instructions

Everything can be tested via Rswag, but I have also tweak our tests to use only json content type and they works as well :)

## Tracking

https://vizzuality.atlassian.net/browse/LET-321


tsubik (approved):
Tested locally with local frontend, works fine for project developer account creation. Project create form is not done yet on the frontend.

PR #156, opened 2022-05-06
feat: New Account-Investor API endpoints

martintomas:
New API endpoints for Account/Investor creation/updation/show. All logic is basically copy&paste of Account/Project Developer API endpoints which already support same functionality.

This PR adds:
```
GET /api/v1/account/investor
POST /api/v1/account/investor
PUT /api/v1/account/investor
```

## Testing instructions

All new API endpoints are covered with tests and everything can be also tested via Rswag UI.

## Tracking

https://vizzuality.atlassian.net/browse/LET-387


tsubik (approved):
:+1:

PR #163, opened 2022-05-10
fix: add max size validation to image uploader

barbara-chaves:
This PR adds an file size validation to the image-uploader component, on the PD form

## Testing instructions

The user can't upload a file larger than 5MB
An error is displayed in this case

## Tracking

It is not a jira task, but is related to https://vizzuality.atlassian.net/browse/LET-352

### DIFF
--- a/frontend/components/forms/image-uploader/component.tsx
+++ b/frontend/components/forms/image-uploader/component.tsx
@@ -36,7 +36,6 @@ export const ImageUploader = <FormValues extends FieldValues>({
       id: 'F++AYx',
     })
   ) => {
-    console.log('error');
     setImagePreview(null);
     setValue(name, undefined);
     setError(name, {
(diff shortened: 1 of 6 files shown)


frontend/components/forms/image-uploader/component.tsx
```diff
@@ -21,34 +23,57 @@ export const ImageUploader = <FormValues extends FieldValues>({
...
   const { formatMessage } = useIntl();
   const [imagePreview, setImagePreview] = useState<string>();
 
+  const setInputError = (
+    message: string = formatMessage({
+      defaultMessage: 'Something went wrong with the picture upload',
+      id: 'F++AYx',
+    })
+  ) => {
+    console.log('error');
```
clementprdhomme, line 39:
```suggestion
```

PR #164, opened 2022-05-10
Let 354 favorite project developer

barbara-chaves:
This PR adds the connection to the PD page favourite button to the endpoints that control the favourite status

## Testing instructions

The user can add and remove a PD to favourites.
Screenshot 2022-05-10 at 12 47 46
When the PD is a favorite, the heart icon turns filled (dark green)

Screenshot 2022-05-10 at 12 47 26

## Tracking

[LET-354](https://vizzuality.atlassian.net/browse/LET-354)


clementprdhomme (changes requested):
In addition to my comments, I noticed that if I favourite a PD and reload the page, the PD is not marked as favourited anymore. Could you have a look?

barbara-chaves:
> 

I added an useQuery on the page to fix the data inconsistency problem on reload. That way we are sure we are always using the updated data.

frontend/pages/project-developer/[id]/index.tsx
```diff
@@ -42,29 +47,42 @@ export const getServerSideProps = async ({ params: { id }, locale }) => {
...
+    {
+      includes: 'projects',
+    },
+    initialProjectDeveloper
+  );
+
+  const [isFavourite, setIsFavourite] = useState(projectDeveloper.favourite);
+
+  useEffect(() => {
+    // this useEffect is needed because the initial PD can be different from the current
```
Comment (clementprdhomme) [inline frontend/pages/project-developer/[id]/index.tsx:75]: I was really confused about why this is still needed but I've finally understood. On the server, when we fetch the PD, we don't send the session cookie so the endpoint doesn't tell us if the PD is in the favourites. When the hook executes on the client (the browser), we do send the token and thus `projectDeveloper.favourite` has a different value.

Since you're storing the “favourite status” in the `useState` hook, you need to update it whenever the `useQuery` hook returns new data.

Another way you could have implemented this is if you wouldn't use `useState` at all and would invalidate the `useQuery` request when the mutation has succeeded.

I think this comment could be improved though ;).

Comment (barbara-chaves) [reply frontend/pages/project-developer/[id]/index.tsx]: You are right, I will ad this explanation to the comment

PR #166, opened 2022-05-11
feat: Compute centroid from Project geometry (GeoJson)

martintomas:
I am moving all GeoJson validation from Project model to special validator called `ProjectGeometryValidator `. This validator:
 - verifies that geojson is `json` and `geojson`
 - makes sure that `geometry` inside geojson is not-empty
 - makes sure that `geometry` is of supported type

Centroid of Geojson geometry is computed via `GeoJsons::ToCentroid` which currently supports `Point, Polygon, Feature and FeatureCollection`. It raises appropriate error in case that geometry is non-existent or not supported.

I am also adding `latitude` and `longitude` to Project serializer. I think that at future it would be better to store coordinates directly at db with help of postgis, but I am keeping these attributes for now.

By the way, had to do one small tweak of workflows, because `rgeos` gem must be installed only after `libgeos-dev` lib is available.

## Testing instructions

I have tried to cover all use cases with tests, but If you wanna test all stuff manually, you can follow examples mentioned at `backend/spec/models/project_spec.rb`.

## Tracking

https://vizzuality.atlassian.net/browse/LET-324


agnessa (approved):
tested, works, thank you!

agnessa:
I found that when testing using swagger docs you need to put the geometry on one line and escape the quotes inside, here's an example polygon I used
`"geometry": "{\"type\": \"Feature\",\"properties\": {},\"geometry\": {\"type\": \"Polygon\",\"coordinates\": [[[-74.10415649414062,3.3050503438610943],[-74.07394409179688,3.2666614116116146],[-74.05059814453125,3.3050503438610943],[-74.091796875,3.3406958799501414],[-74.10415649414062,3.3050503438610943]]]}}",
  `

PR #168, opened 2022-05-12
fix: Removing /backend suffix from BACKEND_URL

martintomas:
Related to issue mentioned at https://vizzuality.atlassian.net/browse/LET-365 by @barbara-chaves . Final absolute url is composed from `"#{BACKEND_URL}/#{RAILS_RELATIVE_URL_ROOT}"`, therefore `/backend` suffix at both of them created redundancy at ActiveStorage urls.

## Testing instructions

Open console and generate absolute url for any blob --> it should contain `/backend` part only once now.

## Tracking

https://vizzuality.atlassian.net/browse/LET-365


martintomas:
Hmm, maybe this is not fix actually 🤔 , I have played with `BACKEND_URL` variable a bit and it looks like for needs of `url_for`, only host part of `BACKEND_URL` is used.
```
(base) Martins-MacBook-Air:backend martintomas$ BACKEND_URL=localhost:4000/backend RAILS_RELATIVE_URL_ROOT=/backend rails c
Loading development environment (Rails 7.0.2.4)
irb: warn: can't alias context from irb_context.
3.1.0 :001 > blob = ActiveStorage::Blob.first
  ActiveStorage::Blob Load (19.6ms)  SELECT "active_storage_blobs".* FROM "active_storage_blobs" ORDER BY "active_storage_blobs"."id" ASC LIMIT $1  [["LIMIT", 1]]
 =>                                                        
#<ActiveStorage::Blob:0x000000010b65fe30                   
...                                                        
3.1.0 :002 > Rails.application.routes.url_helpers.url_for blob
 => "http://localhost/backend/rails/active_storage/blobs/redirect/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaEpJaWt3T0RFNE9UUmlZUzA0TVRrMkxUUmxZalV0T1RNMU9DMW1PREkwWkdOaU0yVTBNREFHT2daRlZBPT0iLCJleHAiOm51bGwsInB1ciI6ImJsb2JfaWQifX0=--1d3c333a42aadd0c49e2a84df529d0ba3554cf72/picture.jpg"
```
in such case, I have no idea where `/backend` duplicity comes from...

tsubik:
I think the configuration in application.rb is wrong (using BACKEND_URL), it should be host only, port is different option, and path probably not needed. Also, please remove `dump.rdb` and add it to gitignore.

tsubik:
Not sure where the duplicaiton comes from, but adding that path `backend` drops the port in url, which is weird. After removing path at least in development in API docs links look good.
image

tsubik:
Ok, @martintomas I know where the duplication comes from, your PR is correct to remove `backend` path. Try to use BACKEND_URL as `localhost/backend` without port as with staging and production domain there is no port in URL. You will see 
image

tsubik (changes requested):
Please first remove extraneous file before merging.

backend/config/application.rb
```diff
@@ -11,7 +11,7 @@ class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 7.0
 
-    Rails.application.routes.default_url_options[:host] = ENV.fetch("BACKEND_URL", "http://localhost:4000")
+    Rails.application.routes.default_url_options[:host] = URI.parse(ENV.fetch("BACKEND_URL", "http://localhost:4000")).host
```
tsubik, line 14: I don't know why you changed that. It was actually ok, to have host with port, now it is incorrect in api-docs (without port). I would drop "/backend" path in BACKEND_URL.

martintomas: Hmm, forgot about port at swagger examples, but BACKEND_URL kind of implies that it is full path to place where Rails is run so it made more sense for me to keep /backend there and fix host. Whatever, I dont mind to change it back, but maybe it would be worthy to rename BACKEND_URL to BACKEND_HOST

martintomas: Or it can be defined same way as MAILER_DEFAULT_HOST:
config.action_mailer.default_url_options = {host: ENV["MAILER_DEFAULT_HOST"], protocol: "https"}
at env files. This way, we can have port and protocol, but still keep /backend path

tsubik (approved):

PR #170, opened 2022-05-13
feat: Store progress of background jobs at Events table

martintomas:
This PR add new model called `BackgroundJobEvent` which is used to store progress of background jobs. Every job is monitored via `action` hooks and actual status is saved into db. In case of exception, even backtrace is stored. 

There is also new endpoints `GET /api/v1/background_job_events` which is available only for authenticated users (for Admins at future) and which returns data from BackgroundJobEvent table. These data can be filtered (`:status, :created_at_min, :created_at_max`) and are paginated.

## Testing instructions

Start new job and send request to `GET /api/v1/background_job_events`. You should see corresponding information -- if even is enqueued, completed or it crashed, etc.

## Tracking

https://vizzuality.atlassian.net/browse/LET-392

### DIFF
--- a/backend/app/controllers/api/v1/background_job_events_controller.rb
+++ b/backend/app/controllers/api/v1/background_job_events_controller.rb
@@ -1,0 +1,32 @@
+module API
+  module V1
+    class BackgroundJobEventsController < BaseController
+      include API::Pagination
+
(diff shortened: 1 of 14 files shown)


agnessa (approved):
looks good, thanks!

PR #172, opened 2022-05-13
feat: Projects Map API endpoint

martintomas:
New API endpoint for Projects Map -- `GET /api/v1/projects/map`. It reuses a lot of functionality from ProjectsController, but is more optimized for needs of Maps.

## Testing instructions

New endpoint can be tested via Rswag UI.

## Tracking

https://vizzuality.atlassian.net/browse/LET-116


backend/app/controllers/api/v1/projects/maps_controller.rb
```diff
@@ -0,0 +1,19 @@
+module API
+  module V1
+    module Projects
+      class MapsController < BaseController
+        def show
+          projects = Project.select(:id, :geometry, :latitude, :longitude)
```
agnessa, line 6: I think we don't need geometry here - just lat / long. Leaving out geometry will make this response significantly smaller I think. But I would add the "trusted" flag here, in case we need to differentiate display of markers.

tsubik (approved):

PR #159, opened 2022-05-10
Search page layout & API connection

SARodrigues:
This PR adds a base working `DiscoverPageLayout` to be used by all four of the search pages _(Projects, Project Developers, Investors, Open Calls)_, as well as a discover projects page (with functional search and pagination), as well as placeholder pages. 

-----

**In this PR:**  
- Layout to be used in the Discover/Search pages  
  - is connected to the API, is currently set up to handle searching for projects
    _with a couple examples on how to add the remaining ones._   
  - can handle search (for searching) and page params (for pagination)   
  - is reactive to query param changes, as well as current page in order to fetch the correct data  
  - is responsive / compatible with mobile
- Projects page, which:  
  - is connected to the API  
  - pagination works  
  - searching works  
  - Will scroll projects list to top when the user changes a page  

-----

The layout makes use of the previously built components:  
- `DiscoverSearch` which takes care of the user search input  
- `BadgeNavigation` which takes care of of the navigation  

This layout is responsible for:  
- Based on the route and existing query params (currently `search` and `page`), fetch the data;  
- Pass the correct dataset to the Child element, which in this case will be the active page  
- Pass the query state to the child page (whether is loading/fetching data, so that a spinner can be shown on the page)  

It also supports:  
- `screenHeightLg` prop, which will make it so that, when the prop is set to `true`, the layout will be of a fixed height matching the screen height. This prop is currently used in the Discover/Projects page.  

**Notes:** 
- User and Menu header to be added separately in [LET-439](https://vizzuality.atlassian.net/browse/LET-439)  
- Project details card to be added separately in [LET-399](https://vizzuality.atlassian.net/browse/LET-399)

## Testing instructions

Verify that:  
- All four pages load correctly (`/discover/` + `projects`|`project-developers`|`investors`|`open-calls`)   
- Visiting `/discover` redirects to `/discover/projects`  
- The project page loads correctly  
- A list of projects is shown  
- User can search projects  
- Projects pagination works  
- Clicking on a project card will take the user to the respective project page
  _project card has not yet been implemented_

## Tracking

[LET-395](https://vizzuality.atlassian.net/browse/LET-395)

## Video

https://user-images.githubusercontent.com/6273795/168284700-8da7ea4e-f173-4524-98fa-94b9a30deea9.mp4

PR #173, opened 2022-05-13
Add UserMenu and NavigationMenuButton to the Discover/Search layout

SARodrigues:
## NOTE

This PR is built on top of #159. 

## Description

This PR Adds the UserMenu and Menu button to the header of the Discover layout. 

In addition, I've split the menus into their own separate components, so that they can be reused in different layouts' headers. 

&nbsp; 

**Note:**  
I ran into `z-index` issues. Due to how the layout was built, the `DiscoverSearch` overlapping the header bar needed to have a higher `z-index` than the header bar. However, due to how the menu `Popover` is built, this would mean they would be displayed _below_ the `DiscoverSearch` component. 

I think a proper solution would to be rebuilding the menu popovers to be broken free from its parent component; I gave it a  quick try with [portals](https://reactjs.org/docs/portals.html) & [popper](https://popper.js.org/react-popper/v2/), however I didn't manage to pull it off in a timely manner. 

As a quick (but perhaps not ideal) workaround, `DiscoverSearch` was also added to the `Header`. Maybe we can revisit this in the future. 

## Testing instructions

Verify that:  
- The user menu displays correctly (and above the `DiscoverSearch` component)  
- The menu button displays correctly  
- The user menu is only visible for `md` and larger screens (the menu button turns into a hamburguer button on small screens and takes over)  
- When the user is not signed in, a Sign In button is shown instead of the user menu  

## Tracking

[LET-439](https://vizzuality.atlassian.net/browse/LET-439)

## Screenshots 
Screenshot 2022-05-16 at 15 48 23

Screenshot 2022-05-16 at 15 48 33


frontend/containers/layouts/navigation-menu-button/component.tsx
```diff
@@ -0,0 +1,124 @@
...
+                <MenuIcon height={20} />
+              </span>
+            </Button>
+          }
+          align="end"
+          onAction={onClickMenuItem}
+          hiddenSections={{ 'user-section': 'sm' }}
+        >
+          <MenuSection>
+            <MenuItem key="/discover">
```
clementprdhomme, line 68: Maybe you could use the `Paths` enum values here? You might need to cast them.

clementprdhomme (approved):

frontend/containers/layouts/navigation-menu-button/component.tsx
SARodrigues: Thank you. Changed the hardcoded links to `Paths` in the navigation menus/bars touched.

PR #175, opened 2022-05-16
feat: Services for importing data from geojsons

martintomas:
This PR adds four new simple importing classes used to seed all geojson data mentioned at https://vizzuality.atlassian.net/browse/LET-383 (colombia, municipality, mosaic and basin). Seeding can be triggered manually, class by class or by new rake task: `rake import_geojsons:colombia` which also deletes obsolete location records not mentioned at geojsons. Main purpose of geojson data is to provide:
 - geometry
 - biodiversity
 - biodiversity_demand
 - climate
 - climate_demand
 - community
 - community_demand
 - water
 - water_demand
 
which are saved as part of `locations` table.

I have not added original geojson files as part of this PR, because I think it is quite a lot of data. If you think it should be added, please let me know ;). For now, for purposes of tests, you will need to download data manually from Jira. For purposes of seeding Staging, we will need to manually transfer these files to server (not sure if it is possible 🤔 ).

Please take into account, that this PR requires installed `postigs`. We will also need to install `postgis` at Stage.

## Testing instructions

Download geojson data from Jira and save them at `db/seeds/files` folder. Run `rake import_geojsons:colombia`.

## Tracking

https://vizzuality.atlassian.net/browse/LET-383


.github/workflows/backend-tests.yml
```diff
@@ -20,7 +20,7 @@ jobs:
 
     services:
       postgres:
-        image: postgres:14-alpine
+        image: postgis/postgis:14-3.2-alpine
```
tsubik, line 23: Please also update both docker-composes file we have.

agnessa:
So this one we cannot merge before we set up infra for postgis; also, we need to think about how we can run the imports, because transferring to the instance is not an option I think. I believe it is possible to run the rake task locally, connecting to the staging db - will that do?

tsubik (changes requested):
If we don't want to include the geojson files in repo, we should have them in .gitignore. But I'm not sure if not just include them, and add this to seeds? Just for everyone new to have local db with everything ready, not sure, or at least add this rake task to seeds but with warning if you don't have those files locally? 

Also, about having big geometries in location model, locally I see API endpoints don't have a big perf issue, although it's 4sec vs 0.4, but when trying `Location.all` in rails console it hangs ;], and if we start joining locations with other models that could be messy.

martintomas:
> If we don't want to include the geojson files in repo, we should have them in .gitignore. But I'm not sure if not just include them, and add this to seeds? Just for everyone new to have local db with everything ready, not sure, or at least add this rake task to seeds but with warning if you don't have those files locally?
> 
> Also, about having big geometries in location model, locally I see API endpoints don't have a big perf issue, although it's 4sec vs 0.4, but when trying `Location.all` in rails console it hangs ;], and if we start joining locations with other models that could be messy.

I have tried to address all your issues @tsubik ;). Geometries for locations are now saved at separate model called `LocationGeometry` --> It was a bit tricky to run `upsert_all` for `Location` and `LocationGeometry` models during import, but should work now 🤞 . I have added GeoJson files to .gitignore and added rake task responsible for running import of Colombia locations to seeding <-- it warns user in case geojson files are missing during seeding (uses location factory in such case).

I have also removed two obsolete parts of code which are no more needed:
 - `location_members` table --> this one was used to store relations between locations, but when we have geometries, it is not need
 - location csv data and its importer --> fully replaced by geojson data and its importers

backend/spec/fixtures/snapshots/api/v1/enums.json
```diff
@@ -334,216 +334,230 @@
...
-      "id": "ngo",
+      "id": "incipient",
       "type": "project_developer_type",
       "attributes": {
         "name": "NGO"
       }
     },
     {
-      "id": "iplc-organization",
+      "id": "consolidaton",
```
tsubik, line 420: This snapshot is totally weird.

tsubik: maybe, this is because `id` is treated as dynamic attribute globally in `rails_helper`. And for some endpoints it is not dynamic, as this one for example. Explicit reset dynamic_attributes in snaphost matches could resolve this, but I'm not sure.

backend/db/migrate/20220516085522_add_impact_demands_to_locations.rb
```diff
@@ -0,0 +1,13 @@
+class AddImpactDemandsToLocations < ActiveRecord::Migration[7.0]
+  def change
+    enable_extension "postgis"
```
tsubik, line 3: I'm not sure if that will work with migration on staging (if application user will have enough rights)

tsubik:
Just that weird enums snapshot issue.

backend/db/migrate/20220516085522_add_impact_demands_to_locations.rb
agnessa: In the end `enable_extension` just worked in migration, which is a surprise, but I'm not going to complain. It is true that with newer version of postgres you no longer need superuser to create extensions, but still surprising to see anything *just work* here.

PR #177, opened 2022-05-16
Project details card (search page)

SARodrigues:
This PR adds a project details card to the discover/search project page. 

**In this PR:**  

- New `ProjectDetails` component:   
  - The project details will be visible in the discover/project page when a `ProjectCard is clicked`:  
    - For large screens, a card with the project details will open next to the list of projects  
      _over the map, as per designs_  
    - For small screens, a modal opens with the project details   
- New `useBreakpoint` hook:  
  - Provides a helpers to allow us to selectively add/remove components from the view, based on Tailwind's breakpoints  
    _This was so that we could switch between displaying a ProjectCard over the map or open it in a modal, depending on screen size._
    _Works virtually the same as Tailwind's own helpers. Eg: `breakpoint('lg')` is equivalent to `'lg:'`_
- `ProjectCard` component changes:  
  - Added an `active` prop  
    _So that when the ProjectDetails card is open for a particular ProjectCard, it is shown in a with a border to indicate it is selected/active_    
- `Modal` component changes:  
  - Added possibility of theming and included a `naked` component, where the modal is shown with no padding nor close button
    _This was necessarily primarily to make the "Looking for funding" ribbon_    
- `Tag` and `CategoryTag` components changes:  
  - Added the possibility to set a `size`
    _Necessary because the ProjectDetails card shows smaller tags_   
- Storybook changes:  
  - Added and configured `storybook-addon-mock`
    _So that we can mock API requests._  
   -  Added `api/v1/enums` mock data  
- `ImpactChart`  component changes:  
  - Made it so it is now responsive  
    _This wasn't strictly necessary for this task, but in order to set proper spacing and test the `ProjectDetails` component I've added an `ImpactChart` with mock data. Needed to know the card wouldn't break with different screen sizes._

**Notes:**

- I believe the ProjectDetails card's accessibility could be improved (with aria-live), but as this task was taking so long and has a huge size I haven't looked much into it. 
- The ribbon/close button have been fixed to the top right, which isn't shown on the screenshots/videos

## Testing instructions

- When clicking one of the projects in the search results list, a “card” will be displayed with the following information
  - Title of the project
  - Instrument type(s)
  - Ticket size
  - Category
  - Verified or not
    _(note that we don't have verified projects yet, but it can be tested in Storybook )_
  - Looking for funding (or not)
  - Description (“About”)
  - Project developer name and picture
  - SDGs

- There will be a way to close the preview
- There will be a way to open the project detail page

## Tracking

[LET-399](https://vizzuality.atlassian.net/browse/LET-399)

## Screenshots 

**Desktop:** 
Screenshot 2022-05-19 at 12 18 11

**Mobile:**
Screenshot 2022-05-19 at 12 18 01

## Video  

https://user-images.githubusercontent.com/6273795/169281457-048fea5c-51b8-4b17-b1d8-90ccd897e7d0.mp4

### DIFF
--- a/frontend/.storybook/main.js
+++ b/frontend/.storybook/main.js
@@ -15,6 +15,7 @@ module.exports = {
     '@storybook/addon-essentials',
     'storybook-addon-next',
     'storybook-react-intl',
+    'storybook-addon-mock/register',
     {
       name: '@storybook/addon-postcss',
       options: {
--- a/frontend/components/impact-chart/component.stories.tsx
+++ b/frontend/components/impact-chart/component.stories.tsx
@@ -9,7 +9,7 @@ export default {
 };
 
 const Template: Story<ImpactChartProps> = (args) => (
-  <div className="w-[500px] h-[500px]">
+  <div className="w-96">
     <ImpactChart {...args} />
   </div>
 );
--- a/frontend/components/impact-chart/component.tsx
+++ b/frontend/components/impact-chart/component.tsx
@@ -21,7 +21,7 @@ import { ImpactChartProps } from './types';
 
 ChartJS.register(RadialLinearScale, PointElement, LineElement, Filler, Tooltip);
 
-export const ImpactChart: FC<ImpactChartProps> = ({ category, impact = [] }) => {
+export const ImpactChart: FC<ImpactChartProps> = ({ className, category, impact = [] }) => {
   const {
     data: { impact: impacts, category: categories },
   } = useEnums();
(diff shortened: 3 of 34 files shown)


clementprdhomme (commented):
I think the only accessibility issue I can find is that the project detail card is not focused when it is opened. I think the easiest way to fix this would be to use an overlay from `react-aria` (the basis of the modals). That would trap the focus within the card.

Good job Simão!

PR #180, opened 2022-05-17
LET-244 - Configure Google Cloud Tasks

tiagojsag:

- You ran the tests
    
---

Description of the PR

## Testing instructions

What to test? How to do it?

## Tracking

Link to the task(s), if any


backend/config/initializers/cloudtasker.rb
```diff
@@ -53,6 +53,7 @@
   # Defaults to :development unless CLOUDTASKER_ENV or RAILS_ENV or RACK_ENV is set to something else.
   #
   # config.mode = Rails.env.production? || Rails.env.my_other_env? ? :production : :development
+  config.mode = :production
```
agnessa, line 56: it should probably be:
`config.mode = Rails.env.production? ? :production : :development`
so that it continues working using local redis in development

or it should read from RAILS_ENV, but I no longer know if we have that available

PR #182, opened 2022-05-17
Let 287 fe prevent non signed in users and investors to access the project creation form

barbara-chaves:
This PR changes the ProtectedPage layout and adds it to the project creation page and investor account creation page

## Testing instructions

When trying to go to projects/new:

* A non signed in user should be redirected from projects/new to sign in page
* An investor user should be redirected to the last page (back)
* A light user should be redirected to sign-up/account-type

## Tracking

[LET-287](https://vizzuality.atlassian.net/browse/LET-287) 
[LET-420](https://vizzuality.atlassian.net/browse/LET-420)

### DIFF
--- a/frontend/layouts/protected-page/component.tsx
+++ b/frontend/layouts/protected-page/component.tsx
@@ -9,20 +9,22 @@ import { ProtectedProps } from 'layouts/protected-page/types';
 
 const Protected: React.FC<ProtectedProps> = ({ permissions, children, ...rest }) => {
   const router = useRouter();
-  const { user, isLoading } = useMe();
+  const { user, isLoading, isError } = useMe();
 
   // Not display anything when me request is on progress
   if (isLoading) return null;
 
-  // Redirect when session doesn't exist
-  if (!user) {
+  // Redirect to sign-in when session doesn't exist
+  if (isError) {
     router.push(Paths.SignIn);
     return null;
   }
+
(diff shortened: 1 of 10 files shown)


frontend/layouts/protected-page/component.tsx
```diff
@@ -4,18 +4,28 @@ import { useRouter } from 'next/router';
 
 import useMe from 'hooks/me';
 
+import { Paths, UserRoles } from 'enums';
 import { ProtectedProps } from 'layouts/protected-page/types';
 
-const Protected: React.FC<ProtectedProps> = ({ children, ...rest }) => {
+const Protected: React.FC<ProtectedProps> = ({ permissions, children, ...rest }) => {
```
clementprdhomme, line 10: Could you also take care of the `isError` case?

clementprdhomme (commented):
Did Simão and you discuss whether SSR would have been a better solution? You could directly redirect from the server instead of in the client: https://nextjs.org/docs/api-reference/data-fetching/get-server-side-props#redirect

barbara-chaves:
> Did Simão and you discuss whether SSR would have been a better solution? You could directly redirect from the server instead of in the client: https://nextjs.org/docs/api-reference/data-fetching/get-server-side-props#redirect

It is not possible to use the serverSideProps because we need to be on the client to get the user information.

clementprdhomme (approved):

PR #189, opened 2022-05-18
Adds cloudtranslate.generalModels.batchPredict permission to service account

agnessa:
https://vizzuality.atlassian.net/browse/LET-472

I used this https://cloud.google.com/iam/docs/permissions-reference

To search for roles which include the `cloudtranslate.generalModels.batchPredict` permission, about which we're getting an error. It seems that the lowest priority role with that permission is `Cloud Translation API User (roles/cloudtranslate.user)`, so I added that to the service account.


agnessa:
I managed to get terraform apply to work with this, it adds the role as expected and that removes the error I was seeing

PR #193, opened 2022-05-19
Feature/let 388

agnessa:
Updated investor columns according to new requirements

Also set application default host in test env, not sure why this is needed all of a sudden, but most specs where failing with  'Missing host to link to'

## Tracking

https://vizzuality.atlassian.net/browse/LET-388?atlOrigin=eyJpIjoiYWYwZjk1YTZkMjQ0NGEwZDk4NDA0ZTU0NjM5ZjE2MDUiLCJwIjoiaiJ9


agnessa:
hold on with review, I'll move mission to account now that it's in both investor and PD

backend/config/environments/test.rb
```diff
@@ -61,3 +61,6 @@
 
   config.active_job.queue_adapter = :test
 end
+
+# Set the default host and port to be the same as Action Mailer.
+Rails.application.default_url_options = Rails.application.config.action_mailer.default_url_options
```
martintomas, line 66: This is probably caused by https://github.com/Vizzuality/heco-invest/pull/168. The `default_url_options` is setup by `BACKEND_URL` which is parsed by `URI.parse` at first, which helps to extra `host`, `port` and `scheme` from it. I would check if your  `BACKEND_URL` variable at your local machine is parsable by `URI.parse`, maybe this will help ;).
By the way, @tsubik already mentioned this, but `BACKEND_URL` and `MAILER_HOST` are kind of duplicate, maybe we could keep just one -- vote to keep `BACKEND_URL`

agnessa: OK, thanks! Sorry I lost track of that

agnessa:
I extracted the work related to moving mission into accounts to a separate PR, because it ended up more work than I thought and I still haven't finished, while these changes are blocking FE

martintomas (approved):
looks good to me ;)

PR #194, opened 2022-05-20
Refactoring/let 479

agnessa:
https://vizzuality.atlassian.net/browse/LET-479


backend/app/services/api/filterer.rb
```diff
@@ -21,7 +21,7 @@ def call
     private
 
     def apply_full_text_filter
-      return if filters[:full_text].blank?
+      return if filters[:full_text].blank? || !query.klass.respond_to?(:translatable_attributes)
```
martintomas, line 24: Yeah, this change is unfortunately not enough. It will turn off full text search for PD and Investor models, but same type of searching needs to be done at Account table --> I believe Account needs to be `Searchable` and this service needs to be a bit tweaked so it takes into account which attributes are at Account table and which are at original query table (something similar is done at Sorting service).

To be honest, not sure if moving mission to Account table is worth it now :).

martintomas: So in the end, I had go for this implementation as well -- please see https://github.com/Vizzuality/heco-invest/pull/203

agnessa:
I'll just close this, no patience for this right now and it won't add value

PR #195, opened 2022-05-20
feat: Impact calculation service

martintomas:
This MR brings new `Imports::CalculateForProject ` service, which is responsible for calculation of all impacts for all impact types and its dimensions. I have basically rewritten logic used at https://docs.google.com/spreadsheets/d/1KFYcduykY_mst41UKgafSAaHFvxp68wMRR0zGfdQkDs/edit#gid=1744503386 and all tests of this service are based on results computed by this sheet. Please @ikerey , I will add you as reviewer to this PR, so if you could just briefly check this service if you see some issues (hopefully, it is readable enough even for non-ruby person). New columns of Project used to store computed impacts are:
 - municipality_biodiversity_impact
 - municipality_climate_impact
 - municipality_water_impact
 - municipality_community_impact
 - municipality_total_impact
 - hydrobasin_biodiversity_impact
 - hydrobasin_climate_impact
 - hydrobasin_water_impact
 - hydrobasin_community_impact
 - hydrobasin_total_impact
 - priority_landscape_biodiversity_impact
 - priority_landscape_climate_impact
 - priority_landscape_water_impact
 - priority_landscape_community_impact
 - priority_landscape_total_impact

I save computed results as decimals, but from design, it looks like it will be in the end multiplied by `100 (to get %)` and rounded -- can be always done later on or by FE itself. All these variables are already part of Project API serializer and based on these values, sorting of Projects can be done.

I have slightly tweaked `backend/app/models/impact_area.rb` enums -- addition of `hydrometerological-risk-reduction` and `water-capacity` and `water-efficiency` were merged to one option called `water-capacity-or-efficiency`.

This MR also replaces `latitude` and `longitude` columns of Project model with `center` column which is of `point` type (lets use postgis when we already have it :)).

## Testing instructions

`ImpactCalculationJob ` should be automatically enqueued when you modify `geometry` or `impact_areas` of Project. Please @agnessa could you check if queue of `ImpactCalculationJob ` is correct, I am not really sure what should be there or if jobs already work at Stage. After job finishes, new impact attributes at Project record should be filled with correct values.

My proposal how to tests this would be following:
 - open spreadsheet and setup demands and impacted areas of project some random way
 - find some suitable `project` at db and setup `impact_areas` to same values as in spreadsheet
 - find location which this project intersect with and setup `demand` attributes based on spreadsheet
 - run `ImpactCalculationJob ` job and after it finishes, make sure that impact attributes of `project` correspond to values at spreadsheet

## Tracking

https://vizzuality.atlassian.net/browse/LET-382
https://vizzuality.atlassian.net/browse/LET-429


ikerey:
Hi @martintomas,
How did you compute the _total_impact_?
There shouldn't be a total impact per region type?

martintomas:
> Hi @martintomas, How did you compute the _total_impact_? There shouldn't be a total impact per region type?

`total_impact` is average of all impacts @ikerey (`project.total_impact = impacts.values.sum / impacts.size`). If formula is different, please let me know, I have improvised here, because there is no total impact at excel. Just total impacts for `municipality`, `hydrobasin` and `priority landscape`. Or should we store three different total impacts (one per type)?

ikerey:
> > Hi @martintomas, How did you compute the _total_impact_? There shouldn't be a total impact per region type?
> 
> `total_impact` is average of all impacts @ikerey (`project.total_impact = impacts.values.sum / impacts.size`). If formula is different, please let me know, I have improvised here, because there is no total impact at excel. Just total impacts for `municipality`, `hydrobasin` and `priority landscape`. Or should we store three different total impacts (one per type)?

Yes, I think we have to store a total impact per type.

martintomas:
> > > Hi @martintomas, How did you compute the _total_impact_? There shouldn't be a total impact per region type?
> > 
> > 
> > `total_impact` is average of all impacts @ikerey (`project.total_impact = impacts.values.sum / impacts.size`). If formula is different, please let me know, I have improvised here, because there is no total impact at excel. Just total impacts for `municipality`, `hydrobasin` and `priority landscape`. Or should we store three different total impacts (one per type)?
> 
> Yes, I think we have to store a total impact per type.

So we will not have `total_impact` per Project?

agnessa:
> > > > Hi @martintomas, How did you compute the _total_impact_? There shouldn't be a total impact per region type?
> > > 
> > > 
> > > `total_impact` is average of all impacts @ikerey (`project.total_impact = impacts.values.sum / impacts.size`). If formula is different, please let me know, I have improvised here, because there is no total impact at excel. Just total impacts for `municipality`, `hydrobasin` and `priority landscape`. Or should we store three different total impacts (one per type)?
> > 
> > 
> > Yes, I think we have to store a total impact per type.
> 
> So we will not have `total_impact` per Project?

I believe it's possible we need a total per type; perhaps one of them is default (to be used where not obvious which type is selected?)

agnessa:
As for the jobs queue, I'd add  `queue_as ENV["CLOUD_TASKS_TEST_QUEUE_NAME"].to_sym` like in the other jobs; for now we're only using 1 queue, so potentially this could go into ApplicationJob as well. The problem is the name of this queue is not necessarily "default", and so to avoid an exception we need the queue name specified.

backend/app/validators/project_geometry_validator.rb
```diff
@@ -27,8 +27,7 @@ def decode_geo_json
   def validation_of(geometry)
     return if geometry.blank?
 
-    center = GeoJsons::ToCentroid.new(geometry).call
-    @project.latitude, @project.longitude = [center.y, center.x]
+    @project.center = GeoJsons::ToCentroid.new(geometry).call
```
tsubik, line 30: I think it would be better to name it `centroid`, because that is the common name we always use. Also, `project.center` does not say much, better `project.centroid` is more descriptive.

backend/app/services/api/sorter.rb
```diff
@@ -15,7 +30,7 @@ def initialize(query, sorting_by:, language: I18n.locale)
     end
 
     def call
-      self.query = query.joins(sort_joins).order(sort_query => sort_direction)
+      self.query = query.joins(sort_joins).order(sort_query => sort_direction) if sorting_column_exists?
```
tsubik, line 33: :+1:

backend/app/services/api/sorter.rb
```diff
@@ -5,6 +5,21 @@ class Sorter
     SORTING_DIRECTIONS = %i[asc desc]
     SORTING_OPTIONS = {
       name: {ProjectDeveloper => :account, Investor => :account},
+      municipality_biodiversity_impact: {},
```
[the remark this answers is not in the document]
martintomas: If I read https://vizzuality.atlassian.net/browse/LET-375 correctly, there is requirement to be able to sort by all 4 dimensions, I will clarify this...

backend/db/migrate/20220518072022_add_center_to_projects.rb
```diff
@@ -0,0 +1,19 @@
+class AddCenterToProjects < ActiveRecord::Migration[7.0]
+  def change
+    add_column :projects, :center, :st_point
```
tsubik, line 3: As I say, I would rename it to `centroid`, what do you think @agnessa?

backend/db/migrate/20220518133729_add_impacts_to_projects.rb
```diff
@@ -0,0 +1,19 @@
+class AddImpactsToProjects < ActiveRecord::Migration[7.0]
+  def change
+    add_column :projects, :municipality_biodiversity_impact, :decimal, precision: 25, scale: 20
```
tsubik, line 3: I'm not sure what was the current plan, but at the beginning we designed impacts to be in a separate table (and I think that is on DB design diagram), as those changes over time. There was a plan that was removed from the scope of MVP to be able to draw impact history timeline, but as I remember, the project should keep old impacts if they wanted to add the timeline in the second phase. I might be mistaken. cc @agnessa

tsubik (commented):
Calculation service looks great, just couple comments about saving impacts and centroid name.

backend/app/validators/project_geometry_validator.rb
martintomas: No problem, I will rename it ;)

backend/app/services/impacts/calculate_for_project.rb
```diff
@@ -0,0 +1,87 @@
...
+
+      values = CALCULATION_SETUP.map do |impact, impact_areas|
+        demand = location.public_send "#{impact}_demand"
+        impact_value_for(impact_areas, demand).tap { |value| impacts["#{impact_type}_#{impact}_impact"] = value }
+      end
+      impacts["#{impact_type}_total_impact"] = values.sum / values.size
+    end
+
+    def impact_value_for(impact_areas, demand)
+      return 0 if (project.impact_areas & impact_areas.keys) == impact_areas.keys || demand.nil?
```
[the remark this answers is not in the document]
martintomas: Yup, that condition comes from excel sheet, will add comment

backend/db/migrate/20220518133729_add_impacts_to_projects.rb
martintomas: I would personally keep it simple at this phase and moved it to separate table when we will know more at future -- otherwise it will add extra complexity to sorting and calculation service at this point, which is not necessary right now

backend/app/services/impacts/calculate_for_project.rb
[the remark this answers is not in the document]
tsubik: Is it in spec also? Maybe worth to add it, I did not see it @martintomas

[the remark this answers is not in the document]
martintomas: It should be there -- impacts `conservation, restoration and pollutants-reduction` are all options from `biodiversity` category, therefore its impact is 0

ikerey (approved):

PR #197, opened 2022-05-20
LET-381 updates to project serializer

agnessa:
https://vizzuality.atlassian.net/browse/LET-381?atlOrigin=eyJpIjoiYzFjZThjYjMwNzVjNGRlN2IzNDY2YjBkZmJiN2MxMjciLCJwIjoiaiJ9

I removed conditional presence validations on received funding fields, since I understand from the excel that those fields are not mandatory even if received funding flag is set.

Screenshot 2022-05-20 at 15 11 39

I added funding plan to the serializer, and also country / department / municipality.

Finally, I added `trusted` to both project and open call serializer.

### DIFF
--- a/backend/app/models/project.rb
+++ b/backend/app/models/project.rb
@@ -58,10 +58,6 @@ class Project < ApplicationRecord
     validates_presence_of :funding_plan
   end
 
-  with_options if: -> { received_funding? } do
-    validates_presence_of :received_funding_amount_usd, :received_funding_investor
-  end
-
   validates_uniqueness_of [*locale_columns(:name)], scope: [:project_developer_id], case_sensitive: false, allow_blank: true
   validate :location_types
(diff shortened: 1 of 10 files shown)


martintomas (approved):

martintomas:
> actually I think instead of department we need to return the mosaic, how do we do it - municipality.regions?

Yes, you are right that to obtain list of mosaics which intersect with project municipality, you can call `municipality.regions`, BUT I believe that we agreed not to use this approach and actually https://github.com/Vizzuality/heco-invest/pull/175 removes this option completely. On other hand, same PR adds municipality and mosaic geometries and https://github.com/Vizzuality/heco-invest/pull/195 ands new scope `intersection_with` for obtaining locations which intersect with provided geometry. From there, it is already quite straightforward to obtain list of mosaics based on intersection between project and mosaic geometries ;).

PR #203, opened 2022-05-23
fix: Full text searching takes into account attributes at different tables (Account)

martintomas:
Full text searching needs to take into account some attributes which are stored at different table -- for example `Account` table keeps `name` of `Investor` or `ProjectDeveloper`. 

I have slightly tweaked logic behind filtering so you can specify which extra tables (relations) should be taken into account during full text searching. In the end, searching is run on top of all defined tables and final result is union of these subqueries. I have split queries to small ones and pass `ids` between them instead of constructing one huge and slow query. It causes that 1 (for queries with Account relation it is 3) extra db hits are done which is quite insignificant from my point of view.

Following tweak also fixes problems with localization, so it takes into account also columns which are not localized such as `name` of ProjectDeveloper.

And one last bug was related to full text searching interfering with sorting --> full text searching query ordered results by weight, which was naturally fixed by using individual subqueries and passing just `ids`.

## Testing instructions

Prepare two Investor/Project Developer records and try to run full text search based on their `name` value.

## Tracking

https://vizzuality.atlassian.net/browse/LET-325 (bug fix ;))

### DIFF
--- a/backend/app/models/account.rb
+++ b/backend/app/models/account.rb
@@ -1,6 +1,7 @@
 class Account < ApplicationRecord
   extend FriendlyId
   include Translatable
+  include Searchable
 
   friendly_id :name, use: :slugged
 
--- a/backend/app/services/api/filterer.rb
+++ b/backend/app/services/api/filterer.rb
@@ -3,6 +3,10 @@ module API
     attr_accessor :query, :filters, :language
 
     FULL_TEXT_FILTERS = %i[name description about mission problem solution expected_impact]
+    FULL_TEXT_EXTRA_TABLES = {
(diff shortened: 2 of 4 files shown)


agnessa (approved):
OK, not a fan of potentially huge IN conditions, but it will work for now.

martintomas:
> OK, not a fan of potentially huge IN conditions, but it will work for now.

There are no extra IN conditions, except of one enforced by pg_search gem. I also dont like them so I use just simple ´pluck´ to get ids which I use in final query. This way I avoid quite complex queries with IN stuff, but price is extra db hit (but for much smaller query), which I think is worthy in this case :)

PR #211, opened 2022-05-24
Let 421 fe investor public page

barbara-chaves:
This PR adds the investor public page

## Testing instructions

testing link: https://heco-invest-frontend-qz2nvuffe-vizzuality1.vercel.app/en/investor/test-investor-3

It should respect the [design](https://www.figma.com/file/KC49uMR9t1YARmFR0Sr6aN/HeCo---Work-in-progress?node-id=36%3A15506)

Screenshot 2022-05-24 at 18 09 29

It does NOT includes the 'add to favorites' task. This task is not created and the endpoints are not available.

## Tracking

[LET-421](https://vizzuality.atlassian.net/browse/LET-421)


frontend/pages/investor/[id]/index.tsx
```diff
@@ -15,104 +18,128 @@ import LayoutContainer from 'components/layout-container';
...
     {
       title: 'Invests in',
       type: 'category',
-      tags: [
-        { id: 'tourism-and-recreation', name: 'Tourism & Recreation' },
-        { id: 'non-timber-forest-production', name: 'Non-timber forest production' },
-        { id: 'sustainable-agrosystems', name: 'Sustainable agrosystems' },
-        { id: 'forestry-and-agroforestry', name: 'Forestry & agroforestry' },
-      ],
+      tags: category.filter(({ id }) => categories.includes(id)),
```
Comment (SARodrigues) [inline frontend/pages/investor/[id]/index.tsx:103]: > `categories.includes(id))`

Looking at the schema in Swagger, I believe `categories` is not mandatory. If the endpoint returns `null` this will crash. 
I'd replace this with `categories.includes(id))`.

Same for `impacts` and `instrument_sizes` below (might as well do them all to foolproof it, in case required/non-required fields change in the future).

frontend/pages/investor/[id]/index.tsx
```diff
@@ -15,104 +18,128 @@ import LayoutContainer from 'components/layout-container';
...
+    impacts,
+    sdgs,
+    about,
+    mission,
+    investor_type,
+    other_information,
+    language,
+    prioritized_projects_description,
+  } = investor;
+  const { category, ticket_size, instrument_type, impact, investor_type: investor_types } = enums;
```
Comment (SARodrigues) [inline frontend/pages/investor/[id]/index.tsx:79]: Nothing wrong with the current implementation, but I'd replace this with: 

```
const { category: allCategories, ticket_size: allTicketSizes, instrument_type: allInstrumentTypes, impact: allImpacts, investor_type: allInvestorTypes } = enums;
````

Just to improve code readability. I found it confusing to read the rest of the [code below](https://github.com/Vizzuality/heco-invest/pull/211/files#diff-473df3881d2d05ea0ec5a8d66693c9202b3c48820f305246b4f80d8eec3d5d73R103) as at a glance it was not very clear which array  was coming from the `enums` and which one from the `investor` endpoint.

SARodrigues:
Looking great!

A couple more things I've noticed with a couple investor profiles that didn't have blank fields at the time I was going through staging:   

&nbsp;

1. In the content, some fields are also optional (eg: `prioritized_projects_description`). I'd add a check to make sure they exist and not display the section/title if there's no content for the field, to prevent section titles with no information:  

investor_no_info_2

&nbsp;  

2. If the user hasn't added an website/social information (those are optional), the spacing in header will look weird. Note that this component is shared with the ProjectDeveloper page (which doesn't suffer from the same issue, because there's always information at the top of that contact box), but changes made to correct this on the Investor public page need to make sure the display doesn't break for the project developer public page. 

investor_no_info

SARodrigues (approved):

PR #212, opened 2022-05-24
Feature/Admin sign in

tsubik:
Admin can sign in to back office. I created a temp dashboards controller for now.

image

The styles are not exact as per design yet, but we will enhance that later.

## Testing instructions

Backoffice url: http://localhost:4000/backend/backoffice

## Tracking

https://vizzuality.atlassian.net/browse/LET-482
https://vizzuality.atlassian.net/browse/LET-485


backend/Dockerfile
```diff
@@ -15,6 +15,7 @@ RUN apk update && \
       postgresql-dev \
       imagemagick \
       geos-dev \
+      yarn \
```
martintomas, line 18: so we are going for yarn? If I remember correctly, Rails 7 adds new support for impartmaps, which should be simplification, but I don't mind yarn stuff

backend/app/views/admins/sessions/new.html.erb
```diff
@@ -0,0 +1,22 @@
+<div class="bg-white p-6 rounded-2xl shadow-lg min-w-[600px]">
+  <h2 class="text-green-dark font-bold text-lg">Welcome to HeCO Invest backoffice.</h2>
+
+  <p class="my-6 text-gray-600">Please enter your details below.</p>
+
+  <%= simple_form_for(resource, as: resource_name, url: session_path(resource_name)) do |f| %>
+    <% flash.each do |type, message| %>
```
martintomas, line 7: maybe special partial for flash messages?

backend/app/views/admins/sessions/new.html.erb
```diff
@@ -0,0 +1,22 @@
+<div class="bg-white p-6 rounded-2xl shadow-lg min-w-[600px]">
+  <h2 class="text-green-dark font-bold text-lg">Welcome to HeCO Invest backoffice.</h2>
+
+  <p class="my-6 text-gray-600">Please enter your details below.</p>
```
martintomas, line 4: If we ever plan to use locales, all text should go to I18n files...

backend/spec/system/backoffice/auth_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  let_it_be(:admin) { create(:admin, email: "admin@example.com", password: "SuperSecret6", first_name: "Admin", last_name: "Example") }
+
+  describe "Login" do
+    before { visit "/backoffice" }
+
+    it "authenticate user successfuly" do
+      fill_in :admin_email, with: "admin@example.com"
+      fill_in :admin_password, with: "SuperSecret6"
+
+      click_on "Sign in"
```
martintomas, line 13: just habit of mine, but used to put all these `fill_in` or `click_on` to `before` action and keep only expectations at `it` block

backend/spec/system_helper.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  config.prepend_before(:each, type: :system) do
+    driven_by Capybara.javascript_driver
+  end
+
+  config.before(:suite) do
+    Rails.application.load_tasks
+    `yarn build`
+    `yarn build:css`
+    Rake::Task["assets:precompile"].execute
```
martintomas, line 43: hmm, didn't this slow down tests a lot. I often run just 1-2 tests during development, but I run them a lot so every second which I spend waiting for tests to finish count. Would leave it to pipeline to run this, but I understand that you wanna ensure that FE is up to date before running tests

martintomas (approved):
Looks good, really liked simple_form and extension for tailwind ;)

backend/spec/system_helper.rb
tsubik: Yeah, I think that was necessary, took it from other project. I want to avoid the pain :D.

backend/app/views/admins/sessions/new.html.erb
tsubik: Does not need it at this moment, maybe when adding forgot password stuff. This is the quirk of devise that is using flash messages, and not kind of form objects with errors collection for error messages.

tsubik: So it is very specific only to this place.

backend/Dockerfile
tsubik: Feel free to refactor this at any moment. I would love to see it with importmaps. I did not have time to check how to configure it with and remove all those packages from package.json. Just doing, the way I know it works at this moment, to do it quickly.

tsubik:
The only problem I see is that we are running tests twice. First time here with GH actions (which is great and not a problem), but the second time using Cloud Build, which is running specs in container. I think I spent enough time on this to try to configure Chrome for headless run on Linux alpine, especially that testing docker build and tests locally takes time (as rebuild takes time, especially some google cloud gem installs). Maybe for now I would just not run system specs in google cloud build? Thoughts @martintomas @agnessa ?

backend/spec/system_helper.rb
tsubik: I don't actually remember what was the problem, but I will try to check if I left any clues in at least commit messages.

backend/spec/system/backoffice/auth_spec.rb
tsubik: I prefer to not write system specs in very BDD way, rspec style. Just to have a minimum number of `it` blocks. Better to add multiple expectation and test more in one block than have a rspec structure, because the specs will eventually be very slow if we add too many of them.

martintomas:
> The only problem I see is that we are running tests twice. First time here with GH actions (which is great and not a problem), but the second time using Cloud Build, which is running specs in container. I think I spent enough time on this to try to configure Chrome for headless run on Linux alpine, especially that testing docker build and tests locally takes time (as rebuild takes time, especially some google cloud gem installs). Maybe for now I would just not run system specs in google cloud build? Thoughts @martintomas @agnessa ?

I don't have problem with it -- not sure why we run tests during Cloud build, GH actions are enough for me ;)

agnessa:
I'm probably doing a few things wrong here, but here's where I got to :)

When I go to http://localhost:4000/backend/backoffice, it tries to redirect to the sign in, but it ends up at http://localhost:4000/backoffice/sign_in (no `backend`...)

If I go to http://localhost:4000/backend/backoffice/sign_in, I get
Screenshot 2022-05-25 at 10 48 29

😅

tsubik:
That is weird. Are you running with bin/dev? To complie assets for
development.

agnessa:
> bin/dev

Obviously I wasn't, thanks :) Now it works, except potentially for the redirect to sign in when you're not logged in.

tsubik:
Let me check it.

tsubik:
I cannot reproduce it on my machine :D @agnessa

agnessa (approved):

PR #217, opened 2022-05-25
Map uses custom Mapbox' basemap

SARodrigues:
## Description

This PR adds the custom Mapbox basemap to the map component, as well as updating the frontend `README.md` & `.env.sample` to describe the new env variable for Mapbox's public access token.  

## Testing instructions

Visit `/discover/projects` and verify that:  
- The map works  
- The map uses the new basemap style  

## Tracking

[LET-407](https://vizzuality.atlassian.net/browse/LET-407)

## Screenshot

Screenshot 2022-06-08 at 12 41 34

Screenshot 2022-06-08 at 12 43 36


SARodrigues:
🤔 Doesn't seem to be working on `staging` , will take a look tomorrow.

clementprdhomme:
Any reason why you removed MapLibre @SARodrigues? Mapbox GL.js 1.x is not supported by Mapbox anymore and 2.x require an active subscription to be used (and is not free anymore).

SARodrigues:
> Any reason why you removed MapLibre @SARodrigues? Mapbox GL.js 1.x is not supported by Mapbox anymore and 2.x require an active subscription to be used (and is not free anymore).

@clementprdhomme Mistake on my part. Since we're using Mapbox Studio and mapbox url schemes are not supported by MapLibre major version 2 I assumed we were moving back to Mapbox. 

I've added back MapLibre and and fixed MapLibre's major version to 1.

frontend/package.json
```diff
@@ -102,6 +102,7 @@
     "react-redux": "^7.2.4",
     "react-select": "^4.3.0",
     "react-stately": "^3.12.2",
+    "remove": "^0.1.5",
```
clementprdhomme, line 105: I think that's a mistake 😉.

frontend/components/map/component.tsx
```diff
@@ -1,7 +1,6 @@
 import { useEffect, useState, useRef, useCallback, FC } from 'react';
 
 import ReactMapGL, { FlyToInterpolator, TRANSITION_EVENTS } from 'react-map-gl';
```
clementprdhomme, line 3: Below this line, there used to be this one, which I think is necessary:

```tsx
import 'maplibre-gl/dist/maplibre-gl.css';
```

frontend/components/map/constants.ts
```diff
@@ -1,38 +0,0 @@
...
-      maxzoom: 22,
-    },
-    // Next layer is needed by the layer-manager to know it can draw layers on top of it
-    {
-      id: 'custom-layers',
-      type: 'background',
-      paint: {
-        'background-opacity': 0,
-      },
-    },
```
clementprdhomme, line 36: This layer must exist on Mapbox Studio (as part of the style we pass to the map), otherwise we won't be able to use the layer-manager.

You can see that if you try to create a project drawing an area, an error is thrown when the area is closed. This is because this layer is missing.

Can you check with Andreia to add this layer?

clementprdhomme (changes requested):

frontend/components/map/component.tsx
SARodrigues: I've tried with and without the import, it doesn't seem to make a difference.

frontend/components/map/constants.ts
SARodrigues: Layer has been added. We can now draw on the map :)

PR #218, opened 2022-05-25
Feature/backoffice project developers

tsubik:
image

Approve/reject buttons for project developers with just simple table design.

## Testing instructions

What to test? How to do it?

## Tracking

https://vizzuality.atlassian.net/browse/LET-488
https://vizzuality.atlassian.net/browse/LET-490

### DIFF
--- a/backend/app/views/backoffice/project_developers/index.html.erb
+++ b/backend/app/views/backoffice/project_developers/index.html.erb
@@ -2,22 +2,22 @@
   <thead>
     <tr>
       <th>
-        Name
+        <%= t("backoffice.name") %>
       </th>
       <th>
-        Account owner
+        <%= t("backoffice.account_owner") %>
       </th>
       <th>
-        Account users
+        <%= t("backoffice.account_users") %>
       </th>
       <th>
-        Projects
+        <%= t("backoffice.projects") %>
       </th>
       <th>
-        Language
+        <%= t("backoffice.language") %>
       </th>
       <th>
-        Status
+        <%= t("backoffice.status") %>
       </th>
       <th>
       </th>
@@ -33,8 +33,8 @@
         <td><%= pd.language %></td>
         <td><%= pd.account.review_status %></td>
         <td>
-          <%= button_to "Approve", approve_backoffice_account_path(pd.account.id), class: "link-button", method: :post unless pd.account.approved? %>
-          <%= button_to "Reject", reject_backoffice_account_path(pd.account.id), class: "link-button", method: :post unless pd.account.rejected? %>
+          <%= button_to t("backoffice.approve"), approve_backoffice_account_path(pd.account.id), class: "link-button", method: :post unless pd.account.approved? %>
+          <%= button_to t("backoffice.reject"), reject_backoffice_account_path(pd.account.id), class: "link-button", method: :post unless pd.account.rejected? %>
         </td>
       </tr>
     <% end %>
--- a/backend/app/views/backoffice/project_developers/index.html.erb
+++ b/backend/app/views/backoffice/project_developers/index.html.erb
@@ -1,4 +1,4 @@
-<table>
+<table class="table">
   <thead>
     <tr>
       <th>
@@ -33,8 +33,8 @@
         <td><%= pd.language %></td>
         <td><%= pd.account.review_status %></td>
         <td>
-          <%= button_to "Approve", approve_backoffice_account_path(pd.account.id), method: :post unless pd.account.approved? %>
-          <%= button_to "Reject", reject_backoffice_account_path(pd.account.id), method: :post unless pd.account.rejected? %>
+          <%= button_to "Approve", approve_backoffice_account_path(pd.account.id), class: "link-button", method: :post unless pd.account.approved? %>
+          <%= button_to "Reject", reject_backoffice_account_path(pd.account.id), class: "link-button", method: :post unless pd.account.rejected? %>
         </td>
       </tr>
     <% end %>
--- a/backend/app/views/backoffice/project_developers/index.html.erb
+++ b/backend/app/views/backoffice/project_developers/index.html.erb
@@ -2,22 +2,22 @@
   <thead>
     <tr>
       <th>
-        <%= t("backoffice.name") %>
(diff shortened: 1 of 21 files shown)


backend/spec/system/backoffice/auth_spec.rb
```diff
@@ -32,9 +32,9 @@
 
     it "works well" do
       visit "/backoffice"
-      expect(page).to have_text("Back office dashboard")
+      expect(page).to have_text("Project developers")
 
-      click_on "Sign Out"
+      click_on "Sign out"
```
martintomas, line 37: maybe when we already use localized text at app, we should also use it at tests :)

backend/spec/system/backoffice/project_developers_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+        expect(page).to have_text("Tom Higgs")
+        expect(page).to have_text("approved")
+      end
+      within_row("Unapproved PD Enterprise") do
+        expect(page).to have_text("John Levis")
+        expect(page).to have_text("unapproved")
+      end
+    end
+
+    context "Approve" do
```
martintomas, line 26: for me personally, context with text: `when approving user` makes a bit more sense here

backend/app/views/backoffice/project_developers/index.html.erb
```diff
@@ -0,0 +1,42 @@
...
+      </th>
+    </tr>
+  </thead>
+  <tbody>
+    <% @project_developers.each do |pd| %>
+      <tr>
+        <td><%= pd.name %></td>
+        <td><%= pd.account.owner.full_name %></td>
+        <td><%= pd.account.users.count + 1 %></td>
+        <td><%= pd.projects.count %></td>
```
martintomas, line 32: I guess this line and previous one will trigger db query per project developer

backend/config/locales/zu.yml
```diff
@@ -1,4 +1,19 @@
...
+    sign_in: Sign in
+    name: Name
+    account_owner: Account owner
+    account_users: Account users
+    projects: Projects
+    language: Language
+    status: Status
+    approve: Approve
+    reject: Reject
+    project_developers: Project developers
```
martintomas, line 16: I think it would be worthy to split translations to subcategories (per tables for example) so we don't have everything under one key (`backoffice `)

martintomas (approved):
Looks really great ;)

backend/config/locales/zu.yml
tsubik: Yeah, I was thinking about it, but then you have multiple same strings to translate. Dunno, I'm kinda like the frontend style of doing this, generate random id for unique strings, but they have tools that does that. Can you give me example how would you split it? Per view? Common actions separate?

backend/app/views/backoffice/project_developers/index.html.erb
tsubik: Those counter caches will be good when adding sorting.

backend/config/locales/zu.yml
martintomas: Hmm, that is always difficult to answer and it happens quite often in Rails that there are duplicities at locales, but I would for example go for `general` section which covers these string which are used everywhere and then add key per all other sections --> `project_developers`, `investors`, etc.

backend/spec/system/backoffice/project_developers_spec.rb
tsubik: Ok, will change that :+1:

backend/config/locales/zu.yml
tsubik: Like this? Kinda rails style so lazy lookup for some pages can be used
```
common:
      name: Name
      status: Status
      language: Language
      approve: Approve
      reject: Reject
    sign_in:
      welcome: Welcome to HeCO Invest backoffice.
      enter_details_below: Please enter your details below.
      sign_in: Sign in
    project_developers:
      index:
        name: Name
        account_owner: Account owner
        account_users: Account users
        projects: Projects
        language: Language
        status: :backoffice.common.status
        approve: :backoffice.common.approve
        reject: :backoffice.common.reject
    layout:
      header: Backoffice
      sign_out: Sign out
      project_developers: Project developers
```

martintomas: yp, looks better for sure :), I normally dont use `:backoffice.common.status` and just take it right from there --> `I18n.t("backoffice.common.status")`, but if this works with transifex, no problem with that

agnessa:
We have a `reviewed_at` timestamp which is not updated when approving / rejecting. I think it would be good to keep it and update it.

At the same time, we have a `review_message` field, which was supposed to be for the admin to be able to give user information about reason for rejection. But it seems that there's no provision for that, so we could remove the message field. What do you think?

backend/config/locales/zu.yml
tsubik: Hmm, I had in this way in different project and translator did not mind, as I see it creates extra strings that should not be changed. Maybe as you said use common, but then the whole point of having things separately falls apart for me haha.

martintomas: :D I think this is more about to keep some kind of structure in locales so you don't have 100 words under one key, but it does not have to be so rigid so you can use at some cases translations from different sections <-- but agree that it is not optimal :)

tsubik:
Hmm, Do we have a design for review_message? Maybe we should ask about that and add enhance story. Maybe we will need a modal for rejection to type the message. I would not worry about message now, and will update `reviewed_at` date.

backend/config/locales/zu.yml
tsubik: Ok, let's experiment a little with this. I think we have some time to refactor the structure until translations start. To be honest, I have nothing against having all keys under just `backoffice` section, as not have to think about the structure too much is also a virtue. Or we can just always duplicate stuff.

tsubik:
Ahh, I remember we should also send emails. Should we add enhance stories to reject/approve or keep adding missing acceptance criteria here :D

agnessa:
OK, will add a task to notify the user!

agnessa (approved):

PR #223, opened 2022-05-26
feat: Adding priority landscape to Projects

martintomas:
I am adding `priority_landscape` attribute to Project, which is optional and computed dynamically before save as intersections between `mosaic` locations and Project `centroid`.

Same attribute is added to Project serializer and returned via API endpoints.

## Testing instructions

Update selected Project with geometry which is inside of some chosen mosaic and make sure that apps assigns correct priority_landscape value.

## Tracking

https://vizzuality.atlassian.net/browse/LET-442


tsubik (approved):
Looks good :+1:

agnessa (approved):

PR #224, opened 2022-05-27
feat: Right policies are applied to API endpoints

martintomas:
I am adding `cancancan` library which is responsible for making sure that:
 - only approved Investors and ProjectDevelopers are visible via public endpoints (`index/show` action)
 - user can always see their own records (even though they are not approved --> `show` action)
 - only Projects and OpenCalls of approved accounts are visible via public endpoints

I have also defined couple other rights for `create` and `update` of records, but `account` API endpoints allow user to manipulate only with its own Investor/ProjectDeveloper data so these policies are a bit extra now. Only endpoint which I have tweaked a bit more is one for `account/projects` which uses cancancan policy to make sure that current_user creates/edits only Projects which he has access to.

I have not touched other already existing endpoints, but they can be always enhanced with cancancan policy if necessary. 

## Testing instructions

We already have Backoffice so it is enough to approve/reject via it some of the records and make sure that approved ones are visible and rejected ones are hidden (with exception if data are your own). Everything is ofc covered with tests so maybe you can also trust them :)

## Tracking

https://vizzuality.atlassian.net/browse/LET-512
https://vizzuality.atlassian.net/browse/LET-513

### DIFF
--- a/backend/app/models/ability.rb
+++ b/backend/app/models/ability.rb
@@ -13,12 +13,23 @@ class Ability
   private
 
   def default_rights
+    # only data from approved users are visible
     can %i[index show], ProjectDeveloper, account: {review_status: :approved}
     can %i[index show], Investor, account: {review_status: :approved}
+    can %i[index show], Project, project_developer: {account: {review_status: :approved}}
+    can %i[index show], OpenCall, investor: {account: {review_status: :approved}}
   end
 
   def user_rights
+    can %i[create update], Investor, account_id: user.account_id
+    can %i[create update], ProjectDeveloper, account_id: user.account_id
+    can %i[create update], Project, project_developer: {account_id: user.account_id}
+    can %i[create update], OpenCall, investor: {account_id: user.account_id}
+
+    # users can always see their own data
     can %i[show], ProjectDeveloper, account_id: user.account_id
     can %i[show], Investor, account_id: user.account_id
+    can %i[show], Project, project_developer: {account_id: user.account_id}
+    can %i[show], OpenCall, investor: {account_id: user.account_id}
   end
 end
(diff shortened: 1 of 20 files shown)


agnessa (approved):

backend/app/models/ability.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def default_rights
+    # only data from approved users are visible
+    can %i[index show], ProjectDeveloper, account: {review_status: :approved}
+    can %i[index show], Investor, account: {review_status: :approved}
+    can %i[index show], Project, project_developer: {account: {review_status: :approved}}
+    can %i[index show], OpenCall, investor: {account: {review_status: :approved}}
+  end
+
+  def user_rights
```
tsubik, line 23: I think we need to also make sure that only approved `Investor` and `ProjectDeveloper` can create/update projects and open calls.

tsubik: I think UI prevents creating/updating projects, waiting for approval, please correct me if not. We can make it consistent in this way.

martintomas: I have checked code and don't see such limitation -- even unapproved Investor/Project Developer can create OpenCall/Project. I have also ticket https://vizzuality.atlassian.net/browse/LET-513 which specifies that Projects/OpenCalls from unapproved Investors/ProjectDevelopers are hidden by default, but there is no word that unapproved users cannot create/update it at all

tsubik: Maybe it was forgotten, what do you think @agnessa? Can unapproved accounts do anything on the platform?

agnessa: do we? that's not in acceptance criteria :) I'm not sure about that actually

tsubik: Ok. Let's then ask questions and add a new story if needed.

tsubik (approved):

PR #227, opened 2022-05-30
fix social media validation schemmas

barbara-chaves:
This PR fixes the validation schemas for the social media links on the investor and project-developer creation forms

## Testing instructions

The forms should accept real social media links 

## Tracking

BUG [LET-476](https://vizzuality.atlassian.net/browse/LET-476)

### DIFF
--- a/frontend/schemas/investor.ts
+++ b/frontend/schemas/investor.ts
@@ -79,22 +79,22 @@ export default (page: number) => {
       facebook: string().test(
         'isSocialMediaLink',
         messages.social_medias,
-        (value) => !value || !!value.match(/https:\/\/www.facebook.com\/.*/)
+        (value) => !value || !!value.match(/https?:\/\/(www.)?facebook.com\/.*/)
       ),
       linkedin: string().test(
         'isSocialMediaLink',
         messages.social_medias,
-        (value) => !value || !!value.match(/https:\/\/www.linkedin.com\/.*/)
+        (value) => !value || !!value.match(/https?:\/\/(www.)?linkedin.com\/.+/)
       ),
       instagram: string().test(
         'isSocialMediaLink',
         messages.social_medias,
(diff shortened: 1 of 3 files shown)


frontend/schemas/investor.ts
```diff
@@ -79,17 +79,17 @@ export default (page: number) => {
...
         'isSocialMediaLink',
         messages.social_medias,
-        (value) => !value || !!value.match(/https:\/\/linkedin.com\/.*/)
+        (value) => !value || !!value.match(/https:\/\/www.linkedin.com\/.*/)
       ),
       instagram: string().test(
         'isSocialMediaLink',
         messages.social_medias,
-        (value) => !value || !!value.match(/https:\/\/instagram.com\/.*/)
+        (value) => !value || !!value.match(/https:\/\/www.instagram.com\/.*/)
```
SARodrigues, line 92: These will be failing with the following: 

```
http://www.instagram.com/vizzuality/
http://instagram.com/vizzuality/
https://instagram.com/vizzuality/
https://instagram.com/
```

&nbsp;  

Since many people type their social media links form memory, I think it may be worth it to make the `s` in `https` and `www.` be optional. Eg: 

`https?:\/\/(www.)?instagram.com\/.*`

&nbsp; 

We can improve it a little bit further by assuming there'll always be a trailing slash after `instagram.com` plus something. Eg:

`https?:\/\/(www.)?instagram.com\/.+`

SARodrigues (approved):

frontend/schemas/investor.ts
SARodrigues: @barbara-chaves You probably read it before I edited the comment, but there's another change: 
`https:\/\/www.instagram.com\/.*` to `^https:\/\/www.instagram.com\/.*$`

SARodrigues (approved):

PR #231, opened 2022-05-31
feat: Added pagination to backoffice tables

martintomas:
This PR adds pagination to backoffice tables ;).

Screenshot 2022-05-31 at 18 09 27

## Testing instructions

Open PDs or Investors Administration and see pagination navigation there.

## Tracking

https://vizzuality.atlassian.net/browse/LET-503

### DIFF
--- a/backend/app/assets/stylesheets/application.tailwind.css
+++ b/backend/app/assets/stylesheets/application.tailwind.css
@@ -34,4 +34,24 @@
   .backoffice-table tr:last-child {
     @apply border-b-0;
   }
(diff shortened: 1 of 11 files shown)


tsubik:
We should stick to the designs :) 
image

martintomas:
> We should stick to the designs :) image

I though design was simplified and it is better to stay consistent with FE, but I don't mind to switch to ones from Figma ;)

tsubik (approved):
Works great :+1:

PR #232, opened 2022-05-31
fix: Geometry param is permitted even for nested attributes

martintomas:
GeoJson with Project geometry can be nested json -- instead of permitting every nested attribute of geojson separately, I permit whole geojson by default. 

I am adding security warning to ignore which should be fine, because there are couple of validations on Project model which ensure that GeoJSON has proper format.

## Testing instructions

Try to create or update Project with complex GeoJSON geometry (coordinates are defined inside feature)

## Tracking

BugFix ;)

### DIFF
--- a/backend/app/controllers/api/v1/accounts/projects_controller.rb
+++ b/backend/app/controllers/api/v1/accounts/projects_controller.rb
@@ -57,15 +57,13 @@ module API
             :progress_impact_tracking,
             :description,
             :relevant_links,
-            :geometry,
-            geometry: {},
             involved_project_developer_ids: [],
             target_groups: [],
             impact_areas: [],
             sdgs: [],
             instrument_types: [],
             project_images_attributes: %i[id file cover _destroy]
-          )
+          ).tap { |r| r[:geometry] = params[:geometry].permit! if params[:geometry].present? }
         end
 
         def fetch_project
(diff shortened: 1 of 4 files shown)


backend/app/controllers/api/v1/accounts/projects_controller.rb
```diff
@@ -57,15 +57,13 @@ def update_params
             :progress_impact_tracking,
             :description,
             :relevant_links,
-            :geometry,
```
tsubik, line 60: I thought I would just allow custom JSON, without having explicitly call `permit!`.

tsubik (approved):

backend/app/controllers/api/v1/accounts/projects_controller.rb
martintomas: How? :) I have not figured out way how to allow basically anything for `geometry` param including nested attributes...

martintomas: Ahh, sry, I though you had some idea how to do it different, because tbh, I don't like that `permit!` part at all, but all other options had problems with nested stuff 🤔

tsubik: No I don't, I thought it just should work as it was with jsonb column :D hehe

PR #235, opened 2022-06-01
LET-268

agnessa:
https://vizzuality.atlassian.net/browse/LET-268?atlOrigin=eyJpIjoiMTY1NmZkYmFmM2RkNDBjMjkxYWNhYWVmOTk0MWZjMDkiLCJwIjoiaiJ9

This is so that we have error reporting where we rescue exceptions. Also removed the test exception controller.


backend/app/jobs/translate_job.rb
```diff
@@ -4,7 +4,7 @@ class TranslateJob < ApplicationJob
   rescue_from(ActiveRecord::RecordNotFound) do |exception|
     Rails.logger.error(exception)
     Rails.logger.error(Rails.backtrace_cleaner.clean(exception.backtrace).join("\n"))
-    # TODO: handle exception tracking
+    Google::Cloud::ErrorReporting.report exception
```
martintomas, line 7: I remember that I have added some hooks for BackgroundJob Events to ApplicationJob and they also store exception to db. Maybe that reporting should be added there and all these `rescue_from ` and `handle_exception ` could be removed so we deal with exceptions at one place for all current and future jobs.

martintomas:
```
def store_exception_at_event
    yield
  rescue Exception => e # rubocop:disable Lint/RescueException
    upsert_event self, status: :crashed, message: e.backtrace
    raise e
  end
```

martintomas (approved):

backend/app/jobs/translate_job.rb
tsubik: I agree we can move it there.

tsubik:
@agnessa please also add this to `api/errors.rb` concern.

backend/app/jobs/translate_job.rb
agnessa: OK! Thanks!

agnessa: Oh but actually in that case we do a `raise` - that will automatically report

tsubik: Ohh, ok. Then, if we still need to have the clean error messages in the logger (not sure if there is any need of that now, if we have error management), we can move Rails.logger part.

tsubik: We can keep it, dunno :)

tsubik: Or actually, raising exception will get a good log as well, so we can just remove the whole thing, lol.

agnessa:
Added to the controller concern and removed from Translate job and service - it will be handled in the parent job.

PR #236, opened 2022-06-01
feat: Making backoffice tables sortable

martintomas:
This PR adds possibility to sort/order Investor and Project Developer page at Backoffice. All heavy lifting is done by `ransack` -- tbh, I have taken huge inspiration from Tota project. The columns which were using `count` are rewritten to use `counter_cache` instead.

Tests are quite simple, but I am trusting that ransack is doing its job, therefore It should be enough :).

I wasn't able to extract correct `arrow` icons which are visible next to every column name from Figma so I used simpler ones, but they work and can be replaced by correct ones at future.

Screenshot 2022-06-01 at 13 55 18

## Testing instructions

Open PDs or Investors table at Backoffice and try to order it by random column values.

## Tracking

https://vizzuality.atlassian.net/browse/LET-504

### DIFF
--- a/backend/Gemfile
+++ b/backend/Gemfile
@@ -27,6 +27,7 @@ gem "friendly_id", "~> 5.4.0"
 gem "devise"
 gem "devise-i18n"
 gem "traco"
+gem "ransack"
 
 # API
 gem "jsonapi-serializer"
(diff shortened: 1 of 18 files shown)


tsubik (approved):
:+1: great job!

PR #239, opened 2022-06-02
feat: Sends emails on Account approve or reject

martintomas:
Two new emails are send after account approval or account rejection. I was not sure about email content so I have improvised a lot, but everything is at templates and locales so super simple to change it at future.

I have added new variable called `FRONTEND_URL` so I can provide link to user at email -- tbh. because I am not sure about emails content, I am not sure if this is required at this point :).

I also had to update `gem "rspec-rails", "6.0.0.rc1"`, because current version does not work properly with Rails 7 --> `have_enqueued_mail ` is broken.

## Testing instructions

Reject or approve any user via Backoffice -- such user should receive appropriate email. Please test only for accounts that you own their user emails :).

## Tracking

https://vizzuality.atlassian.net/browse/LET-514


backend/Gemfile
```diff
@@ -11,7 +11,7 @@ gem "puma", "~> 5.6"
 gem "redis", "~> 4.0"
 gem "rswag"
 gem "active_storage_validations"
-gem "rspec-rails", "~> 5.0.0"
+gem "rspec-rails", "6.0.0.rc1"
```
tsubik, line 14: What is new in rspec 6 that we need rc1 here?

backend/config/locales/zu.yml
```diff
@@ -40,6 +40,15 @@ zu:
...
         no_investor: User is not investor!
+  user_mailer:
+    greetings: Dear %{full_name},
+    farewell_html: Best Regards,<br />Your HeCo Team
+    approved:
+      subject: Your account was approved!
+      content: We are happy to inform you that your account at HeCo platform was approved.
+    rejected:
+      subject: Your account was rejected!
+      content_html: We would like to inform you that your account at HeCo platform was rejected. One of our platform administrators should reach you shortly, but feel free to contact us at %{contact_url}.
```
tsubik, line 51: I would be probably more lazy in terms of making up the content of this email :P

backend/app/mailers/user_mailer.rb
```diff
@@ -0,0 +1,13 @@
+class UserMailer < ApplicationMailer
```
tsubik, line 1: I would probably create `AccountMailer`

tsubik: I don't know, what do you think?

tsubik: On the second thought, actually `UserMailer` is better, I'm not sure if only account owner should get that email or all users of the account? cc @agnessa

backend/Gemfile
martintomas: Tried to mention this at description, but basically `have_enqueued_mail ` was not working properly. Issue is here: https://github.com/rspec/rspec-rails/issues/2531

backend/config/locales/zu.yml
martintomas: :) yeah, not sure if I was not overthinking this, but I am open to put here whatever text you feel would be more appropriate

backend/Gemfile
tsubik: Ok, thanks.

tsubik: I never used that matcher. Have a bad memory to remember that many of them, :D I think I remember only handful.

backend/app/mailers/user_mailer.rb
tsubik: Yeah, I was thinking the same. `UserMailer` makes more sense for both cases, sending to single account owner and sending to all users.

tsubik (approved):

martintomas:
Please @agnessa , I will need your assistance here :). We need to either add FRONTEND_URL variable to stage or rewrite email of text that this variable is not needed <-- text of the emails is huge brainstorming and I hope a bit that somebody will polish them later on :D

agnessa (approved):

PR #242, opened 2022-06-02
Fix flaky backend tests - snapshots ignore included array

tsubik:
Using fork of snaphost gem, until I make a PR to main gem and it's merged.

## Testing instructions

I was no longer able to get any failing specs when running specs multiple times this way 
```
for i in `seq 20`; bundle exec rspec spec/requests; done
```

## Tracking

Link to the task(s), if any

### DIFF
--- a/backend/Gemfile
+++ b/backend/Gemfile
@@ -75,7 +75,7 @@ group :test do
   gem "cuprite"
   gem "rspec-collection_matchers"
   gem "rspec-retry"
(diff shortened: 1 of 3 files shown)


tsubik:
I'm not sure if I did it correctly as the author of the library is sorting ignored Arrays by arrays of elements hash values, maybe there is a better way, but it's ok for now.

PR #243, opened 2022-06-02
Fixing account users count in backoffice

tsubik:
There was a problem with seeds, and account owners did not have `account_id` applied. When I was adding user count + 1, I somehow thought we are not counting owners in `users` relationship :), because I had `0` and it was because of the bug in seeds. Just realized that when entering back office on staging when I saw that all accounts somehow have 2 users which is wrong.

## Testing instructions

What to test? How to do it?

## Tracking

Link to the task(s), if any

### DIFF
--- a/backend/app/views/backoffice/investors/index.html.erb
+++ b/backend/app/views/backoffice/investors/index.html.erb
@@ -28,7 +28,7 @@
       <tr>
         <td><%= investor.name %></td>
         <td><%= investor.account.owner.full_name %></td>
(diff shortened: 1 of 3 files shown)


martintomas (approved):
Good point, yeah, we actually assign newly created account back to owner, therefore `has_many` relation for `users` take into account even owner.

tsubik:
Yeah, and that is the way it should be done. I don't know why I thought account owner is not among users ;]

PR #244, opened 2022-06-03
Let 411 fe display and cluster the projects on the map search page

barbara-chaves:
This PR adds the geographical location of searched projects on the map, with markers and clusters.
Coauthored by @mbarrenechea 

### Acceptance criteria

There will be a way to geographically locate the projects in the search results (map)

The “central location” of the project (of the polygon defined) will be displayed in the map

The projects location will be displayed with a color code according to the project’s category

 * Sustainable agrosystems - yellow
 * Tourism and recreation - light blue
 * Forestry and agroforestry - orange
 * Non-timber forest production - dark blue
 * Human capital and inclusion - brown
 
The approved project should display a white check on Pin marker, but right now we don't have any
Screenshot 2022-06-03 at 11 07 57

It should conform to the [design](https://www.figma.com/file/KC49uMR9t1YARmFR0Sr6aN/HeCo---Work-in-progress?node-id=2%3A4726)

## Tracking

[LET-411](https://vizzuality.atlassian.net/browse/LET-411)

## Screenshots

https://user-images.githubusercontent.com/48164343/171824286-99abeabf-bb12-4202-b485-47ff31967dea.mov


frontend/containers/discover-map/pin-cluster/helper.ts
```diff
@@ -0,0 +1,55 @@
+import { values } from 'lodash-es';
+import Supercluster from 'supercluster';
+
+import { ProjectsMap } from 'types/project';
+
+export const PROJECT_CATEGORY_COLORS = {
```
SARodrigues, line 6: I don't think we should hardcode the colours here. We have the colours configured in `tailwind.config.js` and `/api/v1/enums` also returns them.

barbara-chaves: I changed the function to use the tailwind.config.
I also changed the method to get the cluster points so I don't have to use recursion

SARodrigues (approved):

PR #245, opened 2022-06-03
feat: Favourite Investor API endpoint

martintomas:
New API endpoint which allows to mark/unmark investors as favourite. Controller is basically same as ones used for marking Project or Project Developers as favourite.

I have moved all policy checks to cancancan for these controllers, therefore You are now required to have approved account otherwise you receive forbidden error.

## Testing instructions

New API endpoint for making Investors favourite is available via Rswag documentation, therefore everything can be tested via there. Please make sure to use approved account, otherwise you should receive forbidden error.

## Tracking

https://vizzuality.atlassian.net/browse/LET-526
https://vizzuality.atlassian.net/browse/LET-527
https://vizzuality.atlassian.net/browse/LET-554

### DIFF
--- a/backend/spec/requests/api/v1/favourite_investors_spec.rb
+++ b/backend/spec/requests/api/v1/favourite_investors_spec.rb
@@ -15,7 +15,7 @@ RSpec.describe "API V1 Favourite Investor", type: :request do
       let(:user) { create :user, account: create(:account, :approved) }
 
       it_behaves_like "with not authorized error", csrf: true
-      it_behaves_like "with forbidden error", csrf: true
+      it_behaves_like "with forbidden error", csrf: true, user: -> { create(:user, account: create(:account, :unapproved)) }
 
       response "200", :success do
         let("X-CSRF-TOKEN") { get_csrf_token }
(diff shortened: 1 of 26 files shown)


backend/spec/requests/api/v1/favourite_investors_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      security [csrf: [], cookie_auth: []]
+      parameter name: :investor_id, in: :path, type: :string
+      parameter name: :empty, in: :body, schema: {type: :object}, required: false
+
+      let(:investor) { create :investor }
+      let(:investor_id) { investor.id }
+      let(:user) { create :user, account: create(:account, :approved) }
+
+      it_behaves_like "with not authorized error", csrf: true
+      it_behaves_like "with forbidden error", csrf: true
```
tsubik, line 18: Maybe we need to refactor specs to make those behaviors more clear. Like `it_behaves_like "with forbidden error", user: -> { create(:user) }`. To know for what kind of user we are testing.

martintomas: Like it, will do it :)

backend/spec/support/shared_examples/api/with_forbidden_error.rb
```diff
@@ -1,15 +1,11 @@
...
   response "403", "Forbidden" do
     let("X-CSRF-TOKEN") { get_csrf_token } if csrf
 
     before do
-      if require_project_developer
-        sign_in create(:user_project_developer)
-      else
-        sign_in create(:user)
-      end
+      sign_in instance_exec(&user) if user.present?
```
tsubik, line 8: Nice :+1:

tsubik (approved):
:rocket:

PR #246, opened 2022-06-03
Feature/Backoffice project list

tsubik:
image

1. Category search is kinda workaround as we have translated names in Rails locale files, so I needed to use `ransacker` stuff and change names to numeric indexes and then sorting on that.
2. Sorting by `verified` slash `trusted` is right now done just like sorting by boolean column. I'm not sure if we want to spend more time and complicate, and does it like the category search is done? I'm lazy :P
3. Added a `status_tag` helper.
4. Seeds are updated to return more stuff, and with better properties.
5. Changed default pagination to 10 items as per design.
6. Enabled Turbo, but disabled it for devise as it's not supported. I have managed to enable it for devise in other project, but I have not copied that solution here as there could be some problems with it and resolving them could take more time.

I have not added total count as I think @martintomas is working on this right now, so don't want to duplicate work.

## Testing instructions

Run seeds and click through.

## Tracking

https://vizzuality.atlassian.net/browse/LET-530
https://vizzuality.atlassian.net/browse/LET-535
https://vizzuality.atlassian.net/browse/LET-536

### DIFF
--- a/backend/app/models/concerns/enum_model.rb
+++ b/backend/app/models/concerns/enum_model.rb
@@ -31,7 +31,7 @@ module EnumModel
     end
 
     def find(slug)
-      all.find { |o| o.slug == slug }
+      new(slug: slug)
     end
   end
 end
--- a/backend/app/models/concerns/enum_model.rb
+++ b/backend/app/models/concerns/enum_model.rb
@@ -33,5 +33,14 @@ module EnumModel
     def find(slug)
       new(slug: slug)
     end
+
+    def select_index_sql(column_name = name.underscore)
+      sql = ["CASE"]
+      all.sort_by(&:name).each_with_index do |enum, index|
+        sql << "WHEN #{column_name} ='#{enum.slug}' THEN #{index}"
+      end
+      sql << "END"
+      sql.join(" ")
+    end
   end
 end
--- a/backend/app/views/backoffice/projects/index.html.erb
+++ b/backend/app/views/backoffice/projects/index.html.erb
@@ -14,7 +14,7 @@
         <%= sort_link @q, :priority_landscape_name, t("backoffice.common.mosaic") %>
       </th>
       <th>
-        <%= t("backoffice.common.status") %>
+        <%= sort_link @q, :trusted, t("backoffice.common.status") %>
       </th>
     </tr>
   </thead>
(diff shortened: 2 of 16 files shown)


backend/spec/system/backoffice/projects_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+RSpec.describe "Backoffice: Projects", type: :system do
+  let(:admin) { create(:admin, email: "admin@example.com", password: "SuperSecret6", first_name: "Admin", last_name: "Example") }
+  let!(:project) {
+    project = create(
+      :project,
+      name: "Project ultra name",
+      project_developer: create(:project_developer, account: create(:account, name: "Ultra project developer name")),
+      trusted: true
+    )
+    project.update(priority_landscape: create(:location, location_type: "region", name: "Test mosaic"))
```
martintomas, line 12: haha, this is a bit hacky, I would prefer if you can define geometries for project and this location (they can be same so you are sure they intersect) ;)

backend/app/views/backoffice/projects/index.html.erb
```diff
@@ -0,0 +1,35 @@
+<table class="backoffice-table">
+  <thead>
+    <tr>
+      <th>
+        <%= sort_link @q, "name_#{I18n.locale}", t("backoffice.common.project_name") %>
```
martintomas, line 5: Just two minor points:
 - maybe I would prepare wrapper for `sort_link` called for example `localized_sort_link`, because this will be probably required at other places as well
 - if I remember correctly how translation works so when `name_pt` is for example missing, it fallbacks to `name_en`, so theoretically you can see at this column text from selected language plus fallback one. This sorting will not work in such cases. On other hand, I don't think this is use case which will happen a lot so I would not spend much time with it. Just something to keep in mind :)

backend/app/models/concerns/enum_model.rb
```diff
@@ -31,7 +31,16 @@ def find_many(slugs)
     end
 
     def find(slug)
-      all.find { |o| o.slug == slug }
+      new(slug: slug)
+    end
+
+    def select_index_sql(column_name = name.underscore)
```
martintomas, line 37: need to try this, because this is kind of advanced :D. At first sight `select_index_sql` is class method, but `name` is instance method so not sure how this works out...

martintomas: btw not sure, but this would be probably worthy test so I don't need to test it manually :D

martintomas (approved):
Looks good, just a few minor points, but they are mostly ideas which came to my mind :D

backend/app/views/backoffice/projects/index.html.erb
tsubik: Hmm, not sure how fallbacks work for translated column, if they actually do, but sorting by fallback we will need a `ransacker` helper I guess and complicate this more. I agree, I can add `localized_sort_link` helper.

backend/app/models/concerns/enum_model.rb
tsubik: Ok, I will add some test. `name` is a class.name.

PR #253, opened 2022-06-07
feat: Filtering by status at Backoffice

martintomas:
Adding possibility to filter Projects, PDs and Investors by Status. I have again taken huge inspiration from Tota project ;) and moved full_text filter inside `search_form_for` formular so searching and filters are part of same form.

Styles are really simplified, for example for dropdown, I have used simple selectbox which should be fine from tech point of view and can be replaced later on by something more fancy...

Screenshot 2022-06-07 at 12 31 55

## Testing instructions

Open Backoffice and try to filter by Status Projects, PDs and Investors one by one.

## Tracking

https://vizzuality.atlassian.net/browse/LET-505
https://vizzuality.atlassian.net/browse/LET-531
https://vizzuality.atlassian.net/browse/LET-532
https://vizzuality.atlassian.net/browse/LET-564

### DIFF
--- a/backend/app/services/api/filterer.rb
+++ b/backend/app/services/api/filterer.rb
@@ -38,7 +38,8 @@ module API
       associated_columns = Array.wrap(FULL_TEXT_EXTRA_TABLES[query.klass]).each_with_object({}) do |relation, res|
         res[relation] = localized_columns_for relation.to_s.classify.constantize
       end
-      self.query = query.where id: query.dynamic_search(localized_columns_for(query.klass), filters[:full_text], associated_columns).pluck(:id)
+      ids = query.klass.dynamic_search(localized_columns_for(query.klass), filters[:full_text], associated_columns).pluck(:id)
+      self.query = query.where id: ids
     end
 
     def apply_numerical_filter
--- a/backend/app/controllers/backoffice/investors_controller.rb
+++ b/backend/app/controllers/backoffice/investors_controller.rb
@@ -2,14 +2,8 @@ module Backoffice
   class InvestorsController < BaseController
     def index
       @q = Investor.ransack params[:q]
-      @investors = API::Filterer.new(@q.result, filter_params.to_h).call
+      @investors = API::Filterer.new(@q.result, {full_text: params.dig(:q, :filter_full_text)}).call
       @pagy_object, @investors = pagy @investors.includes(account: [:owner]), pagy_defaults
-    end
-
-    private
-
-    def filter_params
-      params.fetch(:filter, {}).permit :full_text
     end
   end
 end
--- a/backend/app/controllers/backoffice/project_developers_controller.rb
+++ b/backend/app/controllers/backoffice/project_developers_controller.rb
@@ -2,14 +2,8 @@ module Backoffice
   class ProjectDevelopersController < BaseController
     def index
       @q = ProjectDeveloper.ransack params[:q]
-      @project_developers = API::Filterer.new(@q.result, filter_params.to_h).call
+      @project_developers = API::Filterer.new(@q.result, {full_text: params.dig(:q, :filter_full_text)}).call
       @pagy_object, @project_developers = pagy @project_developers.includes(account: [:owner]), pagy_defaults
-    end
-
-    private
-
-    def filter_params
-      params.fetch(:filter, {}).permit :full_text
     end
   end
 end
--- a/backend/app/controllers/backoffice/projects_controller.rb
+++ b/backend/app/controllers/backoffice/projects_controller.rb
@@ -2,8 +2,8 @@ module Backoffice
   class ProjectsController < BaseController
     def index
       @q = Project.ransack params[:q]
-      @projects = @q.result.includes(:priority_landscape, project_developer: [:account])
-      @pagy_object, @projects = pagy @projects, pagy_defaults
+      @projects = API::Filterer.new(@q.result, {full_text: params.dig(:q, :filter_full_text)}).call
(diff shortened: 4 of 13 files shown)


backend/app/services/api/filterer.rb
```diff
@@ -38,7 +38,8 @@ def apply_full_text_filter
       associated_columns = Array.wrap(FULL_TEXT_EXTRA_TABLES[query.klass]).each_with_object({}) do |relation, res|
         res[relation] = localized_columns_for relation.to_s.classify.constantize
       end
-      self.query = query.where id: query.dynamic_search(localized_columns_for(query.klass), filters[:full_text], associated_columns).pluck(:id)
+      ids = query.klass.dynamic_search(localized_columns_for(query.klass), filters[:full_text], associated_columns).pluck(:id)
```
martintomas, line 41: ransack and pg_search have problem to work together so I have split query to two separate ones ;)

tsubik: Good to know.

martintomas:
> I see you added more filters @martintomas. Can you also add reset filter button? Just a link to entity index page, if any filter is applied.

Sure, I don't think it is at design, but don't see problem with adding it ;)

backend/app/views/backoffice/projects/_filters.html.erb
```diff
@@ -0,0 +1,15 @@
+<%= render "backoffice/shared/filters", q: q do |f| %>
+  <%= render partial: "backoffice/shared/full_text_filter", locals: { f: f } %>
+  <div class="w-30 inline-block mx-3">
+    <%= f.input :status_eq, as: :select, required: false, label: false, include_blank: I18n.t("backoffice.common.status"),
+                collection: Project.statuses.map { |k, v| [I18n.t("activerecord.attributes.project.statuses.#{k}"), v] } %>
```
tsubik, line 5: I would add enum model for `ProjectStatus`, and keep the translations there. Maybe frontend will need this as well.

backend/app/views/backoffice/projects/_filters.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  </div>
+  <div class="w-30 inline-block">
+    <%= f.input :category_eq, as: :select, required: false, label: false, include_blank: I18n.t("backoffice.common.category"),
+                collection: Category.all.sort_by(&:name).map { |c| [c.to_s, c.slug] } %>
```
tsubik, line 9: Is the `map` even necessary? I think at least for simple form collection it is not for EnumModels.

backend/app/services/api/filterer.rb
agnessa: It might be possible to use subquery instead of plucking ids? Just in case somebody types, like, "e", and that matches the whole table worth of ids :), maybe it could work almost exactly like you have it, but using select instead of pluck:
fts_subquery = query.klass.dynamic_search(localized_columns_for(query.klass), filters[:full_text], associated_columns).select(:id)

self.query = query.where id: fts_subquery

tsubik (approved):
:+1:

backend/app/models/project.rb
```diff
@@ -28,7 +28,7 @@ class Project < ApplicationRecord
     :progress_impact_tracking,
     :relevant_links
 
-  enum status: {draft: 0, published: 1, closed: 2}, _default: :draft
+  enum status: ProjectStatus::TYPES_WITH_CODE, _default: :draft
```
agnessa, line 31: in my local db project status is still a string field, and so the status filter doesn't work - is there a migration missing or did I mess sth up?

agnessa:
I have trouble with testing filtering by mosaic, because I messed up my db and need to reimport the locations table, I'll try to figure that out. In the meantime:

* I found that filtering projects by status doesn't work for me, because the columns a text one in my db rather than int and  I'm not sure if this is sth I messed up locally (I do that a lot sadly) or there's a migration missing to convert it?
* It appears to me as an inconsistency with the status filters that while on projects you have the ability to select a status, apply and then deselect it by choosing "Status", the same does not happen for project developers / investors (include_blank missing - you can ofc reset all filters).
* Also not sure about the "Unverified" status in the projects status dropdown, shouldn't it read "Draft"? so that we don't confuse this status with the trusted / verified flag
* I added a small suggestion regarding using subquery instead of pluck, unimportant if it creates additional problems

backend/app/services/api/filterer.rb
martintomas: Done, thanks @agnessa ;). I have tried to break query to two by pluck, because that sql is huge and quite complex, but using select has its own advantages, so why not to go for it.

martintomas:
I have tried to address all your issues @agnessa 

> * I have trouble with testing filtering by mosaic, because I messed up my db and need to reimport the locations table, I'll try to figure that out. In the meantime:

It is integer at my db, so I guess problem is with your database.

> * I found that filtering projects by status doesn't work for me, because the columns a text one in my db rather than int and  I'm not sure if this is sth I messed up locally (I do that a lot sadly) or there's a migration missing to convert it?

Status filter now uses `trusted` column for filtering and hopefully should work for you now.

> * It appears to me as an inconsistency with the status filters that while on projects you have the ability to select a status, apply and then deselect it by choosing "Status", the same does not happen for project developers / investors (include_blank missing - you can ofc reset all filters).

Added missing `include_blank` params so it should be consistent now.

> * Also not sure about the "Unverified" status in the projects status dropdown, shouldn't it read "Draft"? so that we don't confuse this status with the trusted / verified flag

Yup, switched to `trusted` column, text is now verified/unverified.

> * I added a small suggestion regarding using subquery instead of pluck, unimportant if it creates additional problems

Good point, I am using select and subquery now.

PR #254, opened 2022-06-08
Feature/backoffice project developer edit

tsubik:
Edit project developer form.
pd-form

I also had to use rails tailwind gem, to make it work with asset pipeline (background images).

## Testing instructions

There is test suite, but could be tested manually.

## Tracking

https://vizzuality.atlassian.net/browse/LET-541
https://vizzuality.atlassian.net/browse/LET-538


backend/app/validators/array_inclusion_validator.rb
```diff
@@ -4,6 +4,8 @@ class ArrayInclusionValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
     return if value.nil?
 
+    value.reject!(&:blank?)
```
tsubik, line 7: Probably this is not a good place to do this, but didn't want to pollute controllers with eliminating all of those blank values that we get by default, in case someone wants to unselect all categories for example. Rails is sending empty value in array unless include_blank is set to `false`. It is sending for example ["", "cat1", "cat2"], just in case if someone unselect cat1, cat2 to send this parameter, otherwise categories will be unchanged.

backend/app/views/backoffice/project_developers/_section_profile.html.erb
```diff
@@ -0,0 +1,14 @@
+<div>
+    <label class="flex items-center gap-3">
+      <%= t("backoffice.common.view_content_in") %>
+      <%= select_tag :content_lang, options_from_collection_for_select(Language.all, "id", "name", content_language),
+         class: "select",
+         onchange: "
```
tsubik, line 6: I was lazy to add stimulus here ;]

backend/app/controllers/backoffice/project_developers_controller.rb
```diff
@@ -1,9 +1,84 @@
...
+    def content_language
+      Language::TYPES.map(&:to_s).include?(params[:content_lang]&.downcase) ? params[:content_lang].downcase : @project_developer.language
+    end
+
+    def section_partial
+      "section_#{section_key}"
+    end
+
+    def section_key
+      %w[language profile status].include?(params[:section]&.downcase) ? params[:section].downcase : "profile"
```
martintomas, line 81: hmm, again, would try to clear this controller a bit so you just need to define absolute minimum here:
```
sections :i[language profile status], default: :profile
```
rest of the code can go to parent or concern

backend/app/models/account.rb
```diff
@@ -25,7 +26,7 @@ class Account < ApplicationRecord
   validates :twitter, url: true
   validates :facebook, url: true
   validates :instagram, url: true
-  validates :language, inclusion: {in: Language::TYPES}
+  validates :language, inclusion: {in: Language::TYPES, allow_blank: true}, presence: true
```
martintomas, line 29: just question, what is reason behind this? I see that you have not touched validation tests so it behaves same way as before so not sure why this change was required

backend/app/validators/array_inclusion_validator.rb
martintomas: Yeah, if put here, it is gonna be a bit of magic at background, a bit difficult to find for our future selfs :).

backend/app/views/backoffice/project_developers/_form.html.erb
```diff
@@ -0,0 +1,34 @@
+<%= simple_form_for [:backoffice, project_developer] do |f| %>
+  <%= f.error_notification %>
+  <%= f.error_notification message: f.object.errors[:base].to_sentence if f.object.errors[:base].present? %>
+
+  <h2 class="mb-4 mt-6 text-gray-600 font-semibold text-xl">
+    <%= t(".general") %>
+  </h2>
+
+  <%= f.simple_fields_for :account do |ff|  %>
+    <%= ff.input :picture %>
```
martintomas, line 10: :) no direct upload?

backend/app/views/backoffice/base/_account_fields.html.erb
```diff
@@ -0,0 +1,28 @@
+<%= f.simple_fields_for :account do |ff|  %>
```
martintomas, line 1: no sure if I am convinced to have this at `base` folder when this is gonna be valid just for PD and Investor. What about Account folder?

backend/app/views/backoffice/project_developers/_section_profile.html.erb
martintomas: feels a bit like something which will be part of every translatable form. Maybe I would move this to `base` folder and added it as partial to `form`.

backend/spec/system/backoffice/project_developers_spec.rb
```diff
@@ -74,11 +91,117 @@
...
+          uncheck t("enums.category.forestry-and-agroforestry.name")
+          check t("enums.impact.community.name")
+          uncheck t("enums.impact.water.name")
+          uncheck t("enums.impact.climate.name")
+          check t("enums.mosaic.amazon-heart.name")
+          click_on t("backoffice.common.save")
+
+          expect(page).to have_text(t("backoffice.messages.success_update", model: t("backoffice.common.project_developer")))
+          approved_pd.reload
+          expect(approved_pd.account.picture.filename.to_s).to eq("picture_2.jpg")
```
martintomas, line 160: This should be probably replaced at future by some page detail for updated PD -- which is basically true even for other tests

martintomas (approved):
Need to try it manually, but code looks quite good, just some minor comments/ideas which can be ignored :). Thanks for tweaking some of my code and styles 👍

backend/app/validators/array_inclusion_validator.rb
tsubik: I think it can stay here as this is a custom validator, we can treat it as a kind of  automatic `before_validation` hook, as this is one of the places where that kind of logic usually exists.

backend/app/views/backoffice/project_developers/_section_profile.html.erb
tsubik: True, I move around files when adding a next form for investor.

backend/app/controllers/backoffice/project_developers_controller.rb
tsubik: I like the idea.

backend/spec/system/backoffice/project_developers_spec.rb
tsubik: I don't understand :)

martintomas: Ha :), sry. Preferred use of expect at system tests should be check that some page contains some expected values. I believe that it is more appropriate after PD gets updated to go to page with detail of PD and check that values are correct there instead of checking ActiveRecord itself. Ofc we dont have such page right now and I dont think we need to be super strict with this so just my opinion on this topic ;)

tsubik: Yeah. I was thinking if to check the inputs fields, after the page is reloaded, but thought this will be faster. We won't have details page. (I saw that message sent by email went to main thread)

tsubik: Maybe it is worth to refactor this to check those inputs and checkboxes if those are still correctly selected.

backend/app/views/backoffice/project_developers/_form.html.erb
```diff
@@ -0,0 +1,34 @@
+<%= simple_form_for [:backoffice, project_developer] do |f| %>
+  <%= f.error_notification %>
+  <%= f.error_notification message: f.object.errors[:base].to_sentence if f.object.errors[:base].present? %>
```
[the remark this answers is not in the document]
tsubik: I'm not even sure if we need default message, but I will leave it.

martintomas:
Tested manually and works for me ;). Just note about ticket https://vizzuality.atlassian.net/browse/LET-538 which specifies that `creating the form in a readonly way`, which I don't think current implementation supports. This is related to previous, but there should be some way how to see existing picture of current PD

martintomas:
Haha :), a few seconds late, np, we can address this at future ;)

PR #259, opened 2022-06-09
feat: Total count of Projects/PDs/Investors at backoffice

martintomas:
I am adding partial with Total number of records which is used by Project/PD/Investor pages.

Screenshot 2022-06-09 at 13 13 13

## Testing instructions

Open backoffice and check that number of records correspond to Total number displayed above every table.

## Tracking

https://vizzuality.atlassian.net/browse/LET-537
https://vizzuality.atlassian.net/browse/LET-556
https://vizzuality.atlassian.net/browse/LET-557

### DIFF
--- a/backend/app/views/backoffice/investors/index.html.erb
+++ b/backend/app/views/backoffice/investors/index.html.erb
@@ -1,4 +1,7 @@
-<%= render partial: "filters", locals: {q: @q} %>
+<div class="flex">
+  <%= render partial: "filters", locals: {q: @q} %>
+  <%= render partial: "backoffice/base/total_records" %>
+</div>
 
 <table class="backoffice-table">
   <thead>
--- a/backend/config/locales/zu.yml
+++ b/backend/config/locales/zu.yml
@@ -22,8 +22,9 @@ zu:
       verified: Verified
       unverified: Unverified
       search_placeholder: Search
-      apply: Apply filters
-      reset: Reset filters
+      apply: Apply
+      reset: Reset
+      total: Total
       navigation:
         prev: "←"
         next: "→"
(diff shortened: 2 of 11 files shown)


backend/config/locales/zu.yml
```diff
@@ -22,8 +22,9 @@ zu:
       verified: Verified
       unverified: Unverified
       search_placeholder: Search
-      apply: Apply filters
-      reset: Reset filters
+      apply: Apply
```
martintomas, line 25: I have slightly shorted text, because there is not much space at top of every table and I believe it is quite clear what Apply or Reset means.

backend/app/views/backoffice/investors/index.html.erb
```diff
@@ -1,4 +1,7 @@
-<%= render partial: "filters", locals: {q: @q} %>
+<div class="flex">
+  <%= render partial: "filters", locals: {q: @q} %>
+  <%= render partial: "backoffice/base/total_records" %>
```
tsubik, line 3: You can just write `total_records`

backend/config/locales/zu.yml
tsubik: Filters will be anyway refactored to match designs later. I think in internal copy, I saw them in popover.

tsubik (approved):

backend/config/locales/zu.yml
martintomas: Noooo :D, I have some wrong internal copy, because design from my figma file more or less match what we have now (ofc it is not pixel perfect :))

tsubik: I checked the one that is on slack channel, but that is ok we will iterate.

PR #267, opened 2022-06-10
feat: Allow to export data as csv at backoffice

martintomas:
I am adding ability to export data for Project, PD and Investors as CSV. All functionality for csv files is done via appropriate exporter services (even though they are quite simple and maybe could be part of controller). I have used logic similar to ActiveAdmin for defining csv output.

Data which are exported take into account active filters ;).
 
Screenshot 2022-06-10 at 13 24 13

## Testing instructions

Open backoffice and try to export data as csv for Projects, PDs and Investors.

## Tracking

https://vizzuality.atlassian.net/browse/LET-569
https://vizzuality.atlassian.net/browse/LET-570

### DIFF
--- a/backend/app/services/backoffice/csv/base_exporter.rb
+++ b/backend/app/services/backoffice/csv/base_exporter.rb
@@ -1,7 +1,7 @@
 require "csv"
 
 module Backoffice
-  module Csv
+  module CSV
     class BaseExporter
       attr_accessor :query, :columns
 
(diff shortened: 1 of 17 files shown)


backend/app/services/backoffice/csv/base_exporter.rb
```diff
@@ -0,0 +1,35 @@
...
+        @query = query
+      end
+
+      def call
+        raise NotImplementedError
+      end
+
+      private
+
+      def generate_csv
```
agnessa, line 18: I like this approach , it's very flexible and allows us to add columns to the csv without affecting the backoffice view.

agnessa (approved):
Works, great!

tsubik (changes requested):
Code looks great! Maybe just one tiny change would be beneficial. Please add "CSV" acronym to `inflections.rb`. I always find weird `Csv` name.

tsubik (approved):
:rocket:

PR #276, opened 2022-06-13
Feature/backoffice investor edit

tsubik:
Investor edit page

## Testing instructions

What to test? How to do it?

## Tracking

https://vizzuality.atlassian.net/browse/LET-539
https://vizzuality.atlassian.net/browse/LET-540


martintomas (approved):
Looks good ;). Just same note as about PD form -- at https://vizzuality.atlassian.net/browse/LET-539 is note about forms at readonly way and I don't think this is supported. It would be also nice to have some possibility to see current `picture` <-- you can edit it, but not see it

tsubik:
As I know we don't need readonly view (it is just complicating things for no value). I think just description of ticket is just to make it look a little easier, and do it in two steps instead of one (just edit). Is that right @agnessa ?

agnessa:
Yes, the idea about "readonly" was just about splitting the task into smaller chunks for "edit" and "update", with edit being the more important that it is the only way to see all attributes. Thanks!

PR #277, opened 2022-06-14
feat: Invitation API endpoints

martintomas:
I am adding three new API endpoints which should handle:

**Generation of Invitation**
 - available for new or existing users (just provide email address) --> User should receive email
 - returns `401` if user is not logged in
 - returns `403` if user is not owner of approved account 
 - returns `409` if user already have account
 - returns `422` if provided emails is not email

**Accepting Invitation**
 - provide invitation token which should be part of link inside email. For already registered users, you can immediately accept invitation, for new users, additional info needs to be provided such as first_name, last_name, password, etc.
 - returns `404` if invitation token is invalid
 - returns `409` if user already have account
 - returns `422` if user needs to provide extra info or such info is invalid for invitation to be accepted

**List of users of account**
 - if you are user of same account, you can see other users
 - if you are owner of account, you can see other users and also invited ones (before they accepted their invitation)
 - added new attribute called `invitation` to User serialized which describes if user was invited and its status

## Testing instructions

All new API endpoints should be available via Rswag documentation and testing should be quite straightforward --> first generate invitation for new or existing user, then accept this invitation. For new user, you also need to provide extra data.

## Tracking

https://vizzuality.atlassian.net/browse/LET-453
https://vizzuality.atlassian.net/browse/LET-454


backend/app/controllers/api/v1/invitations_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+        authorize! :invite, @user
+
+        @user.invite! current_user
+        head :ok
+      end
+
+      def update
+        @user.with_lock do
+          @user.assign_attributes params.permit(:password, :first_name, :last_name, :ui_language)
+          @user.assign_attributes @user.invited_by&.attributes&.slice("role", "account_id") if @user.invited_by.present?
```
tsubik, line 19: I would probably raise an error if a user that invites another user to the platform does not exist.

tsubik: To create this view 
image

martintomas: Yeah, like this should not happen to be honest, probably put these `&` there during testing, will get rid of it ;)

backend/app/models/ability.rb
```diff
@@ -36,6 +37,10 @@ def user_rights
     can %i[show], OpenCall, investor: {account_id: user.account_id}
   end
 
+  def owner_rights
+    can %i[invite], User, account_id: nil
```
tsubik, line 41: must be also approved user to invite others.

tsubik: or to be more precise, owner of approved account :)

backend/app/controllers/api/v1/invitations_controller.rb
martintomas: hmm, good point, will move it to create action

backend/app/models/ability.rb
martintomas: haha, yeah, forgot to check this, nice catch :)

backend/app/controllers/api/v1/invitations_controller.rb
tsubik: It will be a little different than in the flow (we can update that file later), but I think it makes more sense from the future API `/account/users` API perspective.

backend/app/controllers/api/v1/invitations_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+          @user.save!
+          @user.accept_invitation!
```
tsubik, line 21: Meaning the user can log in already? Or maybe we should auto-login that user. I'm not sure. In the flow we have another step to send confirmation email, but first we disabled email confirmations for now, and second not sure if that is needed. I thought it would be more secure, for example if the attacker somehow obtains the invitation token, it still won't be possible to log in into that new account without having access to user's email account.

backend/app/controllers/api/v1/invitations_controller.rb
tsubik: Ok, then let's leave it if it will be fine.

backend/app/controllers/api/v1/invitations_controller.rb
tsubik: That is something to discuss with frontend.

martintomas: Agree, I would prefer no to login user (will double check this and turn it off if it is on right now), because there is gonna be 2FA at future

agnessa (approved):
Great, I love the use of different codes here :)

backend/app/models/user.rb
```diff
@@ -9,8 +9,9 @@ class User < ApplicationRecord
   has_many :project_developers, through: :favourite_project_developers
   has_many :favourite_investors, dependent: :destroy
   has_many :investors, through: :favourite_investors
+  has_many :owner_accounts, class_name: "Account", foreign_key: "owner_id", dependent: :destroy
```
tsubik, line 12: Technically it's possible, but business logic should be `has_one` I think. I would also name it `owned_accounts` or `owned_account` if `has_one`.

martintomas:
I will rewrite this and probably split to separate PRs

PR #278, opened 2022-06-15
Feature/backoffice delete account

tsubik:
Delete investor and project_developer accounts.

image

## Testing instructions

Backoffice UI

## Tracking

https://vizzuality.atlassian.net/browse/LET-542
https://vizzuality.atlassian.net/browse/LET-543


martintomas (approved):
👍

agnessa:
Looks good, I just thought (a bit too late) that maybe we should reset account users' role to light, what do you think? That was not specified, so I'd add a separate task if you agree.

PR #289, opened 2022-06-20
feat: Update of Project via Backoffice

martintomas:
Adding formular for managing Project to backoffice. There is one Todo and discovered couple bugs, but I would address them as separate PR so this one is not too huge.

Screenshot 2022-06-20 at 16 48 24

There is couple changes when compared to designs:
 - `partner` is multiselect, because of `involved_project_developers` is `has_many` relation
 - `category` is changed to selectbox, because project can have only one
 - `ticket_size` is changed to selectbox, because project can have only one

TODO which will be addresses as separate ticket:
 - I need to find way how to transform shapefile which can be uploaded via backoffice to GeoJson which is provided be FE
 - I am personally not big fan of way how project images are handled now (you always need to change whole set), but it corresponds to current proposal --> would probably require something more fancy at future
 
Discovered also two bugs which are related to all forms and localized fields (we will probably need additional tickets for this):
 - localized fields does not show if field is required or not
 - localized fields does not show erorrs

## Testing instructions

Backoffice ;)

## Tracking

https://vizzuality.atlassian.net/browse/LET-598
https://vizzuality.atlassian.net/browse/LET-599

### DIFF
--- a/backend/app/controllers/backoffice/projects_controller.rb
+++ b/backend/app/controllers/backoffice/projects_controller.rb
@@ -1,5 +1,13 @@
 module Backoffice
   class ProjectsController < BaseController
+    include Sections
+    include ContentLanguage
+
+    before_action :fetch_project, only: [:edit, :update]
+    before_action :set_breadcrumbs, only: [:edit, :update]
+    before_action :set_sections, only: [:edit, :update]
+    before_action :set_content_language_default, only: [:edit, :update]
+
     def index
       @q = Project.ransack params[:q]
       @projects = API::Filterer.new(@q.result, {full_text: params.dig(:q, :filter_full_text)}).call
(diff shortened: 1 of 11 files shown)


backend/app/controllers/backoffice/projects_controller.rb
```diff
@@ -16,5 +24,99 @@ def index
...
       end
     end
+
+    def edit
+    end
+
+    def update
+      if @project.update(update_params)
+        redirect_back(
+          fallback_location: edit_backoffice_project_path(@project.friendly_id),
```
tsubik, line 34: Hmm, I see you are using friendly_id for back-office links. I thought differently and was using always UUID at least in back-office as this will be unchangeable, and we don't have to worry about SEO. We should decide to just use one way.

martintomas: No problem, can switch to uuid, I have seen that API is using friendly_id so used same approach, but no big deal to change it here

tsubik:
> I am personally not big fan of way how project images are handled now (you always need to change whole set), but it corresponds to current proposal --> would probably require something more fancy at future
I agree. We should add a dynamic list (with the preview of already uploaded images). We can add a separate story for that.

tsubik (approved):

PR #290, opened 2022-06-20
Tweak Dashboard AccountInformation

SARodrigues:
## Description

This PR adds a few enhancements to the AccountInformation view on the dashboard:  
- Approximates the layout a little bit better to the design  
- Adds loading spinners when information isn't loading  
- Adds the contact admin message   
- When the user clicks "Edit Profile" and goes through the form, the user will be taken back to this particular Dashboard page. 
- Removed a few not in use helpers

## Testing instructions

N/A

## Tracking

N/A

## Screenshot  
Screenshot 2022-06-21 at 11 46 39


frontend/pages/project-developers/edit.tsx
```diff
@@ -30,10 +32,15 @@ export async function getStaticProps(ctx) {
 type ProjectDeveloperProps = InferGetStaticPropsType<typeof getStaticProps>;
 
 const ProjectDeveloper: PageComponent<ProjectDeveloperProps, FormPageLayoutProps> = () => {
+  const router = useRouter();
   const { formatMessage } = useIntl();
   const updateProjectDeveloper = useUpdateProjectDeveloper();
   const { projectDeveloper } = useCurrentProjectDeveloper();
 
+  const handleOnComplete = () => {
+    router.push((router.query?.returnPath as string) || Paths.Dashboard);
```
SARodrigues, line 41: @barbara-chaves I noticed that both when creating or editing the project developer, the user would be sent to the Dashboard entry point instead of the previous page. 

I've modified the `Button` component so it'll accept the `as` decorator from [next/link](https://nextjs.org/docs/api-reference/next/link), so we can add a `returnTo` path to the query (while keeping it hidden from URL bar). See [here](https://github.com/Vizzuality/heco-invest/pull/290/files#diff-f844f0f57fd222d935f2ec23dc1a8c9d95d1cd41c426113dedfc0b699e02266bR95-R98) with the button in use. 

This way we can redirect the user to the previous page 

I haven't done this in the `ProjectDeveloperForm` component just because I don't think the component itself should have that responsibility (plus it potentially minimises future issues with potential re-renders).

SARodrigues:
Rebased on `develop` to sort translation conflicts.

PR #291, opened 2022-06-20
Send a locale param in API requests

SARodrigues:
With this PR, API requests will include a `locale` param, set to the user's current locale. 

## Testing instructions

In the developer tools, verify that requests have a `locale=` appended. This is most important for the discover pages search, which is what prompted this task. 

## Tracking

[LET-425](https://vizzuality.atlassian.net/browse/LET-425)

### DIFF
--- a/frontend/services/api.ts
+++ b/frontend/services/api.ts
@@ -1,7 +1,11 @@
 import axios from 'axios';
 import Jsona from 'jsona';
 
+import { getCookie } from 'helpers/cookies';
+
 import { ErrorResponse } from './types';
+
+const { locales } = require('locales.config.json');
 
 const dataFormatter = new Jsona();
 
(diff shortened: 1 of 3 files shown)


frontend/services/api.ts
```diff
@@ -19,6 +23,16 @@ const API = axios.create({
   xsrfHeaderName: 'X-CSRF-TOKEN',
 });
 
+const onRequest = (config) => {
+  return {
+    ...config,
+    params: {
+      ...config?.params,
+      locale: getCookie('NEXT_LOCALE') || locales.find((locale) => locale.default).locale,
```
SARodrigues, line 31: We can't really use `useRouter` here, so we're reading the cookie that is set when the user switches languages. If no cookie has been set (for whatever reason), then we default to the default locale set in the config.

barbara-chaves: It is working for client and server sides, so it's perfect.

barbara-chaves (approved):

PR #299, opened 2022-06-21
fix: backoffice localized fields show correct validation errors

martintomas:
Fixing bugs with localized attributes at backoffice. These attributes now properly show:
 - if it is required
 - validation error if such error was raised during update

## Testing instructions

Backoffice...

## Tracking

BugFix ;)

### DIFF
--- a/backend/app/controllers/backoffice/investors_controller.rb
+++ b/backend/app/controllers/backoffice/investors_controller.rb
@@ -29,7 +29,7 @@ module Backoffice
     end
 
     def update
-      if @investor.update(update_params)
+      if I18n.with_locale(content_language) { @investor.update(update_params) }
         redirect_back(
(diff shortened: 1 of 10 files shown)


martintomas:
> Ahh, yes I remember now, why I did not use `I18n.with_locale` when fetching object and just showing inputs for normal attributes. Because of default fallback, is that right?

Yes, that is why I still need to use helper at views

martintomas:
> > Ahh, yes I remember now, why I did not use `I18n.with_locale` when fetching object and just showing inputs for normal attributes. Because of default fallback, is that right?
> 
> Yes, that is why I still need to use helper at views

Other problem is that content language uses language from db as default one so you cannot actually fetch object for some language when you need its language to decide its content language :). But workaround with that new helper method should work ;)

PR #302, opened 2022-06-22
Ensure all requests are localised on the server and client

clementprdhomme:
This PR makes sure that:
- All the requests initiated on the server are localised
- All the requests initiated on the server are refetched (and localised) on the client when the locale changes

## Testing instructions

- Choose a non default locale such as Portuguese
- Open a SSR-ed page such as a project's page and check that the correct locale was sent to the API (you can log it)
- Change the locale on that same page and check that all the queries are refetched with the new locale

## Tracking

- [LET-619](https://vizzuality.atlassian.net/browse/LET-619)
- [LET-620](https://vizzuality.atlassian.net/browse/LET-620)
- [LET-621](https://vizzuality.atlassian.net/browse/LET-621)


SARodrigues (approved):
That was quite a lot of work! Thank you.

clementprdhomme:
@SARodrigues I did some changes after I tested the project `/project/pd-organization-test-1-magic-amazonia-project-in-english-test-1` (which has a translated name).

The important bits of the PR are:
- The `getServerSideProps` and `getStaticProps` functions are all wrapped inside a `withLocalizedRequests` HOC. Internally, it retrieves the locale and sets it as a default parameter on all the axios requests initiated on the server.
- The calls to `useQuery` have been replaced by `useLocalizedQuery` which wraps `useQuery` internally, adding the locale to the list of dependencies. That way, whenever the locale changes, the queries are refetched. Invalidating the queries (with `queryClient.invalidateQueries()`) was not enough because I suspect react-query didn't know about the dependency on the locale (the cache wouldn't be invalidated).
- The interceptor you added to axios is still used on the client. It's the one injecting the `locale` parameter to all the requests coming from `useQuery`.

barbara-chaves (approved):

SARodrigues (approved):

SARodrigues:
Quick rebase to resolve a couple conflicts before merging.

PR #303, opened 2022-06-22
Front/improvement/dashboard projects

SARodrigues:
This PR adds a couple enhancements to the dashboard/projects section (but does not complete it): 

- It makes it so that when the user clicks on the "Edit" button on a project, when edition is fixed the user is returned to the same dashboard page. 
  _Virtually the same implementation as in #290_  
- Adds an option to disable table sorting entirely  
  _In case that doesn't get implemented in the API right away_

## Tracking

Continues [LET-551](https://vizzuality.atlassian.net/browse/LET-551) but does not complete it.

### DIFF
--- a/frontend/pages/dashboard/projects.tsx
+++ b/frontend/pages/dashboard/projects.tsx
@@ -81,7 +81,10 @@ export const ProjectsPage: PageComponent<ProjectsPageProps, DashboardLayoutProps
   const handleRowMenuItemClick = ({ key, slug }: { key: string; slug: string }) => {
     switch (key) {
       case 'edit':
-        router.push(`${Paths.Project}/${slug}/edit`);
+        router.push(
+          `${Paths.Project}/${slug}/edit?returnPath=${encodeURIComponent(router.asPath)}`,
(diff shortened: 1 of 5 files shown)


frontend/pages/dashboard/projects.tsx
```diff
@@ -81,7 +81,10 @@ export const ProjectsPage: PageComponent<ProjectsPageProps, DashboardLayoutProps
   const handleRowMenuItemClick = ({ key, slug }: { key: string; slug: string }) => {
     switch (key) {
       case 'edit':
-        router.push(`${Paths.Project}/${slug}/edit`);
+        router.push(
+          `${Paths.Project}/${slug}/edit?returnPath=${encodeURIComponent(router.asPath)}`,
```
SARodrigues, line 85: The URLs in this particular view can be something like: 

`/dashboard/projects?page=1&search=&sorting=category+desc`

While passing that as a `returnPath` param, everything after the first ampersand would be stripped. By encoding this we preserve the full return path.

barbara-chaves (approved):

anamontiaga (approved):

PR #305, opened 2022-06-22
Feature/backoffice shapefile

tsubik:
Moved shapefile upload to backoffice. (using frontend code, changed a little bit to allow ZIP upload)

image

## Testing instructions

I was creating shapefiles using this tool: https://geojson.io

## Tracking

https://vizzuality.atlassian.net/browse/LET-630


backend/config/locales/zu.yml
```diff
@@ -137,13 +139,20 @@ zu:
       project_developers_in_project: Project developers in the project (%{count})
       index:
         priority_landscape: Priority Landscape
-      edit:
+      form:
```
tsubik, line 142: Hmm, I'm not sure what will happen with already translated strings in transifex.

tsubik: But here was a bug as looks like it must be either partial or view name, to use lazy loading i18n feature.

martintomas: Or we could not use lazy loading for this :). I think it will be need to be translated again...

PR #307, opened 2022-06-22
Remove a hard-coded account name in the front-end code

clementprdhomme:
This PR only removes a hard-coded account name in the user invitation modal.

If you think the account name is important, then I can modify the PR to use the `useAccount` hook and dynamically insert it.

## Testing instructions

Look at the code 😉.

## Tracking

Not tracked.


anamontiaga:
Thanks @clementprdhomme! I think we can advanced without name atm.
Merging it ;)

PR #308, opened 2022-06-22
Add uptime checks

tiagojsag:
the user purge scheduler should be merged first, and then rebase

### DIFF
--- a/infrastructure/base/modules/env/main.tf
+++ b/infrastructure/base/modules/env/main.tf
@@ -391,6 +391,21 @@ module "purge_users_cron" {
   service_account_email = module.jobs_cloudrun.service_account_email
 }
 
+module "frontend_uptime_check" {
+  source = "../uptime-check"
+  name   = "${var.project_name} Frontend"
+  host   = var.domain
+  email  = var.uptime_alert_email
+}
+
+module "api_uptime_check" {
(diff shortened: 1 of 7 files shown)


agnessa:
did you say rebase? 😈 😈 😈

tiagojsag:
rebase is the true sign of git proficiency. I already did it, ready for review

infrastructure/base/modules/env/main.tf
```diff
@@ -391,6 +391,21 @@ module "purge_users_cron" {
...
+  name   = "${var.project_name} Frontend"
+  host   = var.domain
+  email  = var.uptime_alert_email
+}
+
+module "api_uptime_check" {
+  source = "../uptime-check"
+  name   = "${var.project_name} API"
+  host   = var.domain
+  path   = "/backend/health_check"
```
agnessa, line 405: I understand we check BE availability by using the health endpoint, but I don't understand how we check FE, i.e. there's no path given in the FE uptime check?

tiagojsag: correct, there is no path, we check if the homepage is up

agnessa: 🧙

agnessa (approved):

PR #274, opened 2022-06-13
Map uses custom Mapbox' basemap

SARodrigues:
Copy of #217 for debugging. 

Unfortunately it was breaking staging, with tests failing. See: https://github.com/Vizzuality/heco-invest/runs/6858283048

PR #317, opened 2022-06-24
Add map layers toggle component to Discover Projects' map

SARodrigues:
## IMPORTANT

This PR includes (because it depends on) #311 and #316 .

## Description

This PR adds a selector so the user can select contextual layers to display on the Discover page map. 

## Testing instructions

Verify that:  
- The user can click on the "Map layers" button and a selector opens  
- The user can toggle layers by clicking on them  
  _Note: the layers display is not part of this task, so some may not work. They've been added for demonstration/testing purposes._  
- If the user clicks outside the selector closes  

## Tracking

[LET-654](https://vizzuality.atlassian.net/browse/LET-654)

## Screenshot  

Screenshot 2022-06-24 at 10 38 24

## Video

https://user-images.githubusercontent.com/6273795/175551759-6ab5e996-7733-4125-a1d2-dd3e6783b747.mp4


SARodrigues:
> the map is not being displayed, 

The map didn't really get touched in this PR, but it was on #274. Do you have `NEXT_PUBLIC_MAPBOX_API_TOKEN` set? 
(See: https://github.com/Vizzuality/heco-invest/blob/develop/frontend/README.md)  

> neither for local or staging envs

It should be working on staging (https://staging.hecopilot.org), but you're right it doesn't work on vercel deployments unfortunately. @tiagojsag

PR #318, opened 2022-06-24
Private endpoint to return project developer's projects

agnessa:
This adds a new list endpoint under `api/v1/account/projects`, which only returns projects of the signed in project developer.

It was implemented similarly to the public search, e.g. filtering by full text was reused.

What I was also hoping to reuse was the sorting. However, here I have a doubt - how do we even sort by the columns in the design?
- name - fine (that works)
- category - theoretically easy, but we only store the English slug in the db, so not easy to sort by translation
- status - same problem
- instrument types - almost same problem, but it's an array of English slugs
- value, i.e. ticket size - almost same problem, but in fact we probably want to sort not by translation, but by the amount...
- location - I'm guessing this column is a combination of municipality and country, so theoretically we could just sort by municipality name?

So in the end I did not implement sorting, except for sorting by name which was already working - wasn't sure how to proceed with this.

## Testing instructions

swagger docs

## Tracking

https://vizzuality.atlassian.net/browse/LET-581

### DIFF
--- a/backend/spec/requests/api/v1/accounts/projects_spec.rb
+++ b/backend/spec/requests/api/v1/accounts/projects_spec.rb
@@ -70,18 +70,6 @@ RSpec.describe "API V1 Account Projects", type: :request do
       parameter name: "fields[project]", in: :query, type: :string, description: "Get only required fields. Use comma to separate multiple fields", required: false
       parameter name: :includes, in: :query, type: :string, description: "Include relationships. Use comma to separate multiple fields", required: false
       parameter name: "filter[full_text]", in: :query, type: :string, required: false, description: "Filter records by provided text."
-      parameter name: :sorting, in: :query, type: :string, required: false, description: "Sort records.",
-        enum: [
-          "name asc", "name desc",
-          # "category asc", "category desc", # TODO: this is stored as a slug in the db, i.e. in English
-          # "municipality asc", "municipality desc", # TODO: this shows as 'Leticia, Colombia' in the design
-          # "ticket_size asc", "ticket_size desc", # TODO: this is stored as a slug in the db, i.e. in English
-          # "instrument_type asc", "instrument_type desc", # TODO: this is an array of slugs
-          # "status asc", "status desc", # TODO: this is stored as a slug in the db, i.e. in English
-          "created_at asc", "created_at desc"
-        ]
-
-      let(:sorting) { "name asc" }
 
       it_behaves_like "with not authorized error", csrf: true, require_project_developer: true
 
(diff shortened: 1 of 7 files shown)


backend/spec/requests/api/v1/accounts/projects_spec.rb
```diff
@@ -62,6 +62,76 @@
...
+      parameter name: "filter[full_text]", in: :query, type: :string, required: false, description: "Filter records by provided text."
+      parameter name: :sorting, in: :query, type: :string, required: false, description: "Sort records.",
+        enum: [
+          "name asc", "name desc",
+          # "category asc", "category desc", # TODO: this is stored as a slug in the db, i.e. in English
+          # "municipality asc", "municipality desc", # TODO: this shows as 'Leticia, Colombia' in the design
+          # "ticket_size asc", "ticket_size desc", # TODO: this is stored as a slug in the db, i.e. in English
+          # "instrument_type asc", "instrument_type desc", # TODO: this is an array of slugs
+          # "status asc", "status desc", # TODO: this is stored as a slug in the db, i.e. in English
+          "created_at asc", "created_at desc"
```
martintomas, line 81: Hmm, I think that to cover all these sorting options with `API::Sorter` is gonna be tricky. Maybe we could rewrite this service to use ransack, which Admin is already using and just add some transformation table which translates FE option to ransack params. Difficult part is that `API::Sorter` is used for sorting everything (PDs/Projects/etc.) so preparation of ransack params would need to take this into account and tweak attributes appropriately --> for example `name` is localized for Projects, but not localized for Accounts, etc.

martintomas (approved):
Looks good ;). Only difference against public endpoint which I see is missing pagination, which is fine 👍

martintomas:
> there are now separate user stories for sorting and searching of this list, probably low priority

Searching should work  - ofc if we want same full text searching. Sorting is different story, that will need separate user story.

backend/spec/requests/api/v1/accounts/projects_spec.rb
tsubik: Yeah, I think using ransack in or instead of `API::Sorter` could make things easier.

tsubik:
I think it's worth to add pagination, as they will need it for sure.

tsubik:
What about pagination? Separate story or here?

agnessa:
I omitted it, because there don't seem to be pagination links in the design; and it's unlikely those lists are going to be long anyway

tsubik (approved):

PR #320, opened 2022-06-27
feat: Allow to change language at Backoffice

martintomas:
I am adding simple selectbox next to user name which allows to switch between currently supported languages. Selected language is stored at session --> not sure if we have some attribute at admin table that I could use for it so I have decided to put it into session.

## Testing instructions

Open Backoffice and try to change language as logged and not-logged user. All text localized with I18n or even with traco should react to this change.

## Tracking

https://vizzuality.atlassian.net/browse/LET-659


agnessa (approved):

tsubik (approved):
Works great :+1:

PR #322, opened 2022-06-27
Fix sign-in / sign-out redirect

barbara-chaves:
This PR fixes the sign-in and sign-out bugs

The bug: 

1. On previous updates, we changed the 'useQuery' for 'useLocalizedQuery', that aways add the locale to the queries on UseQuery.
2. The old mutations was using just the 'clientQuery.setQueryData(QUERYNAME)' without the locale as a query, so the react-query doesn't understand that it was the same query we were using with 'useLocalizedQuery'.  Exemple: useQuery('user') !== useQuery(['user', 'es'])
3. To fix it I added the locale to all the setQueryData after the mutations.

## Testing instructions

The user can sign-in and sign-out with the corrects redirections and menu options

## Tracking

None


SARodrigues:
@barbara-chaves Do you think it would be worth it to have a `setLocalizedQueryData` alongside `useLocalizedQuery` in `hooks/query.ts` file? 

I'm just thinking it may prevent similar issues in the future, in case someone needs to make changes to one hook; they'll see the other and know it has to change too.

SARodrigues (approved):
The note is just really a note. 

Works great!

PR #324, opened 2022-06-28
test: only approved entities can be favourited

martintomas:
Unapproved PDs/Projects/Investors cannot be marked as favourite. In reality they are not even publicly visible so it should not be possible to mark them as favourite at all, but I am adding extra tests to make sure that this policy always supported.

## Testing instructions

Not required :))

## Tracking

https://vizzuality.atlassian.net/browse/LET-528

### DIFF
--- a/backend/spec/requests/api/v1/favourite_projects_spec.rb
+++ b/backend/spec/requests/api/v1/favourite_projects_spec.rb
@@ -15,9 +15,12 @@ RSpec.describe "API V1 Favourite Project", type: :request do
       let(:user) { create :user, account: create(:account, :approved) }
 
       it_behaves_like "with not authorized error", csrf: true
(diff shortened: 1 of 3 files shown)


backend/spec/requests/api/v1/favourite_projects_spec.rb
```diff
@@ -15,9 +15,12 @@
       let(:user) { create :user, account: create(:account, :approved) }
 
       it_behaves_like "with not authorized error", csrf: true
-      it_behaves_like "with forbidden error", csrf: true, user: -> { create(:user, account: create(:account, :unapproved)) }
 
       response "200", :success do
+        schema type: :object, properties: {
```
tsubik, line 20: Hah, I keep forgetting about this :+1:

martintomas: Me too :D

tsubik (approved):

PR #328, opened 2022-06-28
Let 545 fe user can accept invitation to join account

barbara-chaves:
Description

Following an email link, the user should be able to accept or reject the invitation to join an account. If the user is not signed in, they must sign up first and after sign up, be taken back to the account invitation screen. 

Figma: [HeCo - Work in progress](https://www.figma.com/file/KC49uMR9t1YARmFR0Sr6aN/HeCo---Work-in-progress?node-id=743%3A18704) 

It was included the API integration for the user invitation, so the PD or Investor user should be able to send a list of emails to invite other users to they account

## Acceptance criteria

Added user will receive an email notifying him/her that she is invited to join the account - user will need to sign-up to the platform

## Testing instructions

1. Send your email to invitations (You can use [swagger](https://staging.hecopilot.org/backend/api-docs/index.html) - Invitation) or http://localhost:3000/en/dashboard/users. 

- If you already have an account on HeCo with this email, use [EMAIL]+1@[EMAIL.COM] to add as new email.
- I already sent 1 email to each of you for the first test

2. Open the email and click on the link
3. Change the path to localhost:3000
4. Refresh the page and follow the flow

## Tracking

[LET-545](https://vizzuality.atlassian.net/browse/LET-545)


clementprdhomme (changes requested):
In addition to my code review, I noticed that 3 flows may not work as expected. It is possible some issues come from the back-end.

**1/ Logged out light users lose access to their user account**

1. Create a new light user account
2. Sign out
3. Get invited to an existing account
4. Try to sign in with or without the invitation token

You'll see you'll consistently get the “Invalid email or password” error.

**2/ Logged in light users are not redirected**

1. Create a new light user account
2. Get invited to an existing account
3. Open the invitation link

You'll see you're not redirected to the dashboard. Instead, you're stuck on the account type selection screen.

**3/ Invitation opened with another account doesn't sign you out**

1. Invite an existing or non-existing user
2. Open the invitation link while you're logged in with another user/account

In that case, you're redirected to the dashboard (the invitation is ignored). I wonder if that's the correct behaviour since in the [flow diagram](https://github.com/Vizzuality/heco-invest/blob/develop/docs/flows/HeCO%20-%20Invitation%20flow.jpg), we said you should get signed out so you can sign in/sign up with the user that was invited.

frontend/pages/invitation.tsx
```diff
@@ -0,0 +1,142 @@
...
+              defaultMessage="You have been invited to join {accountName} account"
+              id="MZI6bG"
+              values={{
+                accountName: invitedUser?.account_name,
+              }}
+            />
+          </h1>
+          <p className="max-w-md m-auto text-base">
+            <FormattedMessage
+              defaultMessage="By accepting this invitation you will belong to  {accountName} Project Developers’ account."
```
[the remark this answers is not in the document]
barbara-chaves: I don't. I will remove it

frontend/pages/invitation.tsx
```diff
@@ -0,0 +1,137 @@
...
+              defaultMessage="You have been invited to join the {accountName} account"
+              id="zxHx8B"
+              values={{
+                accountName: invitedUser?.account_name,
+              }}
+            />
+          </h1>
+          <p className="max-w-md m-auto text-base">
+            <FormattedMessage
+              defaultMessage="By accepting this invitation you will belong to  {accountName} account."
```
clementprdhomme, line 101:
```suggestion
              defaultMessage="By accepting this invitation you will belong to the {accountName} account."
```

frontend/pages/invitation.tsx
```diff
@@ -0,0 +1,137 @@
...
+    }
+  }, [invitedUser, query.invitation_token, replace, user, userError]);
+
+  const handleAccept = () => {
+    acceptInvitation.mutate(query.invitation_token as string, {
+      onSuccess: () => push(Paths.Dashboard),
+    });
+  };
+
+  return !!user ? (
```
clementprdhomme, line 68: I think you added this check because of the flash I was describing:
> If I'm a logged out user, when invitedUserLoading and userLoading are false, there's a very brief flash of the page before I get redirected to the sign up or sign in page.

Unfortunately, this check created one bug and an undesired behaviour:
- Log out and follow [this link](http://localhost:3000/invitation?invitation_token=TEi1FeWtPyNmYson-2Ts). You'll see you get absolutely no feedback about what's happening. If you remove the check however, you see the loading screen and then the error is displayed: “Invalid invitation token”.
- While the data is loading, there is nothing displayed to the user. I think it was better when we had the loading spinner.

Do you think we could remove this check? You can ignore my previous feedback about the flash. It's not critical.

clementprdhomme (commented):
In addition to my few comments, the 3 flows I described in my previous review are still not working:
1. Logged out light users lose access to their user account
2. Logged in light users are not redirected
3. Invitation opened with another account doesn't sign you out

barbara-chaves:
> 1. Logged out light users lose access to their user acc

barbara-chaves:
> In addition to my few comments, the 3 flows I described in my previous review are still not working:
> 
> 1. Logged out light users lose access to their user account
    This looks like a BE issue. The sign-in endpoint responds with an error if the invited user (light) tries to sign-in
> 2. Logged in light users are not redirected 
    
> 3. Invitation opened with another account doesn't sign you out

@clementprdhomme

clementprdhomme (changes requested):
The code looks good but there are two issues that need addressing:

**3/ Invitation opened with another account doesn't sign you out**

1. Invite an existing or non-existing user
2. Open the invitation link while you're logged in with another user/account

In that case, you're redirected to the dashboard (the invitation is ignored). I wonder if that's the correct behaviour since in the [flow diagram](https://github.com/Vizzuality/heco-invest/blob/develop/docs/flows/HeCO%20-%20Invitation%20flow.jpg), we said you should get signed out so you can sign in/sign up with the user that was invited.

**4/ NEW − Unable to sign out**

1. Sign in
2. Once on the dashboard, immediately sign out
3. Click the link to sign in in the header
4. Click the link to sign up in the header

At this point, the user should see the sign up form. Instead of that, the user is redirected to the dashboard and still signed in.

You can see this issue on the video:
[Reproduction video](https://user-images.githubusercontent.com/6073968/178541112-987faa95-4f8c-4d4f-a8fe-f4298c447d75.webm)

barbara-chaves:
> **3/ Invitation opened with another account doesn't sign you out**
> 
FIxed

clementprdhomme (commented):
I noticed that if I'm signed in with a user that is not the one of the invitation, I'm correctly signed out but I'm not always redirected to sign in or sign up page with the token to accept the invitation.

I've also suggested some changed but please **do not implement them** now. I'd like to get some help with @SARodrigues to understand why they wouldn't work.

Let's take an example. If the content of `pages/sign-out.tsx` would be changed to:
```tsx
import { useEffect } from 'react';

import { useRouter } from 'next/router';

import { Paths } from 'enums';

import { useSignOut } from 'services/authentication/authService';

const SignOut = () => {
  const signOut = useSignOut();
  const { replace } = useRouter();

  useEffect(() => {
    signOut.mutate(
      {},
      {
        onSuccess: () => {
          // Moved the `replace` call here inside and removed the `invalidateQueries` since I
          // believe it wouldn't be necessary anymore (`useSignOut` calls `removeQueries`).
          replace(Paths.Home);
        },
      }
    );
    // I also removed the dependencies here to avoid an infinite loop
  }, []);

  return null;
};

export default SignOut;
```

Why is signing out not working anymore?

clementprdhomme (approved):

PR #334, opened 2022-06-29
Let 651 fe changes on about page

barbara-chaves:
## Description

As an HeCo manager, I need that visitors can see HeCo detailed general information, so that investors and project developers understand its purpose and are willing to participate

See “[About page” design](https://www.figma.com/proto/L9ROteAu5IhPDVe4MzPzyQ/%5BSHARED%5D_-Final-deliver_Phase-01?node-id=7%3A18421&viewport=500%2C48%2C0.15&scaling=min-zoom&starting-point-node-id=7%3A19062&show-proto-sidebar=1) for reference

## Acceptance criteria

There will be a page with detailed general information about HeCo

Note: “About” page.

It will be accessible from the home page (LET-219)

Note: this shouldn’t create a dependency. If this page is developed before the home page, this acceptance criterion should be removed and the link described in a different user story

The page will include the following information, complying with the structure and design defined at “[About page” design](https://www.figma.com/file/KC49uMR9t1YARmFR0Sr6aN/HeCo---Work-in-progress?node-id=2%3A5327)

About the platform

About HeCo

About the partners

About “ARIES”

About the impact

## Tracking

Link to the task(s), if any


SARodrigues (changes requested):
@barbara-chaves It looks like several of the images aren't loading (nor locally nor on vercel/staging). See screenshot below.  

I haven't tested this locally, but I would wage it may be due to the use of `useBreakpoint` in these pages, which wouldn't work with SSR'd pages. 

Either way, I would use tailwind's helper classes to hide/show elements here. `useBreakpoint` is less performant, and really only intended to swap components based on the device size; not to hide/show components like that. 



**Screenshot** 
screencapture-localhost-3000-pt-about-2022-07-05-12_51_00

clementprdhomme (changes requested):
In addition to my comments, some of the images you added don't seem to be used:
- `about-about-heco-large.png`
- `about-about-section.png`
- `about-biodiversity.png`
- `about-climate.png`
- `about-community.png`
- `about-impact.png`
- `about-water.png`

PR #335, opened 2022-06-30
Migrate to new domain with redirects; refactor TF dns resources

tiagojsag:

- You ran the tests
    
---

Description of the PR

## Testing instructions

What to test? How to do it?

## Tracking

Link to the task(s), if any


infrastructure/base/main.tf
```diff
@@ -8,38 +8,57 @@ terraform {
...
-  github_project       = var.github_project
-  github_branch        = "develop"
-  google_analytics_key = var.google_analytics_key
-  google_maps_api_key  = var.google_maps_api_key
-  project_name         = var.staging_project_name
-  transifex_token      = var.transifex_token
-  sendgrid_api_key     = var.sendgrid_api_key
-  mapbox_api_key       = var.mapbox_api_key
+  source                 = "./modules/env"
+  domain                 = var.domain
```
agnessa, line 12: why drop the "staging" here?

tiagojsag: there's a new prefix var to account for that. we need them separated to account for the fact that the DNS zone is centralized

agnessa: OK maybe I should read till the end before asking stupid questions

agnessa:
right, so after reading 3 times I think I get it, safe to merge?

PR #336, opened 2022-06-30
feat: Create Admin via backoffice

martintomas:
Adding formular for Admin creation to backoffice. 

When new Admin is created, email with password reset link is send. Admin can use this link to set up new password -- this part of code uses Forgotten password feature from Devise. When I was implementing this, I have also enabled to trigger Forgotten password flow from sign-in page.

**Note**: We check password complexity and I was not able to find some simple generator which would generate password based on our requirements so I have written simple one just for your needs.

## Testing instructions

Backoffice: list of Admins --> click on `Add administrator` --> fill form and save it

## Tracking

https://vizzuality.atlassian.net/browse/LET-669

### DIFF
--- a/backend/app/controllers/backoffice/admins_controller.rb
+++ b/backend/app/controllers/backoffice/admins_controller.rb
@@ -24,7 +24,7 @@ module Backoffice
     end
 
     def create
-      if @admin.update admin_params.merge(password: GeneratePassword.new(50).call, ui_language: current_admin.ui_language)
+      if @admin.update admin_params.merge(password: GeneratePassword.new(50).call)
         AdminMailer.first_time_login_instructions(@admin).deliver_later
         redirect_to backoffice_admins_url, notice: t("backoffice.messages.success_create", model: t("backoffice.common.admin"))
       else
@@ -38,6 +38,7 @@ module Backoffice
       params.require(:admin).permit(
         :first_name,
         :last_name,
+        :ui_language,
         :email
       )
     end
--- a/backend/app/controllers/backoffice/admins_controller.rb
+++ b/backend/app/controllers/backoffice/admins_controller.rb
@@ -1,5 +1,8 @@
 module Backoffice
   class AdminsController < BaseController
+    before_action :initialize_admin, only: [:new, :create]
+    before_action :set_breadcrumbs, only: [:new, :create]
+
     def index
       @q = Admin.ransack params[:q]
       @admins = @q.result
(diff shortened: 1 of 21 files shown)


backend/app/controllers/backoffice/admins_controller.rb
```diff
@@ -16,5 +19,36 @@ def index
         end
       end
     end
+
+    def new
+    end
+
+    def create
+      if @admin.update admin_params.merge(password: GeneratePassword.new(50).call, ui_language: current_admin.ui_language)
```
martintomas, line 27: `ui_language ` is mandatory for `Admin`, but formular from design does not contain this field soo I have taken its value from current admin. Not sure if it is correct or if I should add this field to formular.

backend/config/locales/zu.yml
```diff
@@ -192,6 +198,12 @@ zu:
     rejected:
       subject: Your account was rejected!
       content_html: We would like to inform you that your account at HeCo platform was rejected. One of our platform administrators should reach you shortly, but feel free to contact us at %{contact_url}.
+  activerecord:
```
tsubik, line 201: Hmm, should we keep consistency with the existing forms and use `simple_forms` namespace?

backend/app/controllers/backoffice/admins_controller.rb
tsubik: I think this should be added. What do you think @agnessa?

backend/config/locales/zu.yml
tsubik: Actually, this will deal with error messages as well, right? And simple_forms will not. Maybe then we should move all simple_forms to attributes, even if those labels are long and not named as attributes, but not in this PR. I would have to check what will happen with existing translations when we change key names but not value, maybe Transifex is intelligent about that.

martintomas: Oki douky, I will move new text under `simple_form` prefix for now and we can refactor it later within separate PR ;)

backend/app/controllers/backoffice/admins_controller.rb
agnessa: agree

backend/config/locales/zu.yml
agnessa: I find the possible aftermath terrifying :) I don't have high hopes for anything working automatically, but I'm not sure. Perhaps rather than renaming we could copy the strings under the new namespace, that will at least make it easier to copy existing translations rather than losing them and starting from scratch?

tsubik: I think I did similar things very long time ago, and it worked, but who knows what have changed. We already know that Transifex CLI is not very friendly.

tsubik (approved):
Works like a charm :+1:

PR #339, opened 2022-07-01
feat: Possibility to edit Admins via backoffice

martintomas:
I am adding possibility to edit already existing Admins. It is possible to edit all attributes of user, even email --> there is no email confirmation for Admin for now (not sure if this will change at future) so changing email is simple.

Password can be changed only if you are editing admin which is one currently logged in. After password is changed, user stays logged in (default behaviour of Devise is logout user).

Btw. I guess this is clear, but be super careful when you register new Admins and use only emails which you own :). If by any chance we would create Admin with wrong email (typo, etc.), owner of such email could login as Admin and take over whole system.

## Testing instructions

Via Bakoffice -- create new Admin and try to edit it ;). You should be able to edit password only if you are editing yourself. You should be also able to delete this newly created record.

## Tracking

https://vizzuality.atlassian.net/browse/LET-672
https://vizzuality.atlassian.net/browse/LET-673
https://vizzuality.atlassian.net/browse/LET-674

### DIFF
--- a/backend/app/controllers/backoffice/admins_controller.rb
+++ b/backend/app/controllers/backoffice/admins_controller.rb
@@ -4,7 +4,7 @@ module Backoffice
 
     before_action :initialize_admin, only: [:new, :create]
     before_action :fetch_admin, only: [:edit, :update, :destroy]
-    before_action :set_breadcrumbs, only: [:new, :create]
+    before_action :set_breadcrumbs, only: [:new, :create, :edit, :update]
     before_action :set_sections, only: [:edit, :update]
 
     def index
--- a/backend/app/controllers/backoffice/admins_controller.rb
+++ b/backend/app/controllers/backoffice/admins_controller.rb
@@ -1,7 +1,11 @@
 module Backoffice
   class AdminsController < BaseController
+    include Sections
+
     before_action :initialize_admin, only: [:new, :create]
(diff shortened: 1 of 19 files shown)


agnessa (approved):

tsubik:
> Btw. I guess this is clear, but be super careful when you register new Admins and use only emails which you own :). If by any chance we would create Admin with wrong email (typo, etc.), owner of such email could login as Admin and take over whole system.

To prevent this, we could remove reset_password token every time email changes. Administrator can also delete a user and create a new one, but I think better to prevent the current user to log.

tsubik (changes requested):
Looks great. Only one small thing.

Changing breadcrumb to "Administrator -> Name" feature is missing.

martintomas:
> > Btw. I guess this is clear, but be super careful when you register new Admins and use only emails which you own :). If by any chance we would create Admin with wrong email (typo, etc.), owner of such email could login as Admin and take over whole system.
> 
> To prevent this, we could remove reset_password token every time email changes. Administrator can also delete a user and create a new one, but I think better to prevent the current user to log.

Hmm, yeah, this could work, but only in case that you realise your mistake right after you have created admin. I will keep it as it is for now and deletion of Admin can be way in case somebody does similar mistake at future, but I agree with you that removing reset_password token is better way --> can do it as part of different PR if you think it is worthy ;).

PR #342, opened 2022-07-05
FE - Fix logic/permissions on favouriting PDs, Investors & Projects

SARodrigues:
## Description

This PR fixes inconsistencies / non-working / issues with favouriting projects, project developers & investors. 
Some didn't work at all (eg: favouriting investors), so the functionality was added. 

**In this PR:**  
- Added the possibility of favouriting an investor; it was not implemented  
- Added the possibility of favouriting a project from the project page; it was not implemented  
- Fixed "Favourite" and "Contact" buttons not working on public pages  
  _The data we fetch in `getServerSideProps` never contains whether a PD/project/investor is a favourite nor the contacts. I don't think the solution in the PR is the best one (added a note to the respective places), but what I've done is re-fetch the data on the client, which will include them (assuming the user is logged in)_
- When favouriting a project from the project details card, I'm setting the new query data with `setQueryData`, but also invalidating the `ProjectList` query, to cause a (hidden) refetch and ensure the project list is up to date. 
  _Reasoning being that if the user closes the project details card and re-opens it, the "favourite" property would be out of date because we pass the project data as a prop._

&nbsp;  

**Others:**
- Added a `projectContacts` helper to centralise how we extract the contacts for a given project and reduce the risk of bugs  
- Fixed a couple incorrect `useQuery` keys making query invalidation not work correctly  
- Swapped the `jsona` package with `deserialize-json-api`  
  Although they both do the same, `jsona` was causing issues due to circular references. We were using `cycle` to solve this, but it wasn't a perfect solution. `deserialize-json-api` handles deserialisation better and removes all issues. 
- Fixed some linting issues  

## Testing instructions

- As a signed-in user:  
  - When viewing a project page, I can:  
    - Favourite & unfavourite a project  
    - The Favourite / Unfavourite button will reflect the current status
  - When viewing the search pages, Projects tab, I can: 
    - Favourite and unfavourite a project by clicking on the Favourite button 
    - The Favourite / Unfavourite button will reflect the current status
    - The project card contains two Favourite buttons, both will update in sync
    - Favouriting / unfavouriting a project will not cause the map pins to disappear 
  - When viewing a project developer page, I can:  
    - Favourite & unfavourite the project developer  
    - The Favourite / Unfavourite button will reflect the current status 
  - When viewing an investor page, I can:  
    - Favourite & unfavourite the investor  
    - The Favourite / Unfavourite button will reflect the current status 

- As a non-signed-in user:  
  - When viewing a project page:  
    - The Favourite / Unfavourite button will be disabled
  - When viewing the search pages, Projects tab:
    - The Favourite / Unfavourite buttons in the Project details card will be disabled
  - When viewing a project developer page:  
    - The Favourite / Unfavourite button will be disabled
  - When viewing an investor page:  
    - The Favourite / Unfavourite button will be disabled
  

## Tracking

[LET-720](https://vizzuality.atlassian.net/browse/LET-720)


clementprdhomme (changes requested):
I found one major issues:
- If you favourite an investor and reload the page, the button doesn't have the correct state

frontend/containers/project-details/component.tsx
```diff
@@ -77,6 +78,20 @@ export const ProjectDetails: FC<ProjectDetailsProps> = ({
...
 
+  const favoriteProject = useFavoriteProject();
+
+  const handleFavoriteClick = () => {
+    // This mutation uses a 'DELETE' request when the isFavorite is true, and a 'POST' request when is false.
+    favoriteProject.mutate(
+      { id: project.id, isFavourite: project.favourite },
+      {
+        onSuccess: (data) => {
+          project.favourite = data.favourite;
```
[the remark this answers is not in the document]
SARodrigues: I refactored the logic pertaining to how a project is selected, and this is no longer necessary.

frontend/services/api.ts
```diff
@@ -48,7 +46,7 @@ const onResponseSuccess = (response) => {
     return {
       data: {
         // Deserialized data
-        data: response.data ? dataFormatter.deserialize(response.data) : null,
+        data: response.data ? deserialize(response.data)?.data : null,
```
clementprdhomme, line 49: I've had a hard time tracking it but I think this new library is flawed or not configured correctly. For example, try to load this page: http://localhost:3000/en/project/pd-organization-test-1-clement-s-project.

You'll see that the page crashes because `project.project_images` is `undefined`. Nevertheless, the API responds correctly with the relationship.

Could you have a look?

clementprdhomme (changes requested):
All good: the code looks good and the feature works as expected. Unfortunately, the current code introduces a regression. Could you have a look?

frontend/services/api.ts
SARodrigues: That's the one difference I found when adding this library. When deserialising and if a relationship doesn't have results, the library will deserialise the empty array into `undefined`/`null`. Unfortunately due to that we always assumed we'd have an array for `involved_project_developers` or `project_images`, even if it was empty. 

I thought I had addressed most circumstances but it looks like I missed too. There was also a hook to deal with extracting a project's contacts which wasn't being used here.

SARodrigues:
@clementprdhomme 
>All good: the code looks good and the feature works as expected. Unfortunately, the current code introduces a regression. Could you have a look?

I've reverted the change, it's no longer needed either. :)

clementprdhomme (approved):

PR #350, opened 2022-07-08
feat: Notify PD on changes related to project involvement

martintomas:
I am adding `ProjectInvolvement` model which covers already existing table called `project_involvements`. I have decided to add class for this, because I have needed `after_create`/`after_destroy` hooks which are used to send notifications to PDs.

Mails for PD are handled by new mailer called `ProjectDeveloperMailer` <-- I have found it more transparent to move it to special mailer, but It can be theoretically also part of `UserMailer`.

## Testing instructions

Everything should be testable via Rswag documentation --> find some project and update `involved_project_developer_ids` param. Added/removed PDs should receive appropriate email.

## Tracking

https://vizzuality.atlassian.net/browse/LET-525
https://vizzuality.atlassian.net/browse/LET-524

### DIFF
--- a/backend/app/mailers/project_developer_mailer.rb
+++ b/backend/app/mailers/project_developer_mailer.rb
@@ -1,0 +1,15 @@
+class ProjectDeveloperMailer < ApplicationMailer
+  def added_to_project(project_developer, project)
+    @user = project_developer.owner
+    @project = project
+
+    mail to: @user.email
+  end
+
+  def removed_from_project(project_developer, project)
+    @user = project_developer.owner
+    @project = project
+
+    mail to: @user.email
+  end
+end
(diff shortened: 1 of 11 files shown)


agnessa:
I think we might have oversimplified here, because we didn't consider the project draft status. Now that I'm trying to think this through, it seems that another project developer would be notified even if the project is still in draft, correct? That probably might be premature, because they won't have permissions to see the project page (unless they should have access?). I don't have a good solution to this, e.g. we could change the after_create callback to an after_update which checks that the transition from draft to published has just occurred, but then that transition can happen multiple times 🙄

martintomas:
I see, I have not taken `draft` status into account at all. It should be quite straightforward to send notification only when PD is added/removed to/from `published` project. What I find more tricky is part about when project changes its status, because If I understand it correctly:
 - when project becomes published --> send notification to all involved PDs
 - when project stops to be published --> ?? should I send notification as well?
I agree with @tsubik with repeated sending -- It makes sense to send notification every time when status changes, because users every time also get or lose access to project. What I am not 100% sure is if e-mails should be same or we should have specific ones which would inform PDs that Project which they are part of became visible.

agnessa (approved):

PR #355, opened 2022-07-11
refactoring: Only Admins can access background job events API

martintomas:
Just minor tweak of `BackgroundJobEvents` API which ensures that only Admins can access it <-- it can contain some sensitive information so that is why I am putting there this restriction.


tsubik (commented):
What is that API for? It is just for us, developers?

agnessa (approved):

martintomas:
> What is that API for? It is just for us, developers?

Yup, it is just for us/devs. Idea was to give us some kind of overview about background jobs. Not sure if this approach is already obsolete and could be either removed or moved to backoffice, but I am just making sure that normal users cannot access it..

PR #365, opened 2022-07-11
fix(front): add map searcher to project form [LET-406]

anamontiaga:
This PR solves a google API load problem on projects list `en/discover/projects` and add Location Searcher component to project form `http://localhost:3000/en/projects/new`

NEED HELP: Google API has problems to load on project form, even though I tried to load it in layout components.

## Testing instructions

Enter in above described pages and try to access them by normal flow and reloading.

## Tracking

[LET-406](https://vizzuality.atlassian.net/browse/LET-406)


clementprdhomme:
Let's not merge just now. I'm testing the Google Maps key.

PR #366, opened 2022-07-11
fix(front): home improvements[LET-650]

anamontiaga:
This PR contains next improvements:

- Some texts should be bold in the “Create open calls”, “Enable positive impact” and “Apply to open calls” section
- The titles of the cards in “How it works” should have a different font weight (should be 600)
- When the screen is between 640px and 768px wide, there is too much space above the “Create account” button at the bottom of the screen

## Tracking

[LET-650](https://vizzuality.atlassian.net/browse/LET-650)


clementprdhomme (approved):
Thank you Ana!

PR #373, opened 2022-07-13
chore: update of Rails to 7.0.2

martintomas:
I have got bundle audit warning, so here comes fix which updates rails (activerecord) version...


tsubik (approved):
Normally I do `bundle update` ALL from time to time when having test coverage above 90%, should we do it also at some point?

PR #374, opened 2022-07-13
feat: List of users at Backoffice

martintomas:
This PR adds list of users to backoffice.

Screenshot 2022-07-13 at 11 53 57

## Testing instructions

Via backoffice ;)

## Tracking

https://vizzuality.atlassian.net/browse/LET-760


agnessa:
You're right it will be necessary, the task was created yesterday: https://vizzuality.atlassian.net/browse/LET-760

agnessa (approved):

PR #378, opened 2022-07-14
Change project data language to the original PD language when editing

barbara-chaves:
This PR fixes the project edition language

## Description

PD users are able to see the platform in any language but when they edit a project, its data must be shown and edited in the language of the account.

For example, an account can be in English, but the user may use the platform in Spanish. If they start editing a project, all the UI should be in Spanish (including the labels of the inputs) but the project’s data must be loaded and displayed in English.

## Reproduction steps

1. Log in with the “HeCo approved PD user” user (the account is in English)
2. Change the language of the platform to Spanish or Portuguese
3. Go edit the project “Awesome project (Edited)”

You must see the UI in Spanish or Portuguese (according to your choice at step 2) but the fields must display the data in English.

Please note that the language of the account (English in this example) will be displayed on the form when [LET-727: FE - Add message to Project, PD & Investor edition forms informing the users of the account original languageIN PROGRESS](https://vizzuality.atlassian.net/browse/LET-727) is done.

## Tracking

[LET-766](https://vizzuality.atlassian.net/browse/LET-766)


clementprdhomme (changes requested):
@barbara-chaves If the UI is in Spanish or Portuguese, the data is still loading in these languages instead of English. Did you not find an issue during your testing?

The problem comes from here: https://github.com/Vizzuality/heco-invest/blob/aa0e0b5a1af752997b23f7cd6b8136daff3f84ed/frontend/services/api.ts#L24-L42

As you see if `config?.params?.locale` is falsy (`undefined` for example), we will still send the UI language. We could circumvent that by replacing the `locale` value by this:
```ts
      locale:
        config?.params?.locale === null
          ? null
          : getCookie('NEXT_LOCALE') ?? locales.find((locale) => locale.default).locale,
```

That also means that inside `pages/project/[id]/edit.tsx`, `locale` should be sent as `null`, not `undefined`.

If I'm not clear, please let me know and I can explain to you.

clementprdhomme:
@barbara-chaves Could you undo the commit [a5f47df](https://github.com/Vizzuality/heco-invest/pull/378/commits/a5f47df1b71a665f936956f4f71d4ebdc72b5e52)? I don't think it's necessary.

PR #381, opened 2022-07-14
feat: Allow edit and deletion of User via Backoffice

martintomas:
I am adding possibility to edit existing users --> you can modify `first_name`, `last_name` and `avatar` (new attribute added to User model).

It is also possible to delete User either via button next to edit formular or at index page after clicking on menu next to appropriate user record. I have made it impossible to delete Users which are owners of accounts (`restrict_with_error`) --> in such case, user receives error that this particular record cannot be deleted. Delete option for such records  will be automatically grayed out as part of some future ticket, but for needs of these tickets, error should be enough ;).

Screenshot 2022-07-14 at 15 26 46
Screenshot 2022-07-14 at 15 26 37
## Testing instructions

Everything should be quite easily testable via Backoffice.

## Tracking

https://vizzuality.atlassian.net/browse/LET-761
https://vizzuality.atlassian.net/browse/LET-762

### DIFF
--- a/backend/app/views/backoffice/base/_menu.html.erb
+++ b/backend/app/views/backoffice/base/_menu.html.erb
@@ -1,9 +1,9 @@
 <div class="relative" data-controller="menu">
-  <button data-action="menu#toggle click@window->menu#hide" class="flex items-center rounded-full font-sans transition-all focus-visible:outline focus-visible:outline-2 focus-visible:outline-offset-2 disabled:opacity-60 disabled:pointer-events-none px-0 py-0 focus-visible:outline-green-dark" role="button" tabindex="0" id="react-aria-30" aria-haspopup="true" aria-expanded="true" aria-controls="react-aria-31" style="user-select: none;">
+  <button data-action="menu#toggle click@window->menu#hide" class="flex items-center rounded-full font-sans transition-all focus-visible:outline focus-visible:outline-2 focus-visible:outline-offset-2 disabled:opacity-60 disabled:pointer-events-none px-0 py-0 focus-visible:outline-green-dark" role="button" tabindex="0" aria-haspopup="true" aria-expanded="true" style="user-select: none;">
     <%= svg "menu" %>
   </button>
   <div>
-    <div data-menu-target="menu" role="menu" tabindex="0" id="react-aria-8" aria-labelledby="react-aria-7" class="hidden z-30 absolute transform whitespace-nowrap bg-white shadow-2xl rounded-md overflow-hidden right-0 -bottom-2 translate-y-full">
+    <div data-menu-target="menu" role="menu" tabindex="0" class="hidden z-30 absolute transform whitespace-nowrap bg-white shadow-2xl rounded-md overflow-hidden right-0 -bottom-2 translate-y-full">
       <ul class="">
         <%= yield if yield.present? %>
       </ul>
--- a/backend/app/views/backoffice/base/_menu.html.erb
+++ b/backend/app/views/backoffice/base/_menu.html.erb
@@ -1,0 +1,12 @@
+<div class="relative" data-controller="menu">
+  <button data-action="menu#toggle click@window->menu#hide" class="flex items-center rounded-full font-sans transition-all focus-visible:outline focus-visible:outline-2 focus-visible:outline-offset-2 disabled:opacity-60 disabled:pointer-events-none px-0 py-0 focus-visible:outline-green-dark" role="button" tabindex="0" id="react-aria-30" aria-haspopup="true" aria-expanded="true" aria-controls="react-aria-31" style="user-select: none;">
+    <%= svg "menu" %>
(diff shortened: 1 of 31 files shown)


backend/app/views/backoffice/base/_menu.html.erb
```diff
@@ -0,0 +1,12 @@
+<div class="relative" data-controller="menu">
+  <button data-action="menu#toggle click@window->menu#hide" class="flex items-center rounded-full font-sans transition-all focus-visible:outline focus-visible:outline-2 focus-visible:outline-offset-2 disabled:opacity-60 disabled:pointer-events-none px-0 py-0 focus-visible:outline-green-dark" role="button" tabindex="0" id="react-aria-30" aria-haspopup="true" aria-expanded="true" aria-controls="react-aria-31" style="user-select: none;">
```
tsubik, line 2: Please remove all `react-...` placed ids. Either remove aria altogether or generate the own IDs to connect various html elements.

backend/app/views/backoffice/users/index.html.erb
```diff
@@ -21,16 +21,25 @@
...
         <td><%= user.full_name %></td>
         <td><%= user.email %></td>
-        <td><%= user.account&.name %></td>
+        <td><%= account_link_for user.account %></td>
         <td><%= I18n.l user.created_at.to_date %></td>
         <td><%= I18n.l user.last_sign_in_at&.to_date, default: '' %></td>
+        <td><%= link_to t("backoffice.common.edit"), edit_backoffice_user_path(user.id), class: "link-button" %></td>
+        <td>
+          <%= render "menu" do %>
+            <%= render "backoffice/base/menu/delete", path: backoffice_user_path(user), model: t("backoffice.common.user") %>
```
tsubik, line 40: Would not be better to have common `menu_item` partial and write this as 

```
<%= render "menu-item %>
  <%= link_to t("backoffice.common.delete"), backoffice_user_path(user),
       data: {
         turbo_method: :delete,
         turbo_confirm: t("backoffice.messages.delete_confirmation", model: model.downcase)
       } %>
<% end %>
```

We need to put Reject and Approve also inside that menu.

tsubik (approved):
Nice :+1:

agnessa (approved):
Works great, I just noticed 2 things which happen here that fall outside of this task:
- no way of knowing if there is an image already or not. We have a [task in backlog](https://vizzuality.atlassian.net/browse/LET-628?atlOrigin=eyJpIjoiOTBhOTNmODg5ZjQ0NDAzN2JhOGJiNGY1ZTA4MDViYzAiLCJwIjoiaiJ9) to deal with that, because the same happens in PD / Investor forms. How difficult is it, can we add it to this sprint?
- we're not letting them modify the ui language. I have a feeling we should, or we should remove that column, so I'll ask about this. I think we need a wider review of how we handle language, because we have that column in every table (since we wanted to be prepared for anything) and maybe there's a bit too much of it (because some cases might never occur) and we have already reached a point where I'm not exactly sure which language is meant, as in [here](https://vizzuality.atlassian.net/browse/LET-766?atlOrigin=eyJpIjoiZWRiNTY2YmUxMmE2NDlhMThkOWU3YWQyY2NlNzQwZjYiLCJwIjoiaiJ9), where the story says to view project in the language of the account, but we also have the language of the project, which might be different if the account language was changed in between.

martintomas:
> Works great, I just noticed 2 things which happen here that fall outside of this task:
> 
> * no way of knowing if there is an image already or not. We have a [task in backlog](https://vizzuality.atlassian.net/browse/LET-628?atlOrigin=eyJpIjoiOTBhOTNmODg5ZjQ0NDAzN2JhOGJiNGY1ZTA4MDViYzAiLCJwIjoiaiJ9) to deal with that, because the same happens in PD / Investor forms. How difficult is it, can we add it to this sprint?

If you are fine with showing existing image above field for uploading new file/image, that should be quite simple to do and feel free to add this to sprint.

> * we're not letting them modify the ui language. I have a feeling we should, or we should remove that column, so I'll ask about this. I think we need a wider review of how we handle language, because we have that column in every table (since we wanted to be prepared for anything) and maybe there's a bit too much of it (because some cases might never occur) and we have already reached a point where I'm not exactly sure which language is meant, as in [here](https://vizzuality.atlassian.net/browse/LET-766?atlOrigin=eyJpIjoiZWRiNTY2YmUxMmE2NDlhMThkOWU3YWQyY2NlNzQwZjYiLCJwIjoiaiJ9), where the story says to view project in the language of the account, but we also have the language of the project, which might be different if the account language was changed in between.

Hmm, this sounds quite complex :). It is simple to add new field for changing ui_language, which will be then reflected for all NEW records, but to modify also languages across other records (account, PD, Investor, Projects, etc.) that could be tricky. I also believe that Admin can change account language manually in case that they need change to be reflected also there.

PR #392, opened 2022-07-18
refactoring: use salt for authentication

martintomas:
This tweak is related to bug discovered by @clementprdhomme and documented at https://vizzuality.atlassian.net/browse/LET-793. I have run couple of tests and I am almost 100% sure that bug is cased by FE implementation. First of all when testing same flow via Rswag documentation (https://staging.hecoinvest.org/backend/api-docs/index.html), logout always works properly. Backoffice uses same logic for login/logout and it works there as well. I believe that when FE calls logout API (which removes active user session from browser), it is returned by some piece of code back into play by FE. I was not able to find exact cause why this happens, sometimes logout works properly, sometimes it does not, from my point of view, everytime I clicked on `Account information` at user profile, I have not logged out properly afterwards so maybe problem is there.

Whatever, I am adding mechanism which invalidates auth session so even when It somehow survives at user browser after logout, It will not be valid anymore and user will be required to login again. There is just minor cons to this approach --> If user is logged in at multiple places at same time (for example desktop and mobile), he gets logged out from both devices on logout. I don't think this is issue for us, but to deal with this cons would require some more robust API key management, which goes again purpose of using sessions and keep auth simple.

## Testing instructions

Just check that logout via Rswag documentation still works. User should be always logged out at FE (test after merge ;))

## Tracking

https://vizzuality.atlassian.net/browse/LET-793

### DIFF
--- a/backend/app/controllers/api/v1/sessions_controller.rb
+++ b/backend/app/controllers/api/v1/sessions_controller.rb
@@ -17,7 +17,9 @@ module API
       end
 
       def destroy
-        sign_out current_user
+        user = current_user
+        sign_out user
+        user.invalidate_session!
         head :ok
       end
     end
--- a/backend/app/models/user.rb
+++ b/backend/app/models/user.rb
@@ -51,6 +51,14 @@ class User < ApplicationRecord
   end
   alias_method :to_s, :full_name
 
+  def invalidate_session!
+    update! token: SecureRandom.hex, remember_created_at: nil
+  end
+
+  def authenticatable_salt
+    "#{super}#{token}"
+  end
+
   private
 
   def confirmation_sent_within_limited_period?
--- a/backend/db/migrate/20220718082600_add_token_to_users.rb
+++ b/backend/db/migrate/20220718082600_add_token_to_users.rb
@@ -1,0 +1,5 @@
+class AddTokenToUsers < ActiveRecord::Migration[7.0]
+  def change
+    add_column :users, :token, :string
+  end
+end
--- a/backend/db/schema.rb
+++ b/backend/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2022_07_13_091005) do
+ActiveRecord::Schema[7.0].define(version: 2022_07_18_082600) do
   # These are extensions that must be enabled in order to support this database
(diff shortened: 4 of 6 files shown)


tsubik:
I think this should be fixed on frontend. Just looked at the requests in Network Tab and I think I know what is the problem. Sign out works well from Home Page it does not from Dashboard. Look what happens when signout out from Dashboard (waterfall column) 
image

There is racing between getting current project developer and DELETE session, result of current developer is later so session is set again. cc @clementprdhomme

tsubik:
The sign out should possible just simple DELETE request followed by frontend redirect to home page in my opinion. I see it is a Next.js page, which probably first loading current user. The solution would be either wait for current user to be loaded there or change it to be just request and redirect in hooks.

tsubik:
I'm not sure if this PR will help with all possible racing conditions.

martintomas:
> I think this should be fixed on frontend. Just looked at the requests in Network Tab and I think I know what is the problem. Sign out works well from Home Page it does not from Dashboard. Look what happens when signout out from Dashboard (waterfall column) image
> 
> There is racing between getting current project developer and DELETE session, result of current developer is later so session is set again. cc @clementprdhomme

Np, If this can be easily fixed on FE, I can close this PR later on ;). This is from my point of view more robust solution which fixes this as well and makes sure that user is truly always logged out even at situations when something like this occurs on FE. On other hand, comes with some cons :)

martintomas:
> I'm not sure if this PR will help with all possible racing conditions.

It should 🤔 . Ofc if FE sends request to some endpoint before logout finishes, it will not help. But all other requests after logout will be blocked and user will be required to sign-in.

tsubik:
Hmm, but I like the idea of invalidate all sessions, kinda another security layer, just thinking if that would help with everything. We don't use "remember me" token so as you said there will be almost no harm with logging out from all devices, as I guess anyone browsers will be closed.

martintomas:
> Hmm, but I like the idea of invalidate all sessions, kinda another security layer, just thinking if that would help with everything. We don't use "remember me" token so as you said there will be almost no harm with logging out from all devices.

Soo we can keep this :), but maybe complete fix of this bug will require also some FE tweak 🙃

tsubik:
Yeah, I guess so, just in case we change our minds or something will not work as we wished for.

tsubik (approved):

PR #394, opened 2022-07-18
test: Invitation API can be used to resend invitation.

martintomas:
I am adding test which ensures that invitation API `POST /api/v1/invitation` can be used to resend invitation (already invited users can be invited again). In such case, user receives new email with NEW invitation token --> that also means that previous email is no more usable.

## Testing instructions

Via Rswag -- invite some user twice. He should receive email for every invitation attempt.

## Tracking

https://vizzuality.atlassian.net/browse/LET-461

### DIFF
--- a/backend/spec/requests/api/v1/invitation_spec.rb
+++ b/backend/spec/requests/api/v1/invitation_spec.rb
@@ -104,7 +104,7 @@ RSpec.describe "API V1 Invitation", type: :request do
           end
 
           it "sends emails" do
-            mails = ActionMailer::Base.deliveries[-2..]
+            mails = ActionMailer::Base.deliveries.last(2)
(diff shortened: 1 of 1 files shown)


backend/spec/requests/api/v1/invitation_spec.rb
```diff
@@ -96,17 +96,37 @@
...
+          run_test!
 
-        it "matches snapshot", generate_swagger_example: true do
-          expect(response.body).to match_snapshot("api/v1/invitation-create")
+          it "matches snapshot", generate_swagger_example: true do
+            expect(response.body).to match_snapshot("api/v1/invitation-create")
+          end
+
+          it "sends emails" do
+            mails = ActionMailer::Base.deliveries[-2..]
```
tsubik, line 107: `last(2)` I think it is easier to read.

tsubik (approved):

PR #395, opened 2022-07-18
Update the Discover map's layers

clementprdhomme:
This PR updates the layers on the Discover maps so that they are functional and display correctly.

## Testing instructions

Turn on the layers one by one.

## Tracking

[LET-413](https://vizzuality.atlassian.net/browse/LET-413)


SARodrigues (approved):
That was a ton of work! Thank you!  

I'm curious as to why I couldn't get the deck layers to render/decode, but unfortunately I no longer have the branch I was working on.

PR #397, opened 2022-07-18
Return entities details in account language when no locale given

agnessa:
Changed default locale for `show` actions of projects, project developers, open calls and investors. When locale not given, these endpoints will return content in the language of the account. Please note: we're disregarding the language of original input here (i.e. the language column in respective tables) and instead we're always checking the account language. That is because currently we can disregard the scenario of account language change.

At the same time I tried to tidy up how we present language settings to FE. So far we were only exposing the "language" column of the 4 entities, which is kind of not used (based on existing user stories), and we were not exposing the account language or user ui language anywhere. So what I did is I added `account_language` to all serializers (user, project, project developer, open call, investor) and `ui_language` to user serializer.

## Testing instructions

I am not sure why locale was not previously possible to pass as parameter in swagger, I added it so this can be tested more easily.

## Tracking

https://vizzuality.atlassian.net/browse/LET-791?atlOrigin=eyJpIjoiZjA1OTU0ZmZiNjlkNGRlOWI3NWY2NTZkYzkwZGRjNDciLCJwIjoiaiJ9

https://vizzuality.atlassian.net/browse/LET-792?atlOrigin=eyJpIjoiNTZkOWI3NDg1MzZiNDYwZjgzMzVkNmI3MjVhZDEwZjYiLCJwIjoiaiJ9

### DIFF
--- a/backend/app/controllers/api/v1/investors_controller.rb
+++ b/backend/app/controllers/api/v1/investors_controller.rb
@@ -4,6 +4,7 @@ module API
       include API::Pagination
 
       before_action :fetch_investor, only: :show
+      around_action(only: [:show]) { |_controller, action| set_locale(@investor&.account&.language, &action) }
       load_and_authorize_resource
 
       def index
--- a/backend/spec/requests/api/v1/investors_spec.rb
+++ b/backend/spec/requests/api/v1/investors_spec.rb
@@ -6,9 +6,14 @@ RSpec.describe "API V1 Investors", type: :request do
     create_list(:investor, 6, sdgs: [1, 5])
     @unapproved_investor = create(:investor, account: create(:account, review_status: :unapproved, users: [create(:user)]))
     @approved_account = create(:account, review_status: :approved, users: [create(:user)])
+    @investor_in_pt = create(
+      :investor,
+      account: create(:account, about_en: "About EN", about_es: "About ES", about_pt: "About PT", language: "pt"),
+      sdgs: [1, 5]
+    )
   end
 
-  include_examples :api_pagination, model: Investor, expected_total: 7
+  include_examples :api_pagination, model: Investor, expected_total: 8
 
   path "/api/v1/investors" do
     get "Returns list of the investors" do
@@ -24,6 +29,7 @@ RSpec.describe "API V1 Investors", type: :request do
       parameter name: "filter[ticket_size]", in: :query, type: :string, required: false, description: "Filter records. Use comma to separate multiple filter options."
       parameter name: "filter[full_text]", in: :query, type: :string, required: false, description: "Filter records by provided text."
       parameter name: :sorting, in: :query, type: :string, enum: ["name asc", "name desc", "created_at asc", "created_at desc"], required: false, description: "Sort records."
+      parameter name: :locale, in: :query, type: :string, required: false, description: "Retrieve content in required language, skip for account language."
 
       let(:sorting) { "name asc" }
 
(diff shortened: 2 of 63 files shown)


backend/app/controllers/api/v1/investors_controller.rb
```diff
@@ -4,6 +4,7 @@ class InvestorsController < BaseController
       include API::Pagination
 
       before_action :fetch_investor, only: :show
+      around_action :set_locale, only: [:show]
```
martintomas, line 7: I have seen that this repeats here soo often and only difference in method implementation is how you retrieve record language. Maybe you could define `proc` (https://guides.rubyonrails.org/action_controller_overview.html#other-ways-to-use-filters) here and reuse `set_locale` everywhere:
```
around_action { |_controller, action| set_locale(@investor&.account&.language, &action) }, only: [:show]
```

backend/app/controllers/api/v1/investors_controller.rb
```diff
@@ -32,15 +33,22 @@ def show
       private
 
       def fetch_investor
-        return @investor = Investor.find(params[:id]) if fetching_by_uuid?
+        return @investor = Investor.includes(:account).find(params[:id]) if fetching_by_uuid?
```
martintomas, line 36: Not 100% sure, but when you use `find` to retrieve only one record, does it makes sense to use preloading with includes?

backend/spec/requests/api/v1/investors_spec.rb
```diff
@@ -126,6 +128,30 @@
 
           run_test!
         end
+
+        context "account language" do
+          let(:investor_in_pt) {
+            account = @investor.account
+            account.update_columns(about_en: "About EN", about_es: "About ES", about_pt: "About PT", language: "pt")
```
martintomas, line 135: To be honest, something about these `let` examples where you update columns feels wrong. I thing it makes more sense to just create new record which suits your needs or put this data manipulation into `before` block.

martintomas (commented):
I know this is just draft, but caught my attention 😊

backend/app/controllers/api/v1/investors_controller.rb
agnessa: sure, less of a problem than with list, but still one query instead of 2? and also makes you remember this other object is used for something in this call, maybe I like to be too explicit sometimes 😊

backend/spec/requests/api/v1/investors_spec.rb
agnessa: yes! it feels so wrong, but I got so tired of trying to create fresh objects, I get validation errors about duplicated email which I think come from a user object created somewhere I cannot put a finger on 😫

backend/app/controllers/api/v1/investors_controller.rb
agnessa: OK, actually I started wondering whether I need this at all, so I was going to remove the around_action, but this is better

backend/app/controllers/api/v1/investors_controller.rb
martintomas: Aha, oki douky. I though that 2 queries will be triggered as well in this case (one for loading investor and one for loading account) and power of includes is that when list of investors is obtained, it needs only one extra query to load all accounts. Whatever, will need to check doc at future to make sure :D

martintomas: Nice description of difference between includes/join/preload, etc. https://scoutapm.com/blog/activerecord-includes-vs-joins-vs-preload-vs-eager_load-when-and-where. From there:
```
Does includes always generate a separate query to fetch the records in the relationship?
No. includes will either use a separate query (like above) or a LEFT OUTER JOIN. If you have a where or order clause that references a relationship, a LEFT OUTER JOIN is used versus a separate query.
```
so in this case, it should truly use 2 queries 🤔

agnessa: OK, you're right, makes no sense! 👍

backend/spec/requests/api/v1/investors_spec.rb
agnessa: right, it seems to work in `before_all` 👍

martintomas (approved):

PR #398, opened 2022-07-18
Unapproved users are shown the "Pending approval" screen when trying to access "My area"

SARodrigues:
## Description

**In this PR:**
- Updated the `ProtectedPage` layout logic:  
  - Show the "Account pending approval" screen if the user's account is not approved  
    - Uses the shared `AccountPendingApproval` component introduced in #386   
    - The screen is displayed directly in the layout, to prevent unnecessary redirects  
    - Added a `allowUnapproved` prop, which defaults to `false`
      _To allow the default "approved accounts only" restriction to be bypassed, when/if necessary_
    - Removed the `allowConfirmed`, which no longer of use  
      _It was being used to prevent unconfirmed users' access to the project edition forms_  
      _Users with unapproved accounts are unable to access the project creation screen, and by extension create a project. Also, if they were to attempt to edit a project via direct url, the API would return a 404._
  - Added a `Loading` spinner to be shown while data is being loaded (along with the site's header)  
    _While the user/account data is loaded, the layout would be returning a blank page, which could be an issue with slower internet connections._ 
  - Minor conditions simplification  
- Updated the `dashboard/index` page logic  
  _This page's only function is to, when a user is redirected to `/dashboard`, identify whether the user is a Project Developer or an Investor and redirect them to their respective Dashboard page (Projects for Project Developers, Open Calls for Investors)_
  _If the user doesn't have an approved account, there's no need to redirect them because the `ProtectedPage`'s logic will kick in and the user will be shown the `AccountPendingApproval` screen_  
- Removed the dedicated `/project-developers/pending` and `investors/pending` pages  
  _And respective `Paths`_
  _Since the `ProtectedPage` layout can display this screen by itself, there's no longer a need to redirect the user to other pages. They all display the same content either way, which was now centralised in the `AccountPendingApproval` component._

**Notes:** We don't have a dedicated screen to show users whose accounts have been explicitly rejected, but they shouldn't be able to access the dashboard either (or create/edit projects). This PR makes it that the Dashboard is inaccessible to anyone without an explicitly approved account. 

## Testing instructions

1. With an **unapproved** _Project Developer_ or _Investor_ account  
    - Sign in  
      - Verify that you are shown the _"Pending Approval"_ screen  
    - With a signed in account
      - Verity that trying to access the dashboard pages via direct URL (eg: `/dashboard/users`) results in being shown the _"Pending Approval"_ screen

2. With an **approved** _Project Developer_ or _Investor_ account  
    - Sign in  
        - Verify that you are redirected to the Dashboard  
        - Verify that the Dashboard works normally  
        - Verify that all Dashboard tabs/pages are accessible

## Tracking

[LET-773](https://vizzuality.atlassian.net/browse/LET-773)

## Screenshot
Screenshot 2022-07-19 at 12 37 47


clementprdhomme (approved):
Great job!

PR #399, opened 2022-07-19
feat: Show image thumbnail inside backoffice forms

martintomas:
I am adding image thumbnail to formulars which allows to upload images (Users/Investors/PDs). Thumbnail is places above upload field.

## Testing instructions

Open Backoffice and check that User/Investor/PD records with images uploaded contain its thumbnail inside edit formular.

## Tracking

https://vizzuality.atlassian.net/browse/LET-628


agnessa (approved):
Love it :)

Screenshot 2022-07-20 at 09 18 20

PR #400, opened 2022-07-19
Fix bug preventing the user from signing out

clementprdhomme:
This PR addresses an issue where the user may not be able to sign out successfully.

The previous sign out flow was the following:
1. The user clicks the sign out button
2. The user would be brought to the `/sign-out` page
3. The page would render the static layout which would request information about the account (inside the header)
4. At the same time, inside the `useEffect`, the mutation to log out would take place
5. The user is redirected to the home

The two requests (to the account and to delete the session) would enter a race condition which would prevent the session from being consistently deleted on the API.

Instead, this is the new flow:
1. The user clicks the sign out button
2. The mutation to log out takes place
5. The user is redirected to the home

As you can see, the `/sign-out` page has been removed.

## Testing instructions

Sign in and sign out. Make sure there are no concurrent requests when signing out.

## Tracking

[LET-793](https://vizzuality.atlassian.net/browse/LET-793)


SARodrigues (approved):
LGTM! Signing out is so much faster now too, thank you @clementprdhomme

PR #403, opened 2022-07-20
feat: Notify involved project developers when project gets destroyed

martintomas:
I am adding `ProjectDeveloperMailer.project_destroyed` which is send when project is deleted. Email is send as part of controller action to `project_developer/owner` and all `involved_project_developers` of project.

Tricky part was to tweak `ProjectInvolvement ` model and its `after_create`/`after_destroy` callbacks which also sends collaboration emails so users are not getting two emails when project is destroyed (one that their collaboration ended, second that project was destroyed). Hopefully I have managed to catch all use cases --> please see tests ;).

During adding of new email, I have also added localization to already existing custom emails.

## Testing instructions

Delete project via Rswag doc or Backoffice --> you should receive information that project was deleted. No other email should be received.

## Tracking

https://vizzuality.atlassian.net/browse/LET-752
https://vizzuality.atlassian.net/browse/LET-759

### DIFF
--- a/backend/app/models/project_involvement.rb
+++ b/backend/app/models/project_involvement.rb
@@ -5,13 +5,13 @@ class ProjectInvolvement < ApplicationRecord
   validates_uniqueness_of :project_id, scope: :project_developer_id
 
   after_create do
-    if project.valid? && project.published? && !project.status_changed?
+    if project.valid? && project.published? && !project.status_changed? # notification is send only when project status has not changed to published, in such case project itself notifies all users on its own
       ProjectDeveloperMailer.added_to_project(project_developer, project).deliver_later
     end
   end
 
   after_destroy do
-    if project.valid? && project.published? && destroyed_by_association.blank?
+    if project.valid? && project.published? && destroyed_by_association.blank? # notification is send only when record is not destroyed by parent association, in such case parent takes care of notification
       ProjectDeveloperMailer.removed_from_project(project_developer, project).deliver_later
     end
   end
--- a/backend/app/models/project_involvement.rb
+++ b/backend/app/models/project_involvement.rb
@@ -5,10 +5,14 @@ class ProjectInvolvement < ApplicationRecord
   validates_uniqueness_of :project_id, scope: :project_developer_id
 
   after_create do
-    ProjectDeveloperMailer.added_to_project(project_developer, project).deliver_later if project.published?
+    if project.valid? && project.published? && !project.status_changed?
+      ProjectDeveloperMailer.added_to_project(project_developer, project).deliver_later
+    end
   end
 
   after_destroy do
-    ProjectDeveloperMailer.removed_from_project(project_developer, project).deliver_later if project.published?
+    if project.valid? && project.published? && destroyed_by_association.blank?
+      ProjectDeveloperMailer.removed_from_project(project_developer, project).deliver_later
+    end
   end
 end
--- a/backend/app/models/project_involvement.rb
+++ b/backend/app/models/project_involvement.rb
@@ -4,15 +4,19 @@ class ProjectInvolvement < ApplicationRecord
 
   validates_uniqueness_of :project_id, scope: :project_developer_id
 
-  after_create do
-    if project.valid? && project.published? && !project.status_changed? # notification is send only when project status has not changed to published, in such case project itself notifies all users on its own
-      ProjectDeveloperMailer.added_to_project(project_developer, project).deliver_later
+  after_create_commit do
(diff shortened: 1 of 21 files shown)


backend/app/models/project_involvement.rb
```diff
@@ -5,10 +5,14 @@ class ProjectInvolvement < ApplicationRecord
   validates_uniqueness_of :project_id, scope: :project_developer_id
 
   after_create do
-    ProjectDeveloperMailer.added_to_project(project_developer, project).deliver_later if project.published?
+    if project.valid? && project.published? && !project.status_changed?
```
agnessa, line 8: @martintomas  can you please add a comment to explain these conditions, i.e. the expected effect? just because I'm reading like 5th time and still not convinced I get it :D so for the `after_destroy`, we'll only notify them about being removed from project as collaborator if the destroy is not a result of a project destroy? And for the `after_create`, I don't even 😅 We only send the collaboration email if the project had already been published before a collaboration was added?

martintomas: Sure, I know it is complicated, didn't figure out better way to do it. For both callbacks applies that we send notification only when project is published and also valid. `after_destroy` is send only when `ProjectInvolvement` is not destroyed by Project itself (in such case we send notification to users that project was destroyed so no need to send extra mail that collaboration was removed). `after_create` is send only when status of project has not changed to `published` --> when you switch project status from `draft` to `published`, you send collaboration notification at once via Project callback so no need to send it twice. Basically tricky part here is that you can collaborators and project status via one update, but you need to send emails only once. Cha, will add comments :)

martintomas:
> I like how you handled the email localisation, although I'm not entirely sure about all cases where account language should have precedence over ui language. Just wondering what you think about this approach I took to localise `devise_invitable` emails? I'm actually not even sure how to test this. #407

Yeah, me neither, not sure about which language should be preferred at every use case, but it should be quite straightforward to change it at future if we decide to prefer first or second option.

backend/app/models/project_involvement.rb
```diff
@@ -5,10 +5,14 @@ class ProjectInvolvement < ApplicationRecord
   validates_uniqueness_of :project_id, scope: :project_developer_id
 
   after_create do
-    ProjectDeveloperMailer.added_to_project(project_developer, project).deliver_later if project.published?
+    if project.valid? && project.published? && !project.status_changed? # notification is send only when project status has not changed to published, in such case project itself notifies all users on its own
```
tsubik, line 8: Do we need `project.valid?` here? My guess this is because it's after_create and not after_commit. I think it is also possible to check changes in after_commit with `saved_change_to_attribute?`

martintomas: Actually we need it here (and it is one of the reasons I hate mails at callbacks). We assign new PD via `involved_project_developer_ids` at controller and flow is following:
 - immediately after `involved_project_developer_ids` changes, appropriate `ProjectInvolvement ` are created/destroyed
 - `ProjectInvolvement` callbacks are triggered
 - `Project` is validated
 - if something fails, it gets rollbacked (which actually does not help with emails)
 
There are two other approaches, but every has its pros and cons:
 - use nested_attributes --> but we would need extra stuff to API controller and probably to backoffice
 - use specialized service to send all emails after all changes to Project are saved --> it would need to collect deleted and added records
 
Soo for now I have added a lot of tests and went for this solution, which I think is easiest one, but we can theoretically go for different one.

tsubik: ok, it can stay this way, but I think `after_commit` won't have those problems, and the problem with rollback and email being already sent.

martintomas: that is also reason why I use `project.status_changed?`, because changes to project are not saved yet.

martintomas: Oki douky, I will try it @tsubik , if it helps, I will switch to `after_commit` callback ;)

agnessa (approved):

backend/app/models/project_involvement.rb
tsubik: But I'm not sure if `after_delete` with `after_commit` on: :destroy you will have that info about `destroyed_by_association`

tsubik: Maybe we cannot run from the service layer and eventually that would be a good choice to add a couple services, at least for projects, to update or delete.

martintomas: Agree with you, at future we should probably move more complex actions to services, sometimes that logic is starting to be quite complex and controller/model stuff is not enough

tsubik (approved):

PR #407, opened 2022-07-21
Send invitation email in account language

agnessa:
Invitation email for joining an account should be sent in account language

## Testing instructions

Invite a user to an account, expect email in correct language

## Tracking

https://vizzuality.atlassian.net/browse/LET-796?atlOrigin=eyJpIjoiZjE3M2FiYjhjZTRhNGEyNTg4M2Q2MjU1ZjkyM2U4YjQiLCJwIjoiaiJ9
https://vizzuality.atlassian.net/browse/LET-794?atlOrigin=eyJpIjoiYmMwYzBlYjI3YTQwNDI3Nzk5MzAyZDVlNjJmMGI0MWMiLCJwIjoiaiJ9

### DIFF
--- a/backend/app/controllers/api/v1/invitations_controller.rb
+++ b/backend/app/controllers/api/v1/invitations_controller.rb
@@ -11,10 +11,12 @@ module API
         response = params[:emails].to_a.each_with_object({}) do |email, statuses|
           next statuses[email] = 422 if email.match(Devise.email_regexp).blank?
 
-          user = User.find_by_email(email) || User.new(email: email)
+          user = User.find_by_email(email) || User.new(email: email, ui_language: current_user.locale)
           next statuses[email] = 409 if user.account_id.present?
 
-          user.invite! current_user
+          I18n.with_locale(current_user.locale) do
+            user.invite! current_user
+          end
           statuses[email] = 200
         end
         render json: response.to_json
--- a/backend/app/models/admin.rb
+++ b/backend/app/models/admin.rb
@@ -16,6 +16,10 @@ class Admin < ApplicationRecord
     Arel.sql("CONCAT_WS(' ', admins.first_name, admins.last_name)")
   end
 
+  def locale
+    ui_language
+  end
+
   def full_name
     "#{first_name} #{last_name}"
   end
--- a/backend/spec/fixtures/snapshots/api/v1/invitation-accepted.json
+++ b/backend/spec/fixtures/snapshots/api/v1/invitation-accepted.json
@@ -17,8 +17,8 @@
         "medium": null,
         "original": null
       },
-      "ui_language": "en",
-      "account_language": "en"
+      "ui_language": "es",
+      "account_language": "pt"
     }
   }
 }
--- a/backend/spec/requests/api/v1/invitation_spec.rb
+++ b/backend/spec/requests/api/v1/invitation_spec.rb
@@ -1,8 +1,8 @@
 require "swagger_helper"
 
 RSpec.describe "API V1 Invitation", type: :request do
-  let(:account) { create :account }
-  let(:user) { create :user }
+  let(:account) { create :account, language: "pt" }
+  let(:user) { create :user, ui_language: "es" }
 
   path "/api/v1/invitation/info" do
     post "Invited User information" do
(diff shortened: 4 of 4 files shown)


martintomas (commented):
Yeah, why not, even though I am not big fan of these magic variables such as `@locale`. On other hand, I guess sooner or later we will switch to `deliver_later` approach and this will stop working (https://github.com/heartcombo/devise#activejob-integration). Also this will not solve issue with other devise emails (for example confirmation). Maybe It would be worthy to customize devise mailer and do localization there (something similar to https://stackoverflow.com/questions/9311956/devise-and-localized-own-mail-template)

martintomas:
Maybe better tutorial how to do it -- https://www.rubyinrails.com/2019/09/16/override-devise-mailer-in-rails/. Methods there can be used theoretically by `User` or `Admin` so to make it workable for both options:
```
# User
def locale
 account_language.presence ||ui_language
end

# Admin
def locale
  ui_language
end

# and wrap with
I18n.with_locale record.locale do
 ....
end
```

martintomas:
Hmm and I have just realized that if you are new user which is invited, you have no ui language set up yet 🤔 . Maybe we could add language here: `user = User.find_by_email(email) || User.new(email: email, ui_language: current_user.locale)` so new users inherit ui_language of user which invited them by default....

tsubik:
Wait, devise is not using `deliver_later` by default?

martintomas:
> Wait, devise is not using `deliver_later` by default?

I think it does not use it by default
```
def send_devise_notification(notification, *args)
  message = devise_mailer.send(notification, self, *args)
  # Remove once we move to Rails 4.2+ only.
  if message.respond_to?(:deliver_now)
    message.deliver_now
  else
    message.deliver
  end
end
```

agnessa:
> , ui_language: current_user.locale

Now that I think about this, we will probably struggle with this when it turns out that only some emails need to go in account language and some in ui language. E.g. invitation email - makes sense for it to be in the account language. But e.g. "you were removed" email, I guess if at that point you have your preferred ui language set, why would the mailer not use it. Happy to cross that bridge when we come to it.

> > Hmm and I have just realized that if you are new user which is invited, you have no ui language set up yet 🤔 . Maybe we could add language here: `user = User.find_by_email(email) || User.new(email: email, ui_language: current_user.locale)` so new users inherit ui_language of user which invited them by default....
> 
> Yes, i.e. they will inherit the account language or ui language of the user who invited them, that should work

OK so actually if we do this like that, we can do it the other way round in the User model, i.e. `ui_language.presence || account_language` - because in such case a user who self registered and already set their ui language will get the email in the language they selected, whereas the user who was invited by account owner will get the email in the language of the account, and have that language set as their ui language. Right?

martintomas:
> > , ui_language: current_user.locale
> 
> Now that I think about this, we will probably struggle with this when it turns out that only some emails need to go in account language and some in ui language. E.g. invitation email - makes sense for it to be in the account language. But e.g. "you were removed" email, I guess if at that point you have your preferred ui language set, why would the mailer not use it. Happy to cross that bridge when we come to it.
> 
> > > Hmm and I have just realized that if you are new user which is invited, you have no ui language set up yet 🤔 . Maybe we could add language here: `user = User.find_by_email(email) || User.new(email: email, ui_language: current_user.locale)` so new users inherit ui_language of user which invited them by default....
> > 
> > 
> > Yes, i.e. they will inherit the account language or ui language of the user who invited them, that should work
> 
> OK so actually if we do this like that, we can do it the other way round in the User model, i.e. `ui_language.presence || account_language` - because in such case a user who self registered and already set their ui language will get the email in the language they selected, whereas the user who was invited by account owner will get the email in the language of the account, and have that language set as their ui language. Right?

Yes, it should be quite easy to switch based on our preference so we can switch logic anytime. It is also possible to tweak every email individually at future, but tbh I have no idea where `ui_language` or `account_language` has higher prio. Just hoping that they will be same the most of the time :D

agnessa:
Ah, after updating from develop branch I see we eventually went different ways about which is the default user locale, so now the tests are failing. I figured if we set a new user's ui_language to account language when inviting them, we can then use ui_language when present to localise the emails for all users, i.e. for those who self registered that's their preferred language, whereas those who were invited had this assigned from the account so have no other choice until they change their settings. Maybe I'll ask about this so we have clarity which way it should be.

agnessa:
Or maybe I'll just change it to match, this is getting more complicated than it needs to be.

martintomas:
> Or maybe I'll just change it to match, this is getting more complicated than it needs to be.

Ouch, sry for this, I have actually merged my stuff before I discovered that you have switched it 🙏

martintomas:
> > Or maybe I'll just change it to match, this is getting more complicated than it needs to be.
> 
> Ouch, sry for this, I have actually merged my stuff before I discovered that you have switched it 🙏

Whatever, I still remember which tests I have added, so I can fix them if you want, could save some of your time...

martintomas (approved):

martintomas:
Looks good for now ;) Do you think @agnessa that you could document (create tickets) for two topics which were mentioned here?
1. Devise should probably use `deliver_later`
2. We should localise all devise emails (confirmation, reset_password, etc.)

tsubik (approved):

PR #411, opened 2022-07-21
feat: API endpoint for changing password

martintomas:
I am adding new API endpoint `POST /api/v1/user/change_password` which can be used to change password of currently signed in user.

I have double checked endpoint for password reset which was added some time ago by @tsubik and it should still be good enough for our needs --> I have just added `password_confirmation` param.

Password change and password reset are two different operations so we have two different endpoint for them. First one requires user to be logged in and current user password, second one does not require user to be logged in, but reset password token needs to be provided instead.

## Testing instructions

Via Rswag documentation

## Tracking

https://vizzuality.atlassian.net/browse/LET-764

### DIFF
--- a/backend/app/controllers/api/v1/reset_passwords_controller.rb
+++ b/backend/app/controllers/api/v1/reset_passwords_controller.rb
@@ -23,7 +23,7 @@ module API
       end
 
       def update_params
-        params.permit(:reset_password_token, :password)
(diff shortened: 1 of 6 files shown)


agnessa (approved):
I ran out of time to check properly, but yes, separate endpoint makes sense.

PR #415, opened 2022-07-22
fix: Approve/Reject action at backoffice uses account language

martintomas:
Fixing of backoffice bug which blocked successful approving/rejecting of record which had empty string at attribute that is mandatory, but fallback string for record default language existed at same time. Approving/rejecting of Investor/PD should be done with respect to `language` so validation error is not raised in such cases.

**EDIT:** Further investigation uncovered a bigger problem where translated attributes always fallback to `I18n.default_locale` which is wrong. I have added fix which makes sure that fallback is always done to record `language`. This ensures that validations are always run for correct language and also correct strings are returned via API.

I am also removing `language` from Investor and PD tables, because Account already contains `language` and I am taking its value from there --> I have discovered several weird situations where Account and Investor/PD language were different which should never happen.

## Testing instructions

Create Investor or PD which have required attribute empty at language which is not selected language of record. For example:
```
I18n.with_locale "pt" do
   unapproved_investor.account.update! language: "pt", about_pt: "ABOUT PT", about_en: ""
end
```
You should be able to approve/reject such record when Language of Backoffice is set up to `EN` <-- `about_en` is empty, but fallback data (`about_pt`) exists, therefore no validation error should be raised.

## Tracking

BugFix: https://vizzuality.atlassian.net/browse/LET-806


tsubik:
Hmm, I don't see that `about` is mandatory. We have other mandatory column like `mission` (BTW. I found a bug just looking at schema.rb - mission is text for project developers and string for investors, I think that should be text for all)

tsubik:
I'm not sure if the change should not be done directly in validations in the model.

martintomas:
> Hmm, I don't see that `about` is mandatory. We have other mandatory column like `mission` (BTW. I found a bug just looking at schema.rb - mission is text for project developers and string for investors, I think that should be text for all)

Not sure about `mission` now, but we could for sure unify it, but probably as not part of this ticket ;). When I check `Account` model, there is `about` as mandatory (https://github.com/Vizzuality/heco-invest/blob/develop/backend/app/models/account.rb#L23). 

About your second note, hmm, I was thinking about it a bit and I agree with you that it makes more sense to have validations based on language of record. There is on other hand couple of issues with this approach which we would need to address:
 - traco fallbacks to I18n default locale, but it should probably fallback to `language` which ensured there to be present
 - fields depends on `presence` validation at backoffice to show if field is required or not, this would be need to be tweaked so this dynamic is taken into account
 - when changing language, you will need to provide all translation for beforehand --> we will need to add new error

backend/db/migrate/20220725074544_remove_language_from_investors_and_p_ds.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveLanguageFromInvestorsAndPDs < ActiveRecord::Migration[7.0]
+  def change
```
tsubik, line 2: :+1:

backend/app/models/concerns/translatable.rb
```diff
@@ -7,6 +7,18 @@ module Translatable
     after_save :translate_attributes, unless: :skip_translation
   end
 
+  class_methods do
+    def translates(*attributes)
+      super(*attributes)
+
+      attributes.each do |attribute|
+        define_method(attribute) do |**kwargs|
```
tsubik, line 15: Great hack :+1:

backend/spec/system/backoffice/investors_spec.rb
```diff
@@ -130,12 +162,30 @@
...
+        it "can update account language" do
+          expect(page).to have_select(t("simple_form.labels.account.language"), selected: "English")
+          select "Spanish", from: t("simple_form.labels.account.language")
+          click_on t("backoffice.common.save")
+          expect(page).to have_text(t("backoffice.messages.success_update", model: t("backoffice.common.investor")))
+          expect(approved_investor.account.reload.language).to eq("es")
+        end
+      end
+
+      context "when new language is missing some translations" do
```
tsubik, line 180: Great to have it and those missing errors in the form in case someone will do that.

backend/app/views/backoffice/accounts/_content_language_form.html.erb
```diff
@@ -1,5 +1,8 @@
 <%= simple_form_for [:backoffice, form_object] do |f| %>
   <%= f.error_notification %>
+  <% f.object.errors.each do |error| %>
```
tsubik, line 3: image
Maybe we need to refactor how errors look in backoffice, but not in this PR.

tsubik (approved):
Great work :+1:

backend/app/views/backoffice/accounts/_content_language_form.html.erb
martintomas: :) yes, these error messages could definitely look better

PR #417, opened 2022-07-22
Let 610 fe project developer and investor account owners can remove users from the account

barbara-chaves:
This PR add the possibility fot the account owners to delete the account users through the page dashboard/users

## Description

This will be done from  “My area“ (Dashboard)

The modal should make use of the existing confirmation prompt

[LET-611: FE - Existing confirmation prompt can also be used as a confirmation prompt for deletionsIN PROGRESS](https://vizzuality.atlassian.net/browse/LET-611) 

Respects the Figma designs [HeCo - Work in progress](https://www.figma.com/file/KC49uMR9t1YARmFR0Sr6aN/HeCo---Work-in-progress?node-id=2695%3A33430) 

## Testing instructions

1. Sign-in as a PD or Investor account owner
2. Go to dashboard/users (my area) page
3. There should be a table with the account users, and the last column should display the DELETE icon. Click on the icon
4. A modal with a confirmation prompt should be appear. 
  - Clicking 'cancel' button should just dismiss the modal.
  - Clicking 'delete' button should trigger a loading icon on the same button. 
  - If the action succeeds, the modal should be dismissed and the users table updated removing the deleted user from the list. 
  - If the action fails an error message should be displayed on the modal and the 'delete' text will change to 'try again'

Notes

The confirmation prompt may need some tweaks to be made us from

## Tracking

[LET-610](https://vizzuality.atlassian.net/browse/LET-610)

### DIFF
--- a/frontend/components/confirmation-prompt/component.tsx
+++ b/frontend/components/confirmation-prompt/component.tsx
@@ -4,8 +4,10 @@ import { FormattedMessage } from 'react-intl';
 
 import classnames from 'classnames';
 
+import Alert from 'components/alert';
 import Button from 'components/button';
 import Icon from 'components/icon';
+import Loading from 'components/loading';
 import Modal from 'components/modal';
 
 import type { ConfirmationPromptProps } from './types';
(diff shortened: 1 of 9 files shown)


SARodrigues (approved):
Working great! 

Deleting users that haven't accepting their invitation doesn't work, but I did see the comment on [JIRA](https://vizzuality.atlassian.net/browse/LET-610?focusedCommentId=17942). It should work when the BE makes their changes.

PR #421, opened 2022-07-25
Fix PD, Investor, Project editing updating the data in the wrong language

SARodrigues:
## Description

This PR adds the possibility of editing ProjectDeveloper and Investor account details in the account's original language. In addition, it fixes a bug where when editing projects, the `locale=` query param was still being sent, causing the endpoint to update the project in the UI language instead of the project one. 

&nbsp; 

**In this PR:**  

- Removed `useCurrentProjectDeveloper` from `projectDeveloperService`  
  _In favor of a `useProjectDeveloper`_ in `services/accounts`, to match the location of the account's `useInvestor` and `useProject`  

- Refactored several services/hooks to be able to support a `params` parameter:   
  _Where we can specify `includes`, but more importantly a `locale`. This allows us to set a `locale=null` where necessary, to retrieve/create/update data in the account/project original language. Updated the app where these services/hooks were used._
  - Account project developers
    - `getProjectDeveloper`
    - `useProjectDeveloper`
    - `updateProjectDeveloper`
    - `useUpdateProjectDeveloper`
  - Account investors  
    - `getInvestor`
    - `useInvestor`
    - `updateInvestor`
    - `useUpdateInvestor`
  - Account projects
    - `createProject`
    - `useCreateProject`
    - `updateProject`
    - `useUpdateProject`
    - `useAccount`

&nbsp;

- `projects/new`  
  Added a `locale: null` query param so that the project is created in the account's original language  

- `project/[id]:edit`  
  Added a `locale: null` query param so that the project is updated in the account's original language  

- `project-developers/edit`  
  Added a `locale: null` query param so that the project-developer is edited the account's original language  

- `investors/edit`  
  Added a `locale: null` query param so that the investor is edited the account's original language  

- `ProjectForm`  
  Removed the `useUpdateProject` in favour of the one passed via prop  
  _To match the logic in all other views/forms_   

&nbsp; 

**Others**  
  - Minor query invalidation improvements and cleanup  

&nbsp;

## Testing instructions

&nbsp;

### Using an Investor account

**Updating the profile does not send a `locale=` param**

1. Sign in into the app  
2. Visit Dashboard / Account Information
3. Set the UI language to a language different than the account one  
4. Edit your profile and submit  

Verify that the `PUT` request does not have a `locale=` query param set

&nbsp; 

### Using a Project Developer  account  

**Updating the profile does not send a `locale=` param**

1. Sign in into the app  
2. Visit Dashboard / Account Information
3. Set the UI language to a language different than the account one  
4. Edit your profile and submit  

Verify that the `PUT` request does not have a `locale=` query param set

&nbsp;

**Creating a project does not send a `locale=` param**  

1. Sign in into the app  
2. Visit Dashboard / Projects  
3. Set the UI language to a language different than the account one  
4. Create a project and submit  

Verify that the `POST` request does not have a `locale=` query param set

**Updating a project does not send a `locale=` param**  

1. Sign in into the app  
2. Visit Dashboard / Projects  
3. Set the UI language to a language different than the account one  
4. Edit a project and submit  

Verify that the `PUT` request does not have a `locale=` query param set

&nbsp;

## Tracking

[LET-811](https://vizzuality.atlassian.net/browse/LET-811)

### DIFF
--- a/frontend/pages/projects/new.tsx
+++ b/frontend/pages/projects/new.tsx
@@ -1,3 +1,5 @@
+import { useState } from 'react';
+
 import { useIntl } from 'react-intl';
 
 import { useRouter } from 'next/router';
@@ -11,7 +13,7 @@ import { useQueryReturnPath } from 'helpers/pages';
 
 import ProjectForm from 'containers/project-form';
 
-import { Paths, UserRoles } from 'enums';
+import { Languages, Paths, UserRoles } from 'enums';
 import FormPageLayout from 'layouts/form-page';
 import ProtectedPage from 'layouts/protected-page';
 import { PageComponent } from 'types';
@@ -19,11 +21,6 @@ import { GroupedEnums } from 'types/enums';
 
 import { useCreateProject } from 'services/account';
 import { getEnums } from 'services/enums/enumService';
-
-const PROJECT_QUERY_PARAMS = {
-  // We set the `locale` as `null` so that we get the project in the account's language instead of the UI language
-  locale: null,
-};
 
 export const getServerSideProps = withLocalizedRequests(async ({ locale }) => {
   const enums = await getEnums();
@@ -49,10 +46,21 @@ const CreateProject: PageComponent<CreateProjectProps> = ({ enums }) => {
   const handleOnComplete = () => {
     router.push(queryReturnPath || Paths.DashboardProjects);
   };
-  const createProject = useCreateProject(PROJECT_QUERY_PARAMS);
+  const [pdLanguage, setPdLanguage] = useState<Languages>();
+  const createProject = useCreateProject({ locale: pdLanguage });
 
   return (
-    <ProtectedPage permissions={[UserRoles.ProjectDeveloper]}>
+    <ProtectedPage
+      ownership={{
+        //The protected page already fetches the user's account, so we don't need to fetch it again. The userAccount is the project-developer. Whe are using the userAccount to determine the project-developer's language.
+        getIsOwner(_user, userAccount) {
+          setPdLanguage(userAccount.language);
+          return true;
+        },
+        allowOwner: false,
+      }}
+      permissions={[UserRoles.ProjectDeveloper]}
+    >
       <ProjectForm
         title={formatMessage({ defaultMessage: 'Create project', id: 'VUN1K7' })}
         leaveMessage={formatMessage({
--- a/frontend/pages/projects/new.tsx
+++ b/frontend/pages/projects/new.tsx
@@ -20,6 +20,11 @@ import { GroupedEnums } from 'types/enums';
 import { useCreateProject } from 'services/account';
 import { getEnums } from 'services/enums/enumService';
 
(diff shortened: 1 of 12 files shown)


frontend/pages/projects/new.tsx
```diff
@@ -34,7 +39,7 @@ type CreateProjectProps = {
 
 const CreateProject: PageComponent<CreateProjectProps> = ({ enums }) => {
   const { formatMessage } = useIntl();
-  const createProject = useCreateProject();
+  const createProject = useCreateProject(PROJECT_QUERY_PARAMS);
 
   return (
     <ProtectedPage permissions={[UserRoles.ProjectDeveloper]}>
```
barbara-chaves, line 45:
```suggestion
    <ProtectedPage
      ownership={{
        //The protected page already fetches the user's account, so we don't need to fetch it again. The userAccount is the project-developer. Whe are using the userAccount to determine the project-developer's language.
        getIsOwner(_user, userAccount) {
          setPdLanguage(userAccount.language);
          return true;
        },
        allowOwner: false,
      }}
      permissions={[UserRoles.ProjectDeveloper]}
    >
```

frontend/pages/projects/new.tsx
```diff
@@ -34,7 +39,7 @@ type CreateProjectProps = {
 
 const CreateProject: PageComponent<CreateProjectProps> = ({ enums }) => {
   const { formatMessage } = useIntl();
-  const createProject = useCreateProject();
+  const createProject = useCreateProject(PROJECT_QUERY_PARAMS);
```
barbara-chaves, line 42:
```suggestion
 const [pdLanguage, setPdLanguage] = useState<Languages>();

  const createProject = useCreateProject({ locale: pdLanguage });
```

frontend/pages/projects/new.tsx
```diff
@@ -17,6 +17,11 @@ import { GroupedEnums } from 'types/enums';
 import { useCreateProject } from 'services/account';
 import { getEnums } from 'services/enums/enumService';
 
+const PROJECT_QUERY_PARAMS = {
+  // We set the `locale` as `null` so that we get the project in the account's language instead of the UI language
+  locale: null,
+};
+
```
barbara-chaves, line 24:
```suggestion

```

barbara-chaves: If we create the project without locale it creates a bug. The locale should be the project developer’s language. 

UPDATE: I asked Martin Tomas and he confirmed that we should send the locale with the project developer’s language

barbara-chaves (changes requested):
The only change needed is adding the project developer language as new project's locale param

SARodrigues:
> The only change needed is adding the project developer language as new project's locale param

Added it both on the project's creation and edition. 
The project now does seem to be created under the correct language in the backoffice, but it looks like the project only gets translated on update, which I think would be a backend bug. (or it takes a bit to get translated).

barbara-chaves (approved):

PR #425, opened 2022-07-27
feat: All endpoints which returns list of favourites

martintomas:
I am adding new API endpoints which returns list of favourites for all entities.
```
GET /api/v1/account/investors/favourites
GET /api/v1/account/project_developers/favourites
GET /api/v1/account/projects/favourites
GET /api/v1/account/open_calls/favourites
```
These endpoints are available only for users with approved account. I have decided to keep endpoints separate (one per entity type) so It follows same approach as rest of our API. 

It took me some time till I have decided if these new endpoints should be part of for example `Account/ProjectsController/favourites` or if it would be better to have them at `FavouriteProjectsController/index` action. In second case, path would be `GET /api/v1/projects/favourite_projects`. I have tried to implement both approaches and first one which I have chosen in the end looks better, code is cleaner.

## Testing instructions

Via Rswag

## Tracking

https://vizzuality.atlassian.net/browse/LET-765

### DIFF
--- a/backend/app/controllers/api/v1/accounts/investors_controller.rb
+++ b/backend/app/controllers/api/v1/accounts/investors_controller.rb
@@ -2,8 +2,11 @@ module API
   module V1
     module Accounts
       class InvestorsController < BaseController
-        before_action :require_investor!, except: :create
+        include API::Pagination
+
+        before_action :require_investor!, except: %i[create favourites]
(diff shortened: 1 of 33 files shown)


tsubik (approved):
Yes, I think the first approach is indeed better :+1:

PR #429, opened 2022-07-27
Create coming soon page

barbara-chaves:
This PR adds the coming-soon page
Screenshot 2022-07-27 at 16 30 34

## Tracking

[LET-819](https://vizzuality.atlassian.net/browse/LET-819)

### DIFF
--- a/frontend/pages/coming-soon.tsx
+++ b/frontend/pages/coming-soon.tsx
@@ -1,0 +1,48 @@
+import { FormattedMessage, useIntl } from 'react-intl';
+
+import { withLocalizedRequests } from 'hoc/locale';
+
+import { InferGetStaticPropsType } from 'next';
+
+import { loadI18nMessages } from 'helpers/i18n';
+
+import Head from 'components/head';
+import LayoutContainer from 'components/layout-container';
+import { StaticPageLayoutProps } from 'layouts/static-page';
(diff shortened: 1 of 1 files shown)


SARodrigues:
Looks good, but shouldn't it be a layout or component we can reuse to create an upcoming, instead of a standalone page?  

The idea is that, say the user visits `/privacy-policy` (which doesn't yet exist), we'd display "Coming soon". With a layout or component, the page would exist at the URL, but display "Coming soon". 

With your implementation of a standard page, I guess we could still make it work via `rewrites` in `next.config.js`, but feels a bit weird. We'd also have a `/coming-soon` page which doesn't make much sense.

SARodrigues (changes requested):
Looks good, but shouldn't it be a layout or component we can reuse to create an upcoming, instead of a standalone page?  

The idea is that, say the user visits `/privacy-policy` (which doesn't yet exist), we'd display "Coming soon". With a layout or component, the page would exist at the URL, but display "Coming soon". 

With your implementation of a standard page, I guess we could still make it work via `rewrites` in `next.config.js`, but feels a bit weird. We'd also have a `/coming-soon` page which doesn't make much sense. 

**Edit:** Another issue I see, is that in the future, search engines will be indexing a `/coming-soon` page.

PR #437, opened 2022-07-28
feat: Devise uses deliver_later method

martintomas:
This PR contains two changes related to Devise :).

First of all, I am moving all locales to `zu.yml`, because these locales are pushed to transifex and we want to have these strings at transifex.

Second change is that devise mails are now send via `deliver_later` method and they use correct `locale` of user --> text of emails are based on account language or `ui_language` variable of appropriate record.

## Testing instructions

All tweaked devise mails share same logic so It should be enough to test one. I think that good candidate is for example forgotten password at backoffice. Create Admin with Spanish ui_language and use forgotten password feature --> such admin should receive email in Spanish.

## Tracking

https://vizzuality.atlassian.net/browse/LET-815
https://vizzuality.atlassian.net/browse/LET-816


backend/spec/requests/api/v1/reset_password_spec.rb
```diff
@@ -21,19 +21,18 @@
...
+          expect {
+            submit_request example.metadata
+            assert_response_matches_metadata example.metadata
+          }.to have_enqueued_mail(DeviseMailer, :reset_password_instructions).with(user, String, fallback_language: :en)
         end
 
         context "invalid email" do
           let(:params) { {email: "invalid"} }
 
+          run_test!
```
tsubik, line 34: I'm sometimes confused when this should be placed ;]

backend/spec/mailers/devise_mailer_spec.rb
```diff
@@ -0,0 +1,48 @@
+require "rails_helper"
+
+RSpec.describe DeviseMailer, type: :mailer do
```
tsubik, line 3: Very meticulous to have all those mailers tested :)

backend/spec/requests/api/v1/reset_password_spec.rb
martintomas: When you look into this method, it is in reality just shortcut, which runs:
```
submit_request example.metadata
assert_response_matches_metadata example.metadata
```
as you have already seen, sometimes I run these commands manually so I can wrap them for example with `have_enqueued_mail`, but If you don't trigger them on your own, you need to use `run_test!` so you get API response.

backend/app/mailers/devise_mailer.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    I18n.with_locale(record.locale || opts[:fallback_language]) do
+      devise_mail(record, :reset_password_instructions, opts)
+    end
+  end
+
+  def invitation_instructions(record, token, opts = {})
+    @token = token
+
+    I18n.with_locale(record.locale || opts[:fallback_language]) do
```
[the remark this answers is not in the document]
martintomas: Unfortunately, this does not work. Your idea is correct, but It would work only if we have used `deliver_now` at `send_devise_notification` so email is send immediately and takes into account `I18n.with_locale`. I have switched to `deliver_later` which pushes mail to queue and when it gets picked up by different process/server, It does not remember what was original locale before it got enqueued --> that is reason why localization needs to be put directly to mailer.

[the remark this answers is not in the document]
tsubik: Ahh, ok, right.

tsubik (approved):
Looks great :+1:

PR #440, opened 2022-07-29
fix: Adding correct FE link to reset password email

martintomas:
I am adding correct link to reset password email, which is `"#{ENV["FRONTEND_URL"]}/sign-in/reset-password?reset_password_token=#{@token}"`.

I have also seen that `user` devise emails (reset password and confirmation) have hardcoded text which is not at transifex yet, so I am adding it.

## Testing instructions

Via Rswag -- double check that reset password email contains new link to FE.

## Tracking

https://vizzuality.atlassian.net/browse/LET-677


tsubik (approved):
:+1:

PR #447, opened 2022-08-02
Fix/transfer ownership of account invitations

tsubik:
Transferring ownership of invitations along with account ownership.

## Testing instructions

Invite user to account, change ownership, check if invited_by_id changed to new owner id

## Tracking

https://vizzuality.atlassian.net/browse/LET-849


tsubik:
I also tested it locally with Barbara branch, which adds transfer ownership to the frontend, just had to change GET to POST (as I don't like testing with api docs lol).

agnessa (approved):

PR #448, opened 2022-08-02
Allow serving images from localhost

tsubik:
I wanted to test account ownership transfer locally with localhost backend as I was lazy to use API docs for doing that (wanted Buttons! :D). Unfortunately, that did not work as images served by localhost were blocked, this fixes that, not sure if there are any negative implications of changing that config or not (never used it).

## Testing instructions

What to test? How to do it?

## Tracking

No tracking


SARodrigues (approved):
All good!

PR #455, opened 2022-08-04
New instance variable to differentiate staging from production

agnessa:
We're using "production" rails env in both staging and production, which is simpler than having 2 environments, but in some cases it's useful to know from the application which it is and this is now available as an env var `INSTANCE_ROLE`.

## Tracking

https://vizzuality.atlassian.net/browse/LET-426?atlOrigin=eyJpIjoiYmE1NGVkYWM4NjUzNDY0YzgxNmM1OWZlYWQ5NTcxMGMiLCJwIjoiaiJ9

### DIFF
--- a/backend/config/environments/production.rb
+++ b/backend/config/environments/production.rb
@@ -16,7 +16,7 @@ Rails.application.configure do
   config.consider_all_requests_local = false
   config.action_controller.perform_caching = true
   # we want to allow non https localhost requests to staging, but not production
-  config.action_controller.forgery_protection_origin_check = (ENV["INSTANCE_ROLE"].present? && ENV["INSTANCE_ROLE"] == "production") || false
+  config.action_controller.forgery_protection_origin_check = false if ENV["INSTANCE_ROLE"] == "staging"
 
   # Ensures that a master key has been made available in either ENV["RAILS_MASTER_KEY"]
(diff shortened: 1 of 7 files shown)


backend/config/environments/production.rb
```diff
@@ -15,8 +15,8 @@
   # Full error reports are disabled and caching is turned on.
   config.consider_all_requests_local = false
   config.action_controller.perform_caching = true
-  # TODO: fix this later, directupload is checkin it on staging but we want to allow non https localhost requests
-  config.action_controller.forgery_protection_origin_check = false
+  # we want to allow non https localhost requests to staging, but not production
+  config.action_controller.forgery_protection_origin_check = (ENV["INSTANCE_ROLE"].present? && ENV["INSTANCE_ROLE"] == "production") || false
```
tsubik, line 19: hmm, it's ok, but maybe a little less readable than just setting this as false if `ENV["INSTANCE_ROLE"] == "staging"`. The default value for that setting is true, and I bet it won't change in future rails versions.

tsubik (approved):

backend/config/environments/production.rb
agnessa: right, I'll change that, thanks!

PR #458, opened 2022-08-05
Fix filters being cleared when using the discover pages' pagination

SARodrigues:
## Description

This PR aims to fix a couple issues with the Discover/Search pages:  
- Using the pagination clears the filters  
- Pagination does not reset to the first page when changing tabs  

Extra:  
- Clearing the search term from the input clears it from the URL  
- Clearing the filters clears them from the URL  
- When page is `1`, it'll be cleared from the URL  
- Fix discover/projects list not displaying the loading spinner

Not done here:  
- Sorting does not reset when changing tabs  
  _eg: while searching for projects, if the user is sorting by impact score, then clicks on another tab, sorting does not get reset_

## Testing instructions

Verify that:  
1. Using the pagination does not clear enabled filters   
    - Visit `discover/projects`
    - Select the `Biodiversity` filter
    - Click on another tab
    - Verify that the filter does not get cleared
2. When changing tabs, pagination gets reset to the first page  
    - Visit `discover/projects`  
    - Click to move to page 3  
    - Click on another tab  
    - Verify that the currently selected page is page 1  

## Tracking

N/A

### DIFF
--- a/frontend/containers/layouts/discover-search/component.tsx
+++ b/frontend/containers/layouts/discover-search/component.tsx
@@ -7,7 +7,7 @@ import cx from 'classnames';
 
 import { useRouter } from 'next/router';
 
-import { useDiscoverPath, useQueryParams } from 'helpers/pages';
+import { cleanQueryParams, useDiscoverPath, useQueryParams } from 'helpers/pages';
 
 import Filters from 'containers/forms/filters';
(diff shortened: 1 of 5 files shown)


SARodrigues:
Note to self: Only sorting is missing.

clementprdhomme (approved):

PR #460, opened 2022-08-08
feat: API for Open Call creation

martintomas:
This PR refactors Open Call model so it contains all required fields mentioned at MVP requirements:
Screenshot 2022-08-08 at 11 03 17

It also adds new API endpoint which allows investors to create new OpenCall from these required data.

## Testing instructions

New endpoint should be available at Rswag

## Tracking

https://vizzuality.atlassian.net/browse/LET-683


agnessa (approved):
Looks good, thanks!

PR #463, opened 2022-08-09
feat: API for Open Call update

martintomas:
New API endpoint which allows to update existing Open Calls ;)

## Testing instructions

Via Rswag

## Tracking

https://vizzuality.atlassian.net/browse/LET-684

### DIFF
--- a/backend/db/migrate/20220808082718_changes_for_open_calls.rb
+++ b/backend/db/migrate/20220808082718_changes_for_open_calls.rb
@@ -28,10 +28,13 @@ class ChangesForOpenCalls < ActiveRecord::Migration[7.0]
   private
 
   def migrate_existing_open_calls!
-    country = Location.find_by! location_type: :country, name_en: "Colombia"
     OpenCall.all.each do |open_call|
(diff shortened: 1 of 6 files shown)


backend/db/migrate/20220808082718_changes_for_open_calls.rb
```diff
@@ -28,10 +28,13 @@ def change
...
   def migrate_existing_open_calls!
-    country = Location.find_by! location_type: :country, name_en: "Colombia"
     OpenCall.all.each do |open_call|
       open_call.assign_attributes country_id: country.id, instrument_types: [open_call.instrument_type]
       open_call.save validate: false
     end
   end
+
+  def country
+    @country ||= Location.find_by! location_type: :country, name_en: "Colombia"
```
martintomas, line 38: just slight tweak of migration --> deployment runs migrations and it fails on empty db which does not have `Colombia` location. This fix will ask for Colombia location only when there are already some Open Calls.

agnessa (approved):

PR #464, opened 2022-08-09
feat: API for deletion of all user favourites

martintomas:
I am adding new API endpoint `DELETE /api/v1/account/users/favourites` which is available to all approved users and which deletes all user's favourites.

Feel free to let me know if you feel that path should look different, I was not 100% sure if putting this under `DELETE` action was good choice.

## Testing instructions

Via Rswag

## Tracking

https://vizzuality.atlassian.net/browse/LET-858

### DIFF
--- a/backend/app/controllers/api/v1/accounts/users_controller.rb
+++ b/backend/app/controllers/api/v1/accounts/users_controller.rb
@@ -31,10 +31,12 @@ module API
 
         # DELETE
         def favourites
-          current_user.favourite_projects.destroy_all
-          current_user.favourite_project_developers.destroy_all
-          current_user.favourite_investors.destroy_all
-          current_user.favourite_open_calls.destroy_all
+          ActiveRecord::Base.transaction do
+            current_user.favourite_projects.destroy_all
+            current_user.favourite_project_developers.destroy_all
+            current_user.favourite_investors.destroy_all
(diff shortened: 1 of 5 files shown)


agnessa:
we need to double check if the intention is to remove all favourites of all sorts, or "remove all projects" separate to "remove all investors"

martintomas:
> we need to double check if the intention is to remove all favourites of all sorts, or "remove all projects" separate to "remove all investors"

Okay, both options are feasible so lets wait for feedback from Su

agnessa:
Susana confirmed this is about removing all sorts of favourites, so let's merge

agnessa (approved):

backend/app/controllers/api/v1/accounts/users_controller.rb
```diff
@@ -29,6 +29,15 @@ def transfer_ownership
           render json: UserSerializer.new(user, params: {current_user: user}).serializable_hash
         end
 
+        # DELETE
+        def favourites
+          current_user.favourite_projects.destroy_all
```
tsubik, line 34: Not better in transaction? Just in `ActiveRecord::Base.transaction`

martintomas: Hmm, yes, it is, I think nothing can blow up here, but just to be on save side ;)

PR #466, opened 2022-08-10
fix: Account can be deleted via Backoffice

martintomas:
I am fixing Account deletion action at backoffice. It now deletes even `Account` record which triggers users deletion as well -- at past, just `Investor` or `ProjectDeveloper` record were destroyed, but `Account` didn't get deleted.

## Testing instructions

Delete Account (PD or Investor) via Backoffice and check that all users are destroyed as well

## Tracking

BugFix

### DIFF
--- a/backend/app/models/user.rb
+++ b/backend/app/models/user.rb
@@ -13,7 +13,7 @@ class User < ApplicationRecord
   has_many :investors, through: :favourite_investors
   has_many :favourite_open_calls, dependent: :destroy
   has_many :open_calls, through: :favourite_open_calls
-  has_one :owner_account, class_name: "Account", foreign_key: "owner_id", dependent: :restrict_with_error
+  has_one :owner_account, class_name: "Account", foreign_key: "owner_id"
 
   has_one_attached :avatar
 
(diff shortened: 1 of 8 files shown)


backend/app/models/user.rb
```diff
@@ -13,7 +13,7 @@ class User < ApplicationRecord
   has_many :investors, through: :favourite_investors
   has_many :favourite_open_calls, dependent: :destroy
   has_many :open_calls, through: :favourite_open_calls
-  has_one :owner_account, class_name: "Account", foreign_key: "owner_id", dependent: :restrict_with_error
+  has_one :owner_account, class_name: "Account", foreign_key: "owner_id"
```
martintomas, line 16: I had to remove `dependent: :restrict_with_error`, because it prevented proper account deletion, but Backoffice disables button for user deletion if user is account owner and just to be save, I have put also
```
return redirect_to(backoffice_users_path, status: :see_other, alert: I18n.t("errors.messages.user.is_owner")) if @user.owner_account.present?
```
to User controller.

tsubik: Are we safe on the API side of user removal?

tsubik: I understand we restrict removing user who is account owner, right?

martintomas: Yes, we are safe on API side, it is handled by cancancan and there is even test for this use case

tsubik (approved):
:+1:

agnessa (approved):

PR #469, opened 2022-08-11
feat: Adding status to Open Call model

martintomas:
Adding status to Open Call model
 - status param is also added to create/edit API endpoints
 - closing/launching of OpenCall can be done via edit API endpoint <-- it is done same way as Project. Makes it easy for FE to use it, but all extra stuff related to status change needs to be handled via callbacks on model, which is a bit tricky (or some extra service in middle can do it)
 - status is added to OpenCall serialiser
 - draft OpenCalls are filtered out from public endpoint

## Testing instructions

Via Rswag now --> you can see that status can be setup via API endpoints

## Tracking

https://vizzuality.atlassian.net/browse/LET-697
https://vizzuality.atlassian.net/browse/LET-699
https://vizzuality.atlassian.net/browse/LET-700
https://vizzuality.atlassian.net/browse/LET-689
https://vizzuality.atlassian.net/browse/LET-688


agnessa (approved):
Cool, happy to keep it analogical to project status

PR #471, opened 2022-08-12
feat: Added impact_calculated attribute to Project model

martintomas:
I am adding new attribute called `impact_calculated` to Project model and its serializer which can be used by FE (@barbara-chaves ) to decide if impact already was calculated for appropriate Project.

## Testing instructions

Via Rswag -- check that Project contains correct value at `impact_calculated` attribute. The easiest approach is probably to change geometry or impact areas of already existing project --> immediately after change, this attribute is gonna be false, but changes to true after impacts are recalculated.

## Tracking

https://vizzuality.atlassian.net/browse/LET-872

### DIFF
--- a/backend/db/migrate/20220812075824_add_impact_calculated_to_projects.rb
+++ b/backend/db/migrate/20220812075824_add_impact_calculated_to_projects.rb
@@ -1,0 +1,6 @@
+class AddImpactCalculatedToProjects < ActiveRecord::Migration[7.0]
+  def change
+    add_column :projects, :impact_calculated, :boolean, default: false
+    Project.update_all impact_calculated: true # we expect that all existing projects already have impact calculated
+  end
+end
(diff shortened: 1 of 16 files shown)


backend/db/migrate/20220812075824_add_impact_calculated_to_projects.rb
```diff
@@ -0,0 +1,6 @@
+class AddImpactCalculatedToProjects < ActiveRecord::Migration[7.0]
+  def change
+    add_column :projects, :impact_calculated, :boolean, default: false
+    Project.update_all impact_calculated: true # we expect that all existing projects already have impact calculated
```
tsubik, line 4: I was about to write a comment to make this reversible, but I see that rollback works just fine with this lol.

tsubik (approved):

PR #475, opened 2022-08-15
feat: Private API endpoint for open calls

martintomas:
I am adding new private API endpoint for OpenCalls. Endpoint itself is quite simple, but what made this task more challenging is searching ability, because:
 - country/municipality/department names are localized
 - name of open call is also localized (we should be able to deal with fallbacks)
 - instrument types is array of enums
 - status is enum
 
It was required to use full text search on all these attributes, BUT instrument_types or status are not texts and names are text, but there is extra logic which is enforced on app layer (fallback). I was able to come with three possible solutions:
 - do searching at app --> just load everything to app, run it through serializer and do searching then (simple, but quite slow for big sets and not really usable for backoffice)
 - modify PgSearch gem --> spend some time reading doc, but didn't find some nice way how to do it
 - use ransack --> backoffice already uses this and it is quite easy to extend, cons is that it is gonna be probably slower then PgSearch (seq scan/filter which it does looks quite intimidating :))
 
Whatever, I have decided to go for ransack option and written extra concern which adds three possible helpers to model:
- `generate_ransackers_for_translated_columns` --> for every translated columns add ransacker option which fallbacks same way as traco at app
- `generate_localized_ransackers_for_static_types` --> adds ransackers for static types. Allows to search or sort based on user language. For arrays, it does `array_to_string`
- `generate_localized_ransackers_for_enums` --> similar as previous, allows to search/sort based on status text which is localized.

## Testing instructions

Via Rswag documentation -- new API endpoint should be part of it, feel free to check all fulltext filters

## Tracking

https://vizzuality.atlassian.net/browse/LET-940

### DIFF
--- a/backend/app/models/concerns/extra_ransackers.rb
+++ b/backend/app/models/concerns/extra_ransackers.rb
@@ -1,0 +1,41 @@
+module ExtraRansackers
+  extend ActiveSupport::Concern
+
+  class_methods do
+    def generate_ransackers_for_translated_columns(fallback = "#{table_name}.language", db_column: true)
+      fallback = db_column ? fallback : "'#{fallback}'"
+
+      translatable_attributes.each do |attribute|
+        ransacker attribute do |parent|
+          parent_table_name = parent.table.table_alias || table_name
+          fallback_sql = (I18n.available_locales - [:zu]).map { |l| "when #{fallback} = '#{l}' then #{parent_table_name}.#{attribute}_#{l}" }.join(" ")
+          Arel.sql("(case when #{parent_table_name}.#{attribute}_#{I18n.locale} <> '' then #{parent_table_name}.#{attribute}_#{I18n.locale} else (case #{fallback_sql} end) end)")
+        end
+      end
+    end
+
+    def generate_localized_ransackers_for_static_types(*types)
+      types.each do |type|
+        ransacker "#{type}_localized" do |parent|
+          parent_table_name = parent.table.table_alias || table_name
+          translated_type_sql = type.to_s.classify.constantize.all.inject("ARRAY[#{parent_table_name}.#{type}]::text[]") do |res, obj|
+            "array_replace(#{res}, '#{obj.slug}', '#{obj.name}')"
+          end
+          Arel.sql("array_to_string(#{translated_type_sql}, ' ')")
+        end
+      end
+    end
+
+    def generate_localized_ransackers_for_enums(**enums)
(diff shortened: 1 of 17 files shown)


tsubik:
Could we also do it like full_text OR contains in the rest of problematic fields?

backend/app/models/concerns/extra_ransackers.rb
```diff
@@ -0,0 +1,41 @@
+module ExtraRansackers
```
tsubik, line 1: :+1:

martintomas:
> Could we also do it like full_text OR contains in the rest of problematic fields?

No sure about question @tsubik :). Basically you can use any matchers mentioned at https://activerecord-hackery.github.io/ransack/getting-started/search-matches/ and some of them should be quite close to full text search (I have used `cont`, but we can use any of them). True full text search is handled by `PgSearch` which we already have at project, but I had problem to enhance it so it works with enum, static types or translation fallbacks. You can ofc use combination --> full text for all string values at db which are not translated (don't have fallback) and ransacker for rest.

tsubik:
I meant something like in pseudo code `OpenCall.full_text_search.or(OpenCall.ransack_contains_for_troubling_attributes_scope)`.

tsubik:
But actually, I see there is no description, only name, so maybe `cont` is good enough as all of the searched fields are short strings.

martintomas:
> I meant something like in pseudo code `OpenCall.full_text_search.or(OpenCall.ransack_contains_for_troubling_attributes_scope)`.

Aha, I get it, hmm, I know that it is possible to combine queries generated by `PgSearch ` and `ransacker`, but I had problems when joins come into play, because they use different aliases. My experience here was that it is possible, but also quite fragile and does not work for all use cases. If it would be required to search also on top of `description` (which is localized with traco), we could either use `PgSearch` and give up on fallbacks or use ransack which is for sure gonna be slower.

tsubik (approved):
:+1:

tsubik:
Yeah, when combining complicated queries it's better to just `pluck(:id)` from one query and combine with simplest `where(id: ...)`, but here it's fine how it is.

martintomas:
> Yeah, when combining complicated queries it's better to just `pluck(:id)` from one query and combine with simplest `where(id: ...)`, but here it's fine how it is.

Yes, exactly, with `pluck`, you can combine everything :D

agnessa (approved):

PR #476, opened 2022-08-16
fix: Backoffice sorting

martintomas:
I am refactoring sorting at backoffice so it uses generated ransackers added by #475 . This have advantage of respecting translation fallbacks and also it adds ransackers for all attributes (no need to add extra stuff to model everytime).

Also, sorting by language was broken so I am fixing this one --> we show static type (`en`) instead of localized text (`English`) at backoffice table so I have not used one of these new ransackers for this (`language_localized`).

## Testing instructions

Via backoffice -- simple check if sorting used there still works

## Tracking

Fixes bug https://vizzuality.atlassian.net/browse/LET-851

### DIFF
--- a/backend/spec/models/project_spec.rb
+++ b/backend/spec/models/project_spec.rb
@@ -7,6 +7,8 @@ RSpec.describe Project, type: :model do
 
   it_behaves_like :searchable
   it_behaves_like :translatable
+  it_behaves_like :with_ransacked_translations
+  it_behaves_like :with_ransacked_static_types, :category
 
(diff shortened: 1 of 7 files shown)


tsubik (approved):
:+1:

backend/spec/models/project_spec.rb
```diff
@@ -7,6 +7,8 @@
 
   it_behaves_like :searchable
   it_behaves_like :translatable
+  it_behaves_like :with_ransacked_translations
```
tsubik, line 10: :+1:

PR #478, opened 2022-08-16
feat: API for destroying open calls

martintomas:
Simple PR which adds new API endpoint for removing existing open calls which you own.

## Testing instructions

Via Rswag -- try to destroy some Open Call which you own

## Tracking

https://vizzuality.atlassian.net/browse/LET-914


tsubik (approved):
:rocket:

PR #481, opened 2022-08-17
refactoring: Use Location model instead of Mosaic enum

martintomas:
This PR merges Mosaic enums and Location data to one entity. At first, I was trying to keep Mosaic enum and find some way how to keep both entities at sync, but all solutions which I have come with were too fragile and they would break sooner or later. In the end, I have decided to completely replace Mosaic enum with Location model and refactored all code based on this. Unfortunately, there are some breaking changes for API (@barbara-chaves , @clementprdhomme).

**Breaking changes:**
 - there are no more mosaic data as part of `/api/v1/enums`
 - `location_type` filter for `/api/v1/locations` is changed from `region` to `priority_landscape`. To obtain all priority landscapes, please use this endpoint with `filter[location_type]=priority_landscape`
 - `mosaics` parameter when creating or updating project developer is replaced by `priority_landscape_ids` param
 - same is true for project developer API response, `mosaics` fields is replaced everywhere by `priority_landscapes`

There are three new migrations:
 - renaming location type from `region` to `priority_landscape`
 - removing `mosaics` attribute from `ProjectDeveloper` model
 - adding `visible` attribute to `Location`
 
There are also two new rake tasks which needs to be run after this PR is deployed:
 - `rake priority_landscapes:translate` --> adds translation for known priority landscapes
 - `rake priority_landscapes:hide` --> hides priority landscapes that we have decided to ignore for now from API 

## Testing instructions

To manually test this you should:
 - verify Backoffice --> you can change Priority Landscapes for PD and you can filter by priority landscape at project
 - check Rswag --> influenced endpoints are creation of PD and list locations filtered by `priority_landscape` type

## Tracking

https://vizzuality.atlassian.net/browse/LET-583


backend/config/locales/zu.yml
```diff
@@ -86,7 +86,7 @@ zu:
       defaults:
         categories: Select the categories that you work on
         mission: What's your mission?
-        mosaics: Select HeCo priority landscapes on which you will have impact
```
agnessa, line 89: this will have an impact on translations, right? I'm wondering if it might make sense to do it in an additive way, i.e. rather than renaming keeping 2 entries for a while, so that translations can be copied from the old one before it disappears (it's my understanding it disappears, but maybe I'm wrong about how transifex works)

martintomas: I also believe it will disappear, I can keep it for now, no problem ;)

backend/app/models/project_developer.rb
```diff
@@ -7,11 +7,12 @@ class ProjectDeveloper < ApplicationRecord
   has_many :favourite_project_developers, dependent: :destroy
   has_many :project_involvements, dependent: :destroy
   has_many :involved_projects, through: :project_involvements, source: :project, dependent: :destroy
+  has_many :project_developer_priority_landscapes, dependent: :destroy
+  has_many :priority_landscapes, through: :project_developer_priority_landscapes, source: :priority_landscape, dependent: :destroy
 
   validates :categories, array_inclusion: {in: Category::TYPES}, presence: true
   validates :impacts, array_inclusion: {in: Impact::TYPES}
   validates :project_developer_type, inclusion: {in: ProjectDeveloperType::TYPES, allow_blank: true}, presence: true
-  validates :mosaics, array_inclusion: {in: Mosaic::TYPES}
```
tsubik, line 14: I see `mosaics` collection was not required, and running migration on re-seeded database does not work.

martintomas: Ouch, I thought that it will be empty array in such case, hmm, will fix migration ;)

tsubik (approved):
:+1

PR #484, opened 2022-08-18
fix: Enforce account language for stuff created via API

martintomas:
I am fixing translation mismatch during create and update of Project/Investor/PD/OpenCall. This mismatch happened when `locale` provided by FE was different than user account language. In such case language of new record corresponded to user account language, but data were saved under `locale` suffix. This PR fixes this bug by ignoring `locale` in these cases and ensuring that only account language is taken into account.

## Testing instructions

Should be possible via Rswag --> when creating/updating Project/Investor/PD/OpenCall, set `locale` param to different value then your account language is. All attributes should be saved under account language <-- can be checked by getting record via API with locale that corresponds to your account language.
When creating Invetor/PD, you can also sent `locale` which is different than `language` param. All data should be saved under `language` param in this case.

## Tracking

https://vizzuality.atlassian.net/browse/LET-956

### DIFF
--- a/backend/app/controllers/api/v1/accounts/investors_controller.rb
+++ b/backend/app/controllers/api/v1/accounts/investors_controller.rb
@@ -5,7 +5,8 @@ module API
         include API::Pagination
 
         before_action :require_investor!, except: %i[create favourites]
-        around_action(only: [:show]) { |_controller, action| set_locale(current_user&.account&.language, &action) }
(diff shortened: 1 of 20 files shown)


agnessa (approved):
That looks good. I started wondering what we're going to do when we start dealing with open call applications, which have some translatable fields, but possibly we will be able to use the applicant's account language to keep it consistent with this approach.

PR #487, opened 2022-08-19
feat: Priority landscape filter

martintomas:
This PR adds `filter[priority_landscape]` option to `project` and `project_developer` endpoints.

## Testing instructions

Via Rswag

## Tracking

https://vizzuality.atlassian.net/browse/LET-927

### DIFF
--- a/backend/app/services/api/filterer.rb
+++ b/backend/app/services/api/filterer.rb
@@ -1,6 +1,6 @@
 module API
   class Filterer
-    attr_accessor :query, :filters, :language
+    attr_accessor :query, :original_query, :filters, :language
 
     FULL_TEXT_FILTERS = %i[name description about mission problem solution expected_impact]
     FULL_TEXT_EXTRA_TABLES = {
@@ -9,9 +9,11 @@ module API
     }
     NUMERICAL_FILTERS = %i[municipality_biodiversity_impact municipality_climate_impact municipality_water_impact municipality_community_impact]
     ENUM_FILTERS = %i[category impact sdg instrument_type ticket_size]
+    ASSOCIATION_FILTERS = %i[priority_landscape]
 
-    def initialize(query, filters, language: I18n.locale)
-      @query = query
+    def initialize(original_query, filters, language: I18n.locale)
+      @original_query = original_query
+      @query = original_query.klass.all
       @filters = filters
       @language = language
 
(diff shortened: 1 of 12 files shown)


backend/app/services/api/filterer.rb
```diff
@@ -23,7 +25,8 @@ def call
       apply_numerical_filter
       filter_by_enums
       filter_by_only_verified
-      query
+      filter_by_relations
+      original_query.where id: query.pluck(:id)
```
martintomas, line 29: I have decided to use `pluck` here, because query generated by all filters (especially when multiple filters are combined) is getting super complex. This breaks this complexity so all additional stuff inside API (sorting, preloading, etc.) is using simple query instead of super complex one.

backend/app/services/api/filterer.rb
```diff
@@ -62,6 +65,12 @@ def filter_by_only_verified
       self.query = query.where trusted: true
     end
 
+    def filter_by_relations
```
tsubik, line 68: Hmm, at least for `priority_landscape` that would not be necessary to make joins and complicate this, it could be done like filtering by attribute, `Project.where(priority_landscape: id)`

tsubik: Looking at specs I see this is used for to filter both project developers and projects, ok then discard my latest comment.

tsubik: But we have join and without distinct, would that not return duplicates when filtering by multiple priority landscapes?

martintomas: Yeah, you are right with this one, I have decided to keep code simple so it works same way for has_many (PD has many priority landscapes) and belongs_to (Project has one priority landscape) association. To be honest, I don't think it will slow down query much, but if you want, I can refactor this method so belongs to associations are handled differently.

martintomas: Nope, duplicates will be removed thanks to `pluck` (`original_query.where id: query.pluck(:id)`). My initial intention for adding pluck there was to split filter queries and rest of the stuff, but it also solves issue with duplicates

tsubik: Ok, I see that later from this query `id` is plucked. Then all good

tsubik: Anyway, tested with api docs.

tsubik (approved):
:+1:

backend/app/services/api/filterer.rb
tsubik: Lol, I did not see your comments. Seeing them just now :D haha

agnessa (approved):

PR #489, opened 2022-08-22
feat: List of the open calls at backoffice

martintomas:
I am adding list of the Open Calls to Backoffice table. Table itself contains `Applications` so I have added `OpenCallApplications` model based on its design. Searching is done fully by ransack (no more `API::Filterer`) and it takes into account only table attributes.

Screenshot 2022-08-22 at 13 57 07

## Testing instructions

List can be checked via Backoffice ;). Please make sure that csv export works, filtering, pagination, total count.

## Tracking

https://vizzuality.atlassian.net/browse/LET-901

### DIFF
--- a/backend/app/controllers/backoffice/open_calls_controller.rb
+++ b/backend/app/controllers/backoffice/open_calls_controller.rb
@@ -1,0 +1,19 @@
+module Backoffice
+  class OpenCallsController < BaseController
+    def index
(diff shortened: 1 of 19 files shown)


tsubik (approved):
:+1:

agnessa (approved):
Perfect, thanks!

PR #491, opened 2022-08-22
Allow investor users to edit their open calls

clementprdhomme:
This PR allows investor users to edit their open calls in the Dashboard.

In addition, this PR fixes an issue in the investor edition form where the previously selected SDGs wouldn't be highlighted.

## Testing instructions

- You can edit draft and published open calls
  - When creating/editing a draft open call, you don't see the screen about verification
- You can see the public page of published open calls (making sure nothing broke)
- You can create new open calls (making sure nothing broke)

## Tracking

[LET-885](https://vizzuality.atlassian.net/browse/LET-885)


SARodrigues (approved):
All I've tested works as expected!

PR #492, opened 2022-08-23
Allow users to favorite open calls

SARodrigues:
## Description

This PR adds the frontend functionality to allow a user to favourite/unfavourite open calls from both the header and CTA at the bottom of the open calls public page. 

**In this PR:**
- Added `useFavoriteOpenCall` to `open-call-service`  
- Implementation to make the _"Favorite"_ button (header) in the open call public page work  
- Implementation to make the _"Favorite"_ button (CTA, above the footer) in the open call public page work   

**Notes:**  
The endpoint hasn't been yet implemented and is tracked by [LET-989](https://vizzuality.atlassian.net/browse/LET-989), so the implementation will not work as of now. However, the logic is a copy-paste of the one used for projects, PDs, investors, and the endpoint was inferred from those as well. 

## Testing instructions

Verify that:  
- When a user is not signed in, the "favorite" buttons are disabled  
- When a user is signed in, clicking the "favorite" buttons sends a POST request to the (inferred, not yet implemented) API endpoint 

One could also pass `openCall={{...openCall, { favourite: true }}` to both `OpenCallHeader` and `OpenCallInvestorAndFooter` in `/pages/open-call/[id]/index.tsx`, to verify that the button icons get filled with a solid color when `favourite` is set to `true`

## Tracking

[LET-962](https://vizzuality.atlassian.net/browse/LET-962)

### DIFF
--- a/frontend/containers/open-call-page/header/component.tsx
+++ b/frontend/containers/open-call-page/header/component.tsx
@@ -70,14 +70,7 @@ export const OpenCallHeader: FC<OpenCallHeaderProps> = ({
 
   const handleFavoriteClick = () => {
     // This mutation uses a 'DELETE' request when the isFavorite is true, and a 'POST' request when is false.
-    favoriteOpenCall.mutate(
-      { id, isFavourite: favourite },
-      {
-        onSuccess: (data) => {
-          openCall.favourite = data.favourite;
-        },
-      }
-    );
+    favoriteOpenCall.mutate({ id, isFavourite: favourite });
   };
 
   return (
--- a/frontend/containers/open-call-page/header/component.tsx
+++ b/frontend/containers/open-call-page/header/component.tsx
@@ -9,6 +9,8 @@ import { useRouter } from 'next/router';
 
 import dayjs from 'dayjs';
 
+import useMe from 'hooks/me';
+
 import { translatedLanguageNameForLocale } from 'helpers/intl';
 
 import Breadcrumbs from 'containers/breadcrumbs';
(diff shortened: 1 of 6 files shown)


frontend/containers/open-call-page/header/component.tsx
```diff
@@ -60,6 +68,18 @@ export const OpenCallHeader: FC<OpenCallHeaderProps> = ({
...
 
+  const handleFavoriteClick = () => {
+    // This mutation uses a 'DELETE' request when the isFavorite is true, and a 'POST' request when is false.
+    favoriteOpenCall.mutate(
+      { id, isFavourite: favourite },
+      {
+        onSuccess: (data) => {
+          openCall.favourite = data.favourite;
+        },
+      }
```
clementprdhomme, line 79: I think that more recently, we've stopped doing this and relied on the invalidation provided by React Query instead (see the investor and PD public page). Do you think we could apply the same here?

frontend/containers/open-call-page/investor/component.tsx
```diff
@@ -5,21 +5,37 @@ import { FormattedMessage } from 'react-intl';
...
+
+  const handleFavoriteClick = () => {
+    // This mutation uses a 'DELETE' request when the isFavorite is true, and a 'POST' request when is false.
+    favoriteOpenCall.mutate(
+      { id: openCall.id, isFavourite: openCall.favourite },
+      {
+        onSuccess: (data) => {
+          openCall.favourite = data.favourite;
+        },
+      }
```
clementprdhomme, line 35: Same comment here.

PR #495, opened 2022-08-23
feat: Notify investor owner when open call is destroyed via API

martintomas:
Send notification to investor owner when his/her open call gets destroyed via API.

## Testing instructions

Via rswag -- check that you get e-mail when your open call is destroyed

## Tracking

https://vizzuality.atlassian.net/browse/LET-959

### DIFF
--- a/backend/config/locales/zu.yml
+++ b/backend/config/locales/zu.yml
@@ -323,6 +323,12 @@ zu:
     project_destroyed:
       subject: Project has been removed.
       content: Project %{project_name} which you have been collaborator of have been removed.
+  investor_mailer:
+    greetings: Dear %{full_name},
+    farewell_html: Best Regards,<br />Your HeCo Team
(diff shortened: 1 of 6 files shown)


backend/config/locales/zu.yml
```diff
@@ -323,6 +323,12 @@ zu:
     project_destroyed:
       subject: Project has been removed.
       content: Project %{project_name} which you have been collaborator of have been removed.
+  investor_mailer:
+    greetings: Dear %{full_name},
+    farewell_html: Best Regards,<br />Your HeCo Team
+    open_call_destroyed:
+      subject: Open Call has been removed.
+      content: Open Call with name %{open_call_name} have been removed.
```
tsubik, line 331: I think it should be *has been removed

tsubik (approved):
Works great! Just small typo

PR #496, opened 2022-08-24
Users can preview drafted open calls

SARodrigues:
## Description

TODO

## Testing instructions

TODO

## Tracking

[LET-893](https://vizzuality.atlassian.net/browse/LET-893)


frontend/containers/dashboard/open-calls/table/cells/actions/component.tsx
```diff
@@ -97,24 +100,24 @@ export const CellActions: FC<CellActionsProps> = ({
...
-        ) : ( */}
-        <RowMenuItem key="open">
-          <FormattedMessage defaultMessage="View open call page" id="7sVapx" />
-        </RowMenuItem>
-        {/* )} */}
+        ) : (
+          <RowMenuItem key="open">
+            <FormattedMessage defaultMessage="View open call page" id="7sVapx" />
+          </RowMenuItem>
+        )}
```
SARodrigues, line 120: We forgot to re-enable this on #491

frontend/containers/dashboard/open-calls/table/cells/actions/component.tsx
```diff
@@ -97,24 +100,24 @@ export const CellActions: FC<CellActionsProps> = ({
           <FormattedMessage defaultMessage="Edit" id="wEQDC6" />
         </a>
       </Link>
-      <RowMenu direction="top" onAction={handleRowMenuItemClick}>
+      <RowMenu direction={isFirst ? 'bottom' : 'top'} onAction={handleRowMenuItemClick}>
```
SARodrigues, line 103: I don't think this is an ideal solution, but it's a quick fix to prevent the first entry from being cut off. 

Screenshot 2022-08-24 at 05 04 51

We've looked into this before (the projects list had a similar issue) and didn't find a solution quickly, (`@react-aria/menu` does have a `shouldFlip` prop, but it was not working) so we ended up "solving it" this way as well.

frontend/containers/dashboard/open-calls/table/cells/actions/component.tsx
```diff
@@ -36,15 +36,15 @@ export const CellActions: FC<CellActionsProps> = ({
...
-      //   return;
-      // case 'preview':
-      //   router.push(`${Paths.OpenCall}/${slug}/preview`);
-      //   return;
+      case 'edit':
+        router.push(
+          `${Paths.OpenCall}/${slug}/edit?returnPath=${encodeURIComponent(router.asPath)}`,
+          `${Paths.OpenCall}/${slug}/edit`
+        );
+        return;
```
SARodrigues, line 44: We forgot to re-enable this on https://github.com/Vizzuality/heco-invest/pull/491

clementprdhomme (approved):

clementprdhomme:
@SARodrigues I've added a small commit to remove the `@ts-ignore` line.

If you look at the type of `withLocalizedRequests`, you'll see that it receives the following generic `T extends GetStaticProps & GetServerSideProps`. `GetStaticProps` and `GetServerSideProps` have some keys in common but not all. Particularly, they don't share the `query` one.

The trick is simply to tell the static analyser that the generic is of type `GetServerSideProps`.

SARodrigues:
@clementprdhomme It makes sense now, thank you for explaining!

PR #498, opened 2022-08-24
feat: API for marking/unmarking open calls as favourite

martintomas:
New API endpoint which adds possibility to mark and unmark open calls as favourite. API is basically same as other favourite endpoints for projects, investors, etc.

## Testing instructions

Via rswag doc -- try that you can mark open call as favourite and also remove it. Don't forget that only approved user can do it ;)

## Tracking

https://vizzuality.atlassian.net/browse/LET-989


tsubik (approved):
Works great! :+1:

PR #507, opened 2022-08-25
Add open call cards,  open call search page

SARodrigues:
## Description

**In this PR:**
- Added an `OpenCallCard` component
  _Based on the existing `ProfileCard` & `ProjectCard`_ 
  - Added its respective Storybook story  
    _also added an investor.json and an openCall.json mock files to help with Storybook stories_
- `open-call-service` changes:  
  - Added a new hook to retrieve the list of open calls from the backend
    _with pagination, filtering, sorting and searching support_
- `DiscoverPageLayout` changes:  
  - Small refactoring to simplify how we load projects/pds/investors  
  - Added support for open calls  
    _including navigation badge with the correct count_
- `discover/open-calls` page changes:  
  - Populated the page with a list of open calls  
    _using the new `OpenCallCard` component_  
    - Pagination is implemented and working
    - Sorting is implemented
      _Note: sorting by `instrument_type` sends the query param to the endpoint, but the backend doesn't have it implemented yet_
    - Filtering is implemented
      _~Note: the frontend sends the query params to the endpoint, but it's not all implemented in the backend~_
    - Searching by keywords is implemented  
- `SortingButtons` component and `useSortingByOptions` hook changes: 
  - ~Added a `instrument_type` to the sorting buttons types~  
  - Refactored the hook to specify the target page we need the sorting options for
    _eg: if we pass `projects` it'll return the sorting options for projects and so on. By default it'll only return the `name` and `date` options_. 
- Added `Head` titles
  _to the discover pages_

## Testing instructions

Verify that:  
- `discover/open-calls` loads correctly  
  - The correct number of search results appears in the navigation badges  
  - Non-ongoing open calls have a ribbon with the text "Open call ended"  on their cards  
  - The information displayed in the cards is correct  
  - The investor info / tags displayed in the cards is correct  
- Pagination works as expected  
- Sorting works as expected  
  **Note:** Sorting by `instrument_type` doesn't seem to be implemented in the backend, but we can check for correct query params in the url/dev tools
- Searching works as expected  
- Filtering: check that the correct query params are sent in the url/dev tools
- Sorting: check that the correct query params are sent in the url/dev tools

## Tracking

[LET-888](https://vizzuality.atlassian.net/browse/LET-888)

## Screenshot

Screenshot 2022-08-25 at 13 45 07


frontend/layouts/discover-page/component.tsx
```diff
@@ -38,87 +40,56 @@ export const DiscoverPageLayout: FC<DiscoverPageLayoutProps> = ({
...
-  // This shouldn't be needed, but due to CSS positioning / z-index issues we need to have the DiscoverSearch
-  // components both in the header and in this layout; which one is visible depends on the screen resolution.
-  // These states are here to keep both DiscoverSearch in sync, in case the user resizes their screen.
-
   const [sorting, setSorting] =
     useState<{ sortBy: SortingOptionKey; sortOrder: SortingOrderType }>(defaultSorting);
 
+  const sortingOptionsTarget = useMemo(() => {
+    if (pathname.startsWith(Paths.Projects)) return 'projects';
+    if (pathname.startsWith(Paths.OpenCalls)) return 'openCalls';
```
clementprdhomme, line 48: You could use existing strings:
```suggestion
    if (pathname.startsWith(Paths.Projects)) return Queries.Project;
    if (pathname.startsWith(Paths.OpenCalls)) return Queries.OpenCall;
```

frontend/layouts/discover-page/helpers.ts
```diff
@@ -1,32 +1,50 @@
 import { useIntl } from 'react-intl';
 
+import { compact } from 'lodash-es';
+
 import { SortingOptionType } from 'components/sorting-buttons';
 
 /** Returns the sorting-by types array with translated labels  */
-export const useSortingByOptions = (): SortingOptionType[] => {
+export const useSortingByOptions = (target?: 'projects' | 'openCalls'): SortingOptionType[] => {
```
clementprdhomme, line 8: If you used my suggestion above:
```suggestion
export const useSortingByOptions = (target?: Queries.Project | Queries.OpenCall): SortingOptionType[] => {
```

You'd have to change the keys below as well.

frontend/pages/discover/open-calls.tsx
```diff
@@ -1,40 +1,106 @@
...
 
-const OpenCallsPage: PageComponent<OpenCallsPageProps, DiscoverPageLayoutProps> = ({ enums }) => {
+const OpenCallsPage: PageComponent<OpenCallsPageProps, DiscoverPageLayoutProps> = ({
+  data: openCalls = [],
+  loading = false,
+  meta,
+}) => {
+  const openCallsContainerRef = useRef(null);
+  const { props: paginationProps } = usePagination(meta);
+  const [showNotice, setShowNotice] = useState<boolean>(false);
```
clementprdhomme, line 44: Good that it's off by default 👍️.

clementprdhomme (commented):
Unrelated but I think we could fix it in this PR: the Discover page is missing a title.

SARodrigues:
Along with the code review changes, I've also fixed an issue where the open call cards were displaying the investors' instrument types instead of the open call's, which is why I thought filtering by instrument_type was not working.

clementprdhomme (approved):

PR #508, opened 2022-08-25
Fixes inconsistencies in displayed languages for first time users (mostly)

clementprdhomme:
This PR fixes an issue where requests made on the client could use the application's default locale instead of the locale detected by Next.js. This would result in pages with a mix of two languages:
- The server-side content would render in the locale detected by Next.js
- The content coming from client-side requests would be in the application's default locale

## Testing instructions

This steps can be replicated with either Chrome or Firefox.

1. Set your browser to English
2. If you have any private windows opened, close them
3. Open a new private window
4. Open the homepage

The whole content will be in English.

5. Navigate to the Discover page

The whole content should be in English. Particularly, have a look at the projects cards, since this is what was displayed in Spanish (the app's default language).

## Tracking

- [LET-958](https://vizzuality.atlassian.net/browse/LET-958)
- [LET-948](https://vizzuality.atlassian.net/browse/LET-948) (likely to be the same issue; to test it follow the same steps but sign up as a PD and check if the dropdown options are in English)

### DIFF
--- a/frontend/pages/_app.tsx
+++ b/frontend/pages/_app.tsx
@@ -1,3 +1,5 @@
+import { useEffect } from 'react';
+
 import { IntlProvider } from 'react-intl';
 import { QueryClient, QueryClientProvider } from 'react-query';
 import { Provider as ReduxProvider } from 'react-redux';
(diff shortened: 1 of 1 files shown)


SARodrigues (approved):
Nice detective work!  

When I was looking at it yesterday, I didn't even come _close_ to finding the actual cause.

PR #513, opened 2022-08-29
feat: API for list of OpenCallApplications for private area

martintomas:
New API endpoint (`GET /api/v1/account/open_call_applications`) which returns list of open call applications for:
 - investor: applications for investor open calls
 - project_developer: applications for PD projects

Data returned by this endpoint are for both type of accounts same, therefore FE is in full control of what is shown (works same way as other endpoints). Endpoint supports `includes` param so it is possible to obtain details of investor/project/project_developer/open_call at any point <-- required for example for details.

Searching works slightly differently for each account type, so I have split it and different ransack query is used for Investor and PD.

## Testing instructions

New endpoint is again part of rswag documentation. Login as PD and check that you get list of applications which you have created. Login as Investor and check that you see all applications for your all open calls.

## Tracking

https://vizzuality.atlassian.net/browse/LET-890
https://vizzuality.atlassian.net/browse/LET-923
https://vizzuality.atlassian.net/browse/LET-895
https://vizzuality.atlassian.net/browse/LET-972
https://vizzuality.atlassian.net/browse/LET-924
https://vizzuality.atlassian.net/browse/LET-971

### DIFF
--- a/backend/app/controllers/api/v1/accounts/open_call_applications_controller.rb
+++ b/backend/app/controllers/api/v1/accounts/open_call_applications_controller.rb
@@ -6,6 +6,8 @@ module API
         load_and_authorize_resource
 
         def index
+          @open_call_applications = @open_call_applications.where project_id: filter_params[:project_id] if filter_params[:project_id].present?
+          @open_call_applications = @open_call_applications.where open_call_id: filter_params[:open_call_id] if filter_params[:open_call_id].present?
           @open_call_applications = search_by filter_params[:full_text], @open_call_applications if filter_params[:full_text].present?
           @open_call_applications = @open_call_applications.order(created_at: :desc)
           render json: OpenCallApplicationSerializer.new(
@@ -49,7 +51,7 @@ module API
         end
 
         def filter_params
-          params.fetch(:filter, {}).permit :full_text
+          params.fetch(:filter, {}).permit :project_id, :open_call_id, :full_text
         end
       end
     end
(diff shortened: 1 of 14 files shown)


tsubik (commented):
At some point it would be worth it to update seeds to contain more test data, some project developers and investors with open calls applications, as creating that manually is very time-consuming.

backend/app/controllers/api/v1/accounts/open_call_applications_controller.rb
```diff
@@ -5,6 +5,20 @@ class OpenCallApplicationsController < BaseController
         around_action(only: %i[create]) { |_, action| set_locale(language: current_user&.account&.language, &action) }
         load_and_authorize_resource
 
+        def index
+          @open_call_applications = @open_call_applications.includes(open_call: {investor: :account}, project: {project_developer: :account})
+          @open_call_applications = @open_call_applications.where project_id: filter_params[:project_id] if filter_params[:project_id].present?
```
tsubik, line 10: I did not find a requirement for this and to filter by open_call_id but I trust this is needed :)

tsubik: Ahh, I guess this is for details endpoint, to get only those for project/open call.

tsubik: or should we add a separate show endpoint for details?

martintomas:
> At some point it would be worth it to update seeds to contain more test data, some project developers and investors with open calls applications, as creating that manually is very time-consuming.

I can add `open call application` creation to seeding, np

tsubik (approved):
Looks great! :+1:

PR #515, opened 2022-08-29
feat: API for updating open call applications (funding/not_funding)

martintomas:
This PR adds three new endpoints:
 - `PUT /api/v1/account/open_call_applications/{id}` -- available to PD, allows to change message of application
 - `POST /api/v1/account/open_call_applications/{id}/funding` -- available to Investor, allows to mark application as funded
 - `POST /api/v1/account/open_call_applications/{id}/not_funding` -- available to Investor, allows to mark application as not funded

## Testing instructions

All new endpoints can be found at rswag documentation

## Tracking

https://vizzuality.atlassian.net/browse/LET-897

### DIFF
--- a/backend/app/controllers/api/v1/accounts/open_call_applications_controller.rb
+++ b/backend/app/controllers/api/v1/accounts/open_call_applications_controller.rb
@@ -2,7 +2,7 @@ module API
   module V1
     module Accounts
       class OpenCallApplicationsController < BaseController
-        around_action(only: %i[create]) { |_, action| set_locale(language: current_user&.account&.language, &action) }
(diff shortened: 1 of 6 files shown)


tsubik (approved):
I would probably do it just by one `update` endpoint and permit different parameters, but I read in the ticket description that this was the recommended approach. I would not argue, both ways are good.

martintomas:
> I would probably do it just by one `update` endpoint and permit different parameters, but I read in the ticket description that this was the recommended approach. I would not argue, both ways are good.

I liked that forbidden error is returned when investor/pd tries to edit what he is not allowed so I have preferred this approach, but for sure just allowing some params for different roles would work as well ;)

PR #524, opened 2022-08-31
Fix an issue uploading images to projects

clementprdhomme:
This PR fixes an issue where the application would crash when uploading images to a project.

## Testing instructions

- Upload images
- Delete images
- Change which image is the cover

## Tracking

[LET-1019](https://vizzuality.atlassian.net/browse/LET-1019)


SARodrigues (changes requested):
It looks like the error has gone away, but there's a weird behaviour when attempting to delete images:  


https://user-images.githubusercontent.com/6273795/187656274-4a25cdf5-ad1d-4682-a9e3-f1ac24298b7b.mp4

SARodrigues (approved):

PR #529, opened 2022-09-01
Investor users can see the details of a project that has applied to an open call in Dashboard

SARodrigues:
## Description

This PR adds a page where an Investor can see the details of a project/PD that has applied to an open call. 

**In this PR:**  
- Minor simplification of how open calls are fetched  
- Added a placeholder page for the list of applications to an open call (Investor dashboard)  
  _This was used mostly to set-up/test breadcrumbs_  
- Added the open-call-application/project page to the Investor dashboard  

**Notes:**  
- The funding toggle is tracked by a [different task](https://vizzuality.atlassian.net/browse/LET-986)

## Testing instructions

- Using a PD account:  
  - Create an Open call application
  - Visit dashboard/open-call-applications, open the details page of the application, copy its `id` from the url  
- Using an Investor account:  
  - Visit the open call public page and copy its `slug` from the URL  
  - Access the open call application details page via url in the following format: 
    `http://localhost:3000/en/dashboard/open-call/${openCallSlug}/application/${applicationId}`

Verify that:  
  - Breadcrumbs match the designs  
  - Project name is correct  
  - The project developer information is correct  
  - The open call application message is correct  
  - Links work correctly 

## Tracking

[LET-912](https://vizzuality.atlassian.net/browse/LET-912)

## Screenshot

Screenshot 2022-09-01 at 11 30 02

### DIFF
--- a/frontend/pages/dashboard/open-call/[openCallId]/application/[applicationId].tsx
+++ b/frontend/pages/dashboard/open-call/[openCallId]/application/[applicationId].tsx
@@ -111,12 +111,13 @@ export const OpenCallDetailsPage: PageComponent<OpenCallDetailsPageProps, Dashbo
   return (
     <ProtectedPage permissions={[UserRoles.Investor]}>
       <Head
-        title={`${
-          project?.name || intl.formatMessage({ defaultMessage: 'Project', id: 'k36uSw' })
-        } ${intl.formatMessage({
-          defaultMessage: 'application',
-          id: 'HFalmD',
-        })}`}
+        title={intl.formatMessage(
+          { defaultMessage: '{projectName} application', id: 'sVbuuM' },
+          {
+            projectName:
+              project?.name || intl.formatMessage({ defaultMessage: 'Project', id: 'k36uSw' }),
+          }
+        )}
       />
       <DashboardLayout
         header="breadcrumbs"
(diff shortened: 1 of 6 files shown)


frontend/pages/dashboard/open-call/[openCallId]/application/[applicationId].tsx
```diff
@@ -0,0 +1,235 @@
...
+
+  return (
+    <ProtectedPage permissions={[UserRoles.Investor]}>
+      <Head
+        title={`${
+          project?.name || intl.formatMessage({ defaultMessage: 'Project', id: 'k36uSw' })
+        } ${intl.formatMessage({
+          defaultMessage: 'application',
+          id: 'HFalmD',
+        })}`}
```
Comment (clementprdhomme) [inline frontend/pages/dashboard/open-call/[openCallId]/application/[applicationId].tsx:119]: I believe this should be one string only: `{projectName} application` and `projectName` is either `project?.name` or the translated string “Project”.

clementprdhomme (approved):
Feel free to merge after the small change.

frontend/pages/dashboard/open-call/[openCallId]/application/[applicationId].tsx
Comment (SARodrigues) [reply frontend/pages/dashboard/open-call/[openCallId]/application/[applicationId].tsx:119]: You're right, makes more sense the way you described it.

PR #530, opened 2022-09-01
feat: Notify PD/Investors on account deletion

martintomas:
This PR adds all required notifications when:
 - open call is deleted --> notifies investor and project developers which applied for it
 - project is deleted --> notifies involved project developers and investor which had application for this project
 - account is deleted --> all use cases described at https://vizzuality.atlassian.net/browse/LET-876

I was playing with idea if it wouldn't be easier to move email sending to callbacks, but have bad experience with it so decided in the end to keep it in controllers. To keep controllers slim, I have moved all notification logic to new services which are responsible for sending emails upon deletion.

## Testing instructions
  - Apply to some Open call as project developer and delete this open call as investor. Project developer should be notified.
  - Apply to some Open call as project developer and delete project which you have applied with. Investor of open call should be notified.
  - Apply to some Open call as project developer and delete project developer --> project developer should be notified and investor of open call which you have applied for.
  - Apply to some Open call as project developer and delete investor of open call --> investor should be notified and project developer of project that application was for.

All use cases should be testable via backoffice, first two are also testable via rswag documentation.

## Tracking

https://vizzuality.atlassian.net/browse/LET-915
https://vizzuality.atlassian.net/browse/LET-909
https://vizzuality.atlassian.net/browse/LET-877
https://vizzuality.atlassian.net/browse/LET-881

### DIFF
--- a/backend/app/services/accounts/with_email_notification.rb
+++ b/backend/app/services/accounts/with_email_notification.rb
@@ -1,0 +1,28 @@
+module Accounts
+  class WithEmailNotification
+    attr_accessor :account, :emails
+
+    def initialize(account)
+      @account = account
+      @emails = []
+    end
+
+    def destroy!
+      collect_emails
+      account.destroy!
+      emails.map(&:deliver_later)
+    end
+
+    private
+
+    def collect_emails
+      account.users.each { |user| emails << UserMailer.destroyed(user.email, user.full_name, user.locale) }
+      account.investor&.open_call_applications.to_a.each do |open_call_application|
+        emails << ProjectDeveloperMailer.open_call_destroyed(open_call_application.project_developer, open_call_application.open_call.name)
+      end
+      account.project_developer&.open_call_applications.to_a.each do |open_call_application|
+        emails << InvestorMailer.project_destroyed(open_call_application.investor, open_call_application.project.name)
+      end
+    end
+  end
+end
--- a/backend/app/services/open_calls/destroy.rb
+++ b/backend/app/services/open_calls/destroy.rb
@@ -1,5 +1,5 @@
 module OpenCalls
-  class WithEmailNotification
+  class Destroy
     attr_accessor :open_call, :emails
 
     def initialize(open_call)
@@ -7,7 +7,7 @@ module OpenCalls
       @emails = []
     end
 
-    def destroy!
+    def call
       collect_emails
       open_call.destroy!
       emails.map(&:deliver_later)
--- a/backend/app/services/open_calls/destroy.rb
+++ b/backend/app/services/open_calls/destroy.rb
@@ -17,8 +17,12 @@ module OpenCalls
 
     def collect_emails
       emails << InvestorMailer.open_call_destroyed(open_call.investor, open_call.name)
-      open_call.project_developers.each do |project_developer|
-        emails << ProjectDeveloperMailer.open_call_destroyed(project_developer, open_call.name)
+      open_call.open_call_applications.each do |open_call_application|
+        emails << ProjectDeveloperMailer.open_call_destroyed(
+          open_call_application.project_developer,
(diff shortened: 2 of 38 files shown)


backend/app/services/accounts/with_email_notification.rb
```diff
@@ -0,0 +1,28 @@
+module Accounts
+  class WithEmailNotification
```
tsubik, line 2: Are there more methods than `destroy!` planned for this service? I find it little weird to be about email notifications than destroying, have you thought about services like `Accounts::Destroy`, than would just have `call` or `perform` method?

tsubik: I imagine when adding more services there could be more actions performed than just sending email notification on common actions like `destroy` or `update`

martintomas: I could not actually decide :). My first idea was to write it as wrapper (use `method_missing` to forward all stuff to main object), but then I got feeling that maybe it is more robust approach then it is required now. So in the end, wrapper looks more like service class and from this view, `call` or `perform` makes more sense. What do you think? Should I finalise wrapper idea and add `method_missing` stuff which can be later on enhanced with email notification on update/create/etc. when it gets too complex at controller or should I transform it to single oriented service with `call` method?

tsubik: I'm not sure what do you mean by wrapper? I just thought about changing names of services to be name of actual command so here `Accounts::Destroy`, the inside of the service looks perfectly fine :)

On the other hand service objects could quickly get messy without good conventions, and as there are so many conventions out there so I'm also not sure what would be the best.

agnessa: I think I get what you mean Tomasz, i.e. service called `Accounts::Destroy` and `destroy!` method renamed to `call` - I think I prefer it this way

martintomas: Okay, I still like idea of wrapper/decorator, but truth is that it is over-engineering at this point so I give up on this approach for now :D. My idea was that we could do something similar at future `3rdPartyNotification(SmsNotification(EmailNotification(account))).destroy!` or `SmsNotification(EmailNotification(account)).update!(params)` which makes it easy to combine and encapsulate all notification responsibilities.

Whatever, it is overkill now, will rename services to `Accounts::Destroy` so it follows our services approach ;)

backend/app/services/open_calls/destroy.rb
```diff
@@ -0,0 +1,25 @@
...
+      collect_emails
+      open_call.destroy!
+      emails.map(&:deliver_later)
+    end
+
+    private
+
+    def collect_emails
+      emails << InvestorMailer.open_call_destroyed(open_call.investor, open_call.name)
+      open_call.project_developers.each do |project_developer|
```
tsubik, line 20: Is it possible that project developer applies twice with different project for the same open call? I just tested this and got 2 same emails for one project developer, the project name that applied for open call is not in the emails, so the emails were the same. I wonder if we should do a distinct here or change email text, or it should not be possible to apply twice. Dunno :)

tsubik (approved):
Overall looks great! Just added a question about duplicate emails edge case, if we should care about it now or later.

backend/app/services/open_calls/destroy.rb
martintomas: Hmm, there is restriction where you cannot apply for same open call with same project, but you can apply for same open call with different project. To by honest, I dont know what is right solution here :). Do you have some idea @agnessa or should I ask at Slack? From my point of view, the easiest is to add project name to email...

agnessa: I agree, easiest to inject project name somewhere in the email 👍

PR #539, opened 2022-09-06
Refactor of the table component to solve column content overlapping

SARodrigues:
## Description

This PR fixes an issue where the content of the last two columns of a table can overlap, especially if the second to last's content is long. 

Other changes:  
- I've tweaked the table component styling so that the content/columns adapt a bit better
  _Which should help with the fact that we cannot control their content length, especially given translations in different languages_
  _It should also be easier to customise the table further in the future as well_  
- Fixed tables not overflowing with a scroll bar laterally when the screen is too small  
  _this restores the initial behaviour they had initially_
- Minor tweaks to better approximate the designs  

**Notes:**  
I tried to keep the table's component code changes to the very minimum, given that this is meant as a quick bug-fix and the fact it's used throughout the dashboard; I didn't want to risk introducing breaking changes. Most of the refactoring was to: 
  - Not use the `style` s that `react-table` introduces  
  - Avoid non-standard styling of `th`'s, `tr`'s and the likes (eg: `flex`), as this will prevent the headers and cells from being aligned (this was one of the causes)  
  - Ensure sorting still works correctly  

## Testing instructions

- Verify that the bug no longer exists and the content from different columns doesn't overlap  
- Verify that all tables display correctly, sorting works correctly and there are no crashes

## Tracking

[LET-1026](https://vizzuality.atlassian.net/browse/LET-1026)

## Screenshots  

**Before**
Screenshot 2022-09-06 at 15 07 30

**After**  
Screenshot 2022-09-06 at 15 07 05

### DIFF
--- a/frontend/components/table/component.tsx
+++ b/frontend/components/table/component.tsx
@@ -26,15 +26,6 @@ export const Table: FC<TableProps> = ({
   isOwner,
   accountName,
 }: TableProps) => {
-  const DEFAULT_COLUMN = React.useMemo(
-    () => ({
-      minWidth: 30,
-      width: 150,
-      maxWidth: 200,
-    }),
-    []
-  );
-
   const {
     getTableProps,
     // headers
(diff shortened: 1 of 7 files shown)


clementprdhomme (approved):
Everything's good!

PR #540, opened 2022-09-06
feat: Adding impact filter to project maps endpoint

martintomas:
I have double checked projects API and projects map API so they use same service for filtering/searching and allow same params. Project map endpoint missed `impact` filter so I allowed it there. There is no extra test, because both endpoints use `API::Filterer.new(projects, filter_params.to_h).call` which is tested quite well ;)

## Testing instructions

Just check that same params can be send to project and map endpoint at rswag doc.

## Tracking

https://vizzuality.atlassian.net/browse/LET-1034

### DIFF
--- a/backend/app/controllers/api/v1/projects/maps_controller.rb
+++ b/backend/app/controllers/api/v1/projects/maps_controller.rb
@@ -12,8 +12,8 @@ module API
         private
 
         def filter_params
-          params.fetch(:filter, {}).permit :category, :sdg, :instrument_type, :ticket_size, :only_verified,
-            :full_text, :priority_landscape
+          params.fetch(:filter, {}).permit :category, :sdg, :instrument_type, :ticket_size, :impact,
+            :only_verified, :full_text, :priority_landscape
         end
       end
     end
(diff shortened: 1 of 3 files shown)


agnessa (commented):
Thanks! just double checking, don't we need this kind of line in the maps endpoint for the keyword search to work properly?
`projects = projects.dynamic_search ["name_#{I18n.locale}"], filter_params[:full_text], [] if filter_params[:full_text].present?`

martintomas:
> Thanks! just double checking, don't we need this kind of line in the maps endpoint for the keyword search to work properly? `projects = projects.dynamic_search ["name_#{I18n.locale}"], filter_params[:full_text], [] if filter_params[:full_text].present?`

We should be good @agnessa . `API::Filterer` already handles localization of columns which are used for searching (https://github.com/Vizzuality/heco-invest/blob/develop/backend/app/services/api/filterer.rb#L44).

martintomas:
> > Thanks! just double checking, don't we need this kind of line in the maps endpoint for the keyword search to work properly? `projects = projects.dynamic_search ["name_#{I18n.locale}"], filter_params[:full_text], [] if filter_params[:full_text].present?`
> 
> We should be good @agnessa . `API::Filterer` already handles localization of columns which are used for searching (https://github.com/Vizzuality/heco-invest/blob/develop/backend/app/services/api/filterer.rb#L44).

Public endpoint for Projects uses more complex full text searching (more attributes) so that is why `API::Filterer` is used there. Private API for Projects does full text just for name and no additional filters, that is why it have just `projects.dynamic_search ["name_#{I18n.locale}"], filter_params[:full_text], []`. My guess was that `map` endpoint should copy behaviour of public endpoint. If we wanna copy behavior of private endpoint, `projects.dynamic_search ["name_#{I18n.locale}"], filter_params[:full_text], []` is better choice.

agnessa:
you're right, it needs to behave like the public endpoint 👍

PR #542, opened 2022-09-07
refactoring: Prefer usage of account name at backoffice

martintomas:
Ticket mention that `Investor name` and `Project developer name` should be renamed to `Account name`, but none of these two strings are used at backoffice at all. On other hand, we are using `Name` or `Profile name` at backoffice so I have decided to rename these two strings to `Account name` so we follow same logic which is probably gonna be implemented by FE.

## Testing instructions

Just briefly check backoffice ;)

## Tracking

https://vizzuality.atlassian.net/browse/LET-1035

### DIFF
--- a/backend/app/views/backoffice/investors/index.html.erb
+++ b/backend/app/views/backoffice/investors/index.html.erb
@@ -7,7 +7,7 @@
   <thead>
     <tr>
       <th>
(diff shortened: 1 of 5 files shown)


agnessa (approved):
Thanks! yes, makes sense to get rid of "Profile name".

PR #545, opened 2022-09-07
Fix broken language notice when editing an open call

SARodrigues:
## Description

This PR fixes an issue causing the _"The content of this open call should be written (...)"_ notice to not display the language the open call's content should be written in. 

## Testing instructions

1. Sign in with an investor account
2. Visit the dashboard, open calls tab  
3. Click to “Edit“ an open call  

Verify that the notice is no longer broken. 

## Tracking

[LET-1070](https://vizzuality.atlassian.net/browse/LET-1070)

## Screenshots  

**Before** 
Screenshot 2022-09-07 at 13 20 00

**After** 
Screenshot 2022-09-07 at 13 20 19

### DIFF
--- a/frontend/pages/open-call/[id]/edit.tsx
+++ b/frontend/pages/open-call/[id]/edit.tsx
@@ -101,7 +101,6 @@ const EditOpenCall: PageComponent<EditOpenCallProps, FormPageLayoutProps> = ({
         onComplete={handleOnComplete}
         initialValues={openCall}
         isLoading={!openCall && isFetchingProject}
-        locale={openCall?.language}
         enums={enums as GroupedEnums}
       />
     </ProtectedPage>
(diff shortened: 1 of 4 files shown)


frontend/pages/open-call/[id]/edit.tsx
```diff
@@ -101,9 +101,7 @@ const EditOpenCall: PageComponent<EditOpenCallProps, FormPageLayoutProps> = ({
         onComplete={handleOnComplete}
         initialValues={openCall}
         isLoading={!openCall && isFetchingProject}
-        // The language of the content will be the same as the language of the account, which the
-        // API will automatically set
-        locale={null}
+        locale={openCall?.language}
```
Comment (clementprdhomme) [inline frontend/pages/open-call/[id]/edit.tsx:104]: For the projects, we use:
```ts
const contentLocale = defaultValues?.language || userAccount?.language;
```

I wonder if we shouldn't fallback to the account's language here as well. What do you think?

clementprdhomme (approved):

frontend/pages/open-call/[id]/edit.tsx
Comment (SARodrigues) [reply frontend/pages/open-call/[id]/edit.tsx:104]: I think it's a good idea. I'll add the fallback in the open calls form (much how it was implemented for the projects).

PR #546, opened 2022-09-07
Add open calls to the dashboard favorites

SARodrigues:
## Description

This PR adds the open calls to the dashboard favourites. 

## Testing instructions

Visit the dashboard, favourites tab. Verify that: 
- The user can see a list of favourited open calls  
- The sidebar badges display the count of favourited open calls
- Above the list there is a count of favourited open calls
- The user can unfavourite an open call by hovering the card and clicking the trash can icon
  _Same behaviour as implemented for the other cards in the favourites_

## Tracking

[LET-1063](https://vizzuality.atlassian.net/browse/LET-1063)

## Screenshot

Screenshot 2022-09-07 at 13 43 43

### DIFF
--- a/frontend/containers/open-call-card/component.tsx
+++ b/frontend/containers/open-call-card/component.tsx
@@ -19,7 +19,7 @@ import { useEnums } from 'services/enums/enumService';
 
 import type { OpenCallCardProps } from './types';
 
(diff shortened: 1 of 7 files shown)


clementprdhomme (approved):
Everything's good.

If I may, I would suggest adding a small margin to either the investor's name or category tags so that they are not touching:
The investor's name and category tags are touching in the favourites list

PR #552, opened 2022-09-08
fix: API includes returns only projects which user can see

martintomas:
It was possible to obtain draft projects or open calls through public project developer or investor endpoint when `includes` param is used. Problem is that same serializers are used across whole app (for public or private endpoints), therefore final solution needed to be robust enough to deal with this.

In the end, I have decided to use `cancancan` which is great place to define what user actually can/cannot see. I have added new method to base serializer called `has_many_restricted` which takes into account user rights. This method makes sure that projects or open calls which are drafts are hidden at public endpoints even when data are loaded by serializer.

I was playing with idea to use `has_many_restricted ` everywhere (not just projects and open calls), because there can be other leaks, but because we have added `cancancan` later, it is not complete and some stuff is missing there or not defined enough. At this point, it would be huge change to add it everywhere which would probably broke something so I have decided to add it only at places which are related to current bug and we can eventually add it later, step by step to other serializers.

## Testing instructions

Use Rswag documentation -- get project developer or investor and verify that projects or open calls records obtained via `includes` does not contain drafts when public endpoints are used.

## Tracking

https://vizzuality.atlassian.net/browse/LET-1071


agnessa (approved):
OK, I knew the `includes` would come to bite us back, thanks!

PR #556, opened 2022-09-09
feat: Memoize filters/sorting/pagination at backoffice

martintomas:
Task was about maintaining sorting during opening/editing, but basically same problem happens during canceling, deletion of records, etc. At beginning, I have started to put to every influenced link or form `q` param so this gets remembered, but it got quickly out of hand --> there is just too many places where this param needs to be added and  it is so easy to forget to add it to every new piece of code. In the end, I have decided to use `session` which keeps saved used filters and last page of table per every controller. Until you reset filters manually (or you logout --> login again) `q` and `page` params are kept at session and pull out from there everytime when you open some Backoffice table.

## Testing instructions

Open backoffice and verify that sorting/pagination/filters are remembered between jumps to edit forms.

## Tracking

https://vizzuality.atlassian.net/browse/LET-955

### DIFF
--- a/backend/app/controllers/backoffice/base_controller.rb
+++ b/backend/app/controllers/backoffice/base_controller.rb
@@ -3,6 +3,7 @@ module Backoffice
     include Pagy::Backend
     include Breadcrumbs
     include Localization
+    include FilterMemoization
 
     layout "backoffice"
 
--- a/backend/app/controllers/concerns/backoffice/filter_memoization.rb
+++ b/backend/app/controllers/concerns/backoffice/filter_memoization.rb
@@ -1,0 +1,22 @@
+module Backoffice
+  module FilterMemoization
+    def self.included(base)
+      base.before_action :save_filters, only: :index
+      base.before_action :load_filters, only: :index
+    end
+
+    private
+
+    def save_filters
+      session[:table_filters] ||= {}
+      session[:table_filters][controller_name] ||= {}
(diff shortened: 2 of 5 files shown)


martintomas:
> 



> I have a mixed feeling about this. In one project we have exactly this, because client wanted to persist filters. This is kind of feature, but that depends on preference. For me, it was weird going back to the page and having some filters applied already (also not synced with query string). I don't know, I normally would just pass those filter params to actions url, maybe with simple `helper_method` like this
> 
> ```
> def memoized_params
>   params.permit(:q, :page).to_h
> end
> ```



> I have a mixed feeling about this. In one project we have exactly this, because client wanted to persist filters. This is kind of feature, but that depends on preference. For me, it was weird going back to the page and having some filters applied already (also not synced with query string). I don't know, I normally would just pass those filter params to actions url, maybe with simple `helper_method` like this
> 
> ```
> def memoized_params
>   params.permit(:q, :page).to_h
> end
> ```

It was actually my first idea to use something similar to memoized_params, but this needs to be added to a lot of places (my guess is like 20-30 places) so I decided to use sessions which do all the work at background. On other hand I agree that forwarding params is more transparent then some magical session :). What do you think @agnessa , params or session? I dont mind to refactor this feature and use Tomasz idea ;).

agnessa:
well I see the point about the query string, but it works and as far as I'm concerned that's perfect :)

agnessa (approved):

PR #557, opened 2022-09-09
LET-937 BC3 API examples

agnessa:
This is a demonstration of retrieving values from the BC3 API in ruby. I used the [Java client](https://github.com/integratedmodelling/klab-api) for guidance. This is a naive implementation without error handling or tests, as I would like to double check the approach with BC3 and there is still some figuring out to do around passing geometries.


backend/app/services/klab/run.rb
```diff
@@ -0,0 +1,26 @@
...
+      # TODO: geometry
+      geometry = "aries.heco.locations.colombia_continental"
+      context_req = Klab::SubmitContextRequest.new(client)
+      context_resp = Klab::SubmitContextResponse.new(context_req.call(geometry))
+      ticket_req = Klab::TicketRequest.new(client, context_resp.ticket_id)
+      ticket_resp = Klab::TicketResponse.new(ticket_req.call)
+      # this part can take a few minutes - ticket gets resolved when observations for all indicators completed
+      until ticket_resp.resolved?
+        ticket_resp = Klab::TicketResponse.new(ticket_req.call)
+        sleep(5)
```
martintomas, line 14: Maybe it would be better to have background job to actually collect results, but sleep is probably fine for now

martintomas (approved):
Looks good, but I think it would be great to have tests for all these services. I like how readme describes how this stuff works, but also tests can be quite easy way how to document what is happening and what are expected responses.

backend/app/services/klab/run.rb
agnessa: yes, I guess a polling job every few seconds 👍

PR #558, opened 2022-09-09
refactoring: Message of OpenCallApplication is no more translated

martintomas:
Name of the PR is pretty self-explanatory. This PR removes translation from `message` attribute of open call application model.

## Testing instructions

I think it should be enough to test via rswag that new open call application can be still created.

## Tracking

https://vizzuality.atlassian.net/browse/LET-1065


backend/db/migrate/20220909092937_remove_translated_message_from_open_call_applications.rb
```diff
@@ -0,0 +1,15 @@
+class RemoveTranslatedMessageFromOpenCallApplications < ActiveRecord::Migration[7.0]
+  def change
+    add_column :open_call_applications, :message, :text
+    migrate_translated_messages!
```
agnessa, line 4: we could even make it not null now that we got rid of localisation :)

agnessa (approved):

PR #569, opened 2022-09-14
Let 1046 fe additional information about impact

barbara-chaves:
This PR adds the legends on the mini impact charts and the impact score on the project card  

## Description of the PR

Acceptance criteria

There will be a way to see the impacts and the scale on the impact graphs

- Design's decision:
~~- Display average score all the time~~ REMOVED FROM TASK 
- Display scale etc when hovering

Applies to 

- The project items in the projects list of the Discover page
- The list of projects at the bottom of a PD public page
- The list of favourites in the Dashboard (My area)

## Testing instructions

- Go to discover/projects. 
- The total impact score should be displayed in the project cards
- The impact charts on the project cards should display the legends when hovered, if the impact is calculated

* The same should applied to all project cards

## Tracking

[LET-1046](https://vizzuality.atlassian.net/browse/LET-836)


frontend/containers/impact-chart/component.tsx
```diff
@@ -128,11 +145,41 @@ export const ImpactChart: FC<ImpactChartProps> = ({
...
+                  <div className="px-1 -translate-x-[calc(100%+42px)] translate-y-5 bg-white border rounded-md w-min border-beige">
+                    {impactData[ImpactsEnum.Water].name}
+                  </div>
+                  <div className="px-1 bg-white border rounded-md translate-x-[42px] w-min border-beige">
+                    {impactData[ImpactsEnum.Climate].name}
+                  </div>
+                  <div className="px-1 translate-x-[-50%] translate-y-8 bg-white border rounded-md w-min border-beige">
+                    {impactData[ImpactsEnum.Community].name}
+                  </div>
+                </div>
```
clementprdhomme, line 175: A few comments on the tooltips:
- The border radius is smaller, you should use `rounded` instead of `rounded-md`
- The padding is bigger: `px-2 py-1.5`
- The font size is smaller: `text-2xs`
- The colour is a lighter gray: `text-gray-400`
- The font is bolder: `font-semibold`
- You should remove the red background that you used for testing

clementprdhomme: I also think you should create separate tooltips. This will allow you to have arrows on each of them.

For example, the top tooltip can be:
```jsx
            <Tooltip
              placement="top"
              arrow
              arrowClassName="border border-transparent border-b-beige border-r-beige"
              content={
                <div className="font-semibold text-center text-gray-400 text-2xs px-2 py-1.5 bg-white border rounded w-min border-beige">
                  {impactData[ImpactsEnum.Biodiversity].name}
                </div>
              }
            />
```

Inside of this, you would put the left tooltip, inside the right, etc. You will have to update the arrow slightly for each of them. For example, for the left tooltip, the `arrowClassName` prop should have this value: `-rotate-45 border border-transparent border-b-beige border-r-beige`.

Initially, this will not work because you need to make a small change to the `<Tooltip />` component so it passes a ref. Here is the details of the changes you need to make:
```jsx
import {  forwardRef } from 'react';

// More code

// Notice I've renamed this component
export const Inner: React.ForwardRefRenderFunction<any, TooltipProps> = (
  { children, content, arrow, maxWidth, arrowClassName, ...props },
  ref
) => {
  // More code

  return (
    <Tippy
      {...props}
      // Very important, set the `ref` prop on the `<div />`
      render={(attrs) => (
        <motion.div style={{ scale, opacity, maxWidth: maxWidth || 'none' }} {...attrs}>
          <div className="relative" ref={ref}>
            {/* More code */}
          </div>
        </motion.div>
      )}
      // More props here
    >
      {children}
    </Tippy>
  );
};

// You forward the ref here
export const Tooltip = forwardRef(Inner);

export default Tooltip;
```

Then don't forget to remove `useHover` from the the chart component.

clementprdhomme (commented):
I can't add this comment on the code so I'm giving you further feedback here.

The background colour of both circles should be a plain white:
https://github.com/Vizzuality/heco-invest/blob/0f17bc9c0b37e01b4d3b4610375f34c32eea027a/frontend/containers/impact-chart/component.tsx#L210-L211
You can replace them by:
```jsx
              <span className="absolute bg-white rounded-full opacity-50 top-2 right-2 bottom-2 left-2" />
              <span className="absolute scale-50 bg-white rounded-full top-2 right-2 bottom-2 left-2 opacity-60" />
```

There are further changes on the project public page (for example the chart is bigger) but I don't think we should make these right now. Let's keep the PR simple.

frontend/containers/impact-chart/component.tsx
barbara-chaves: I couldn’t make it work this way, so I used the 'reference'

clementprdhomme (approved):

PR #570, opened 2022-09-16
refactoring: enable email confirmation at API

martintomas:
PR which enforces user to confirm his email address after registration (applies only for new registered users, not invited ones). User is not able to login, before they finish email confirmation.

Users should receive email after registration with following link `#{ENV["FRONTEND_URL"]}/sign-up?confirmation_token=#{@token}`.  API used to confirm email is already implemented and documented at Rswag doc (`/api/v1/email_confirmation`). There is also possibility to re-send confirmation email via API.

@clementprdhomme @barbara-chaves  

## Testing

Use rswag documentation
 - register new user --> you should receive email which allows you to confirm your email address
 - take token from url inside email
 - and send it to `GET /api/v1/email_confirmation` and confirm
 - your user should get confirmed and you should get logged in (session is created)

## Tracking

Part of https://vizzuality.atlassian.net/browse/LET-203 -- enabling email confirmation

### DIFF
--- a/backend/app/controllers/api/v1/users_controller.rb
+++ b/backend/app/controllers/api/v1/users_controller.rb
@@ -4,10 +4,7 @@ module API
       before_action :authenticate_user!, only: [:update, :show, :change_password]
 
       def create
-        user = params[:invitation_token].present? ? preload_invited_user : User.new(create_params)
-        user.skip_confirmation!
-        user.save!
-        sign_in user
+        user = params[:invitation_token].present? ? preload_invited_user : User.create!(create_params)
         render json: UserSerializer.new(user)
       end
 
@@ -49,6 +46,8 @@ module API
 
           user.assign_attributes create_params.except(:email)
           user.skip_confirmation!
+          user.save!
+          sign_in user
         end
       end
     end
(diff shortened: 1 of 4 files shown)


clementprdhomme:
Can we not merge it (or have it disabled) until the front-end is done? The front-end team won't have capacity to work on this before the end of the sprint and thus before we deploy to production.

We also need to evaluate the front-end work to do. I know this was enabled at some point but the code has changed quite a lot since then and we need to make sure everything is in order.

Thanks for tagging us and adding some documentation!

martintomas:
> Can we not merge it (or have it disabled) until the front-end is done? The front-end team won't have capacity to work on this before the end of the sprint and thus before we deploy to production.
> 
> We also need to evaluate the front-end work to do. I know this was enabled at some point but the code has changed quite a lot since then and we need to make sure everything is in order.
> 
> Thanks for tagging us and adding some documentation!

I think disabling does not make much sense because purpose of this PR is to enable it, but we can wait with merge -- It will appropriate label to PR for now and when you are ready to work on this, feel free to merge this.

agnessa (approved):

PR #571, opened 2022-09-19
Add function to reset sort query when changing from projects tab

barbara-chaves:
This PR changes the behaviour of discover tabs changing: the sorting will always be reseted to default value

## Testing instructions

Go to discover and change the tabs. 

- The sorting should be consistent on the same tab (changing sorting order and key should work as expected)
- Te sorting key and order should change to default:
  -  (Date /  Descending) on buttons
  - sorting param on url should disappear  when changing tabs

## Tracking

[LET-1088](https://vizzuality.atlassian.net/browse/LET-1088)


clementprdhomme (approved):
All good. I've rebased your branch on `develop` and it's ready to merge now.

PR #577, opened 2022-09-21
Fix close filters suggestions on home page

barbara-chaves:
This PR fixes the homepage close autosuggestions button behaviour.

## Testing instructions

- On the home page, type some characters on the search input to open the autosuggestions
- Click on the close button 
The input should be cleared and the autosusegstion closed, keeping at the home page

## Tracking

[LET-1147](https://vizzuality.atlassian.net/browse/LET-1147)

### DIFF
--- a/frontend/containers/layouts/discover-search/component.tsx
+++ b/frontend/containers/layouts/discover-search/component.tsx
@@ -183,6 +183,7 @@ export const DiscoverSearch: FC<DiscoverSearchProps> = ({ className }) => {
             filtersData={filtersData}
             filters={filters}
             searchText={searchInputValue}
+            closeSuggestions={clearInput}
           />
         )}
         {/* Filters accordion pannel */}
--- a/frontend/containers/layouts/discover-search/component.tsx
+++ b/frontend/containers/layouts/discover-search/component.tsx
@@ -129,7 +129,7 @@ export const DiscoverSearch: FC<DiscoverSearchProps> = ({ className }) => {
                 onClick={handleClickInput}
               />
             </div>
-            {showSuggestion && (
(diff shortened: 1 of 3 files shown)


clementprdhomme:
@barbara-chaves I found another issue that I think we could include in this PR:

1. Go to Discover
2. Type “Hello” in the search bar
3. Click the suggestion

The menu should close but it does not do so.

barbara-chaves:
> @barbara-chaves I found another issue that I think we could include in this PR:
> 
> 1. Go to Discover
> 2. Type “Hello” in the search bar
> 3. Click the suggestion
> 
> The menu should close but it does not do so.

Added!

clementprdhomme:
@barbara-chaves I think we need some more work.

**Search not applied**

1. Type “Hello”
2. Click the suggestion

The projects are not filtered and the URL doesn't contain the keyword. In contrast, it I press <kbd>ENTER</kbd> after typing “Hello”, then the projects are filtered.

**Not easy to remove a keyword search**

1. Perform a keyword search with “Hello” (you can open `/en/discover/projects?search=Hello`)
2. Focus on the search input

The cross icon to remove the keyword search (on the right of the input) is only visible if I press another key. At the moment, I can only remove this search if I manually empty the input and press <kbd>ENTER</kbd>. Don't you think we could always display the cross icon if the input isn't empty?

barbara-chaves:
> The cross icon to remove the keyword search (on the right of the input) is only visible if I press another key. At the moment, I can only remove this search if I manually empty the input and press ENTER. Don't you think we could always display the cross icon if the input isn't empty?

I agree. I'm going to change it to display the "clear search" button when there is some text on the search box.

PR #593, opened 2022-09-26
Limit the selected layers to one of each group

barbara-chaves:
This PR limits the selectable map layers to one of each group 

## Testing instructions

Go to discover/projects and select map layers

- You should be able to select just one layer of each group.
- If the group already has a selected layer, it will change to the last selected value
- If you click twice on a layer option, the input should uncheck 

- Make sure that the layers legend are displayed in the correct order and if you close a layer through the legend it will unselect properly 

## Tracking

[LET-1053](https://vizzuality.atlassian.net/browse/LET-1053)

### DIFF
--- a/frontend/components/map/legend/types.d.ts
+++ b/frontend/components/map/legend/types.d.ts
@@ -17,4 +17,5 @@ export type Legend = {
   name: string;
   items: LegendItem[];
   type: LegendType;
+  group: LAYER_GROUPS;
 };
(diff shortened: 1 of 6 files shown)


clementprdhomme (commented):
Do you mind adding “(max 1 layer)” next to the group titles?

The layers menu has groups but no mention of the layers limit

I also pushed a small commit to fix a typo 😉.

clementprdhomme (approved):

PR #601, opened 2022-09-29
Let 1141 fe add zoom buttons on the discover map

barbara-chaves:
This PR adds the discover map zoom buttons and updates the search locations input button

## Description

As an HeCo visitor, I need a way to control the zoom of the map in the project search results, so that I can see the necessary level of detail of the geography, improving user experience

Acceptance criteria

There will be a way to control the zoom level of the map in the project search results

Adjust the current “find location” functionality to match the designs: [HeCo - Design System](https://www.figma.com/file/3epSYh4KQZBCyh4Y5OzmHw/HeCo---Design-System?node-id=3570%3A12388)

## Tracking

[LET-1154](https://vizzuality.atlassian.net/browse/LET-1154)
[LET-1141](https://vizzuality.atlassian.net/browse/LET-1141)


frontend/containers/discover-map/component.tsx
```diff
@@ -138,22 +150,28 @@ export const DiscoverMap: FC<DiscoverMapProps> = ({ onSelectProjectPin }) => {
...
-          <MapHelp />
         </div>
 
-        <Controls className="absolute h-fit max-h-[45%] bottom-4 right-4 overflow-y-auto">
-          <LayerLegend
-            className="bg-white"
-            onCloseLegend={(layerGroup) => resetField(layerGroup as keyof MapLayersSelectorForm)}
-            layersLegends={layerLegends}
-          />
+        <Controls className="w-full h-full">
```
clementprdhomme, line 155: The buttons have different hover and focus styles:
- when hovered, the two zoom buttons change the icon colour to dark green and don't have any border
- when focused, the map layers and help button use a ring

Can we have the same behaviour on all the buttons? It doesn't matter which one you pick.

https://user-images.githubusercontent.com/6073968/193027335-13b1d340-9fd9-4a08-8a0c-1847db6c80c6.mp4

clementprdhomme (changes requested):

frontend/containers/discover-map/component.tsx
clementprdhomme: I don't think you've addressed all the styles inconsistency yet.

**Hover**

The “Map layers” button has a different hover style:

The button has a green ring when hovered

It should look like this to match the other buttons:

The button has a green text and no ring

You can simply remove the current `hover:*` classes and add `hover:text-green-dark transition-all` instead.

**Focus**

The “Map layers” button has also a different focus style:

The button has a thick green ring when focused

To match the other buttons, it should look like this:

The button's ring is less think and offset by a few pixels

To achieve this, remove all the previous `focus-visible:*` classes and replace them by `focus-visible:outline-green-dark focus-visible:outline focus-visible:outline-2 focus-visible:outline-offset-2`.

The styles of the search component also don't match the rest:

The search component has a thick green ring when focused

I think it should look like this:

The search component has a slimmer green ring offset by a few pixels

You can achieve this by applying these classes to the `<div />` that has the `onMouseEnter` callback when it's open and focused: `outline outline-green-dark outline-2 outline-offset-2`.

clementprdhomme (approved):

PR #604, opened 2022-09-29
Improve the legibility of the SDGs in the project details card

clementprdhomme:
This PR improves the legibility of the SDGs in the project details card by:

- Displaying them bigger
- Adding a tooltip on hover

Note the tooltip is also displayed on the other pages that use the same component: public project page and public open call page.

## Testing instructions

- On desktop, 4 SDGs are displayed side-by-side in the project details card. On mobile, we display as much as possible in a row (previous behaviour).
- When you hover an SDG, a tooltip with the name of the SDG is shown.

## Tracking

[LET-1133](https://vizzuality.atlassian.net/browse/LET-1133)


barbara-chaves:
I think we could add the size=large to the open call public page too, to be consistent with the project page layout. What do you think?

barbara-chaves (approved):

clementprdhomme:
Thanks for the suggestion. Andreia agrees so I've made the change on both the open call and investor public pages.

PR #616, opened 2022-10-04
Update the login flow to account for 2 fa

barbara-chaves:
This PR adds the 2fa to sign in flow:

Adds the need of using 2fa to sign-in page
Creates the verification code page

## Testing instructions

1. Go to sign-in
2. Fill the form with any user with enabled 2fa ( [barbara.chaves+1@vizzuality.com] / 1234qwerASDF )  
3. You should be redirected to sign-in/code and receive an email with the authentication code
4. Fill the input with the code and confirm

- The inputs should accept only numbers and they are required 
- If the code is wrong, an error message will be shown

5. You should be redirected to the next screen (the example user is a light user, so you go to the accounts screen)

Try to login with a user without 2fa to make sure the flow wasn't affected

## Tracking

[1136](https://vizzuality.atlassian.net/browse/LET-1136)


frontend/pages/sign-in/index.tsx
```diff
@@ -90,130 +95,152 @@ const SignIn: PageComponent<ProjectDeveloperProps, AuthPageLayoutProps> = () =>
...
-              name="email"
-              id="email"
-              placeholder={intl.formatMessage({
-                defaultMessage: 'Insert your email',
-                id: 'ErIkUS',
-              })}
-              aria-describedby="email-error"
-              register={register}
-              className="mt-2.5"
+      {
```
clementprdhomme, line 135: Is there any need for this bracket? If so, can you add a comment?

frontend/pages/sign-in/code.tsx
```diff
@@ -0,0 +1,192 @@
...
+  return {
+    props: {
+      intlMessages: await loadI18nMessages({ locale }),
+    },
+  };
+});
+
+type ProjectDeveloperProps = InferGetStaticPropsType<typeof getStaticProps>;
+
+const CODE_LENGTH = 5;
```
clementprdhomme, line 41: The code is 6-digit long.

frontend/pages/sign-in/code.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+const SignInCode: PageComponent<ProjectDeveloperProps, NakedLayoutProps> = () => {
+  const { formatMessage } = useIntl();
+  const { replace } = useRouter();
+  const signIn = useSignIn();
+  const { user, isLoading: isUserLoading } = useMe();
+
+  // Email and password got from login page
+  const sessionData = useAppSelector(session);
```
clementprdhomme, line 50: I understand that you need to share the user's email and password between the `index.tsx` and `code.tsx` page but I wonder if using Redux is the best solution, given this state is not used anywhere else and can be discarded as soon as the user is logged in.

I also wonder if that pauses a security risk that we're storing the password in clear.

Have you thought about using some hidden URL params instead? I believe that we could send the user to `/sign-in/code?email=XXX&password=XXX` and actually hide the params so the URL would look like `/sign-in/code`. I think it can be achieved with the `as` property of Next.js' router.

There are several places where we use this technique in the code already, like here: https://github.com/Vizzuality/heco-invest/blob/63111388539ae3c2c818425626933b75dcac1f32/frontend/containers/dashboard/account-information/public-profile/component.tsx#L75-L80

frontend/pages/sign-in/code.tsx
```diff
@@ -0,0 +1,192 @@
...
+    }
+  }, [isUserLoading, replace, user]);
+
+  // Create an array of numbers from 0 to the code's length
+  const code = Array.from({ length: CODE_LENGTH }, (_, i) => i);
+
+  useEffect(() => {
+    // Set focus at the first input box
+    setFocus('otp_attempt.0');
+  }, [setFocus]);
```
clementprdhomme, line 83: Nice touch!

frontend/pages/sign-in/code.tsx
```diff
@@ -0,0 +1,192 @@
...
+    [signIn, replace]
+  );
+
+  const handlePasteCode = (ev: ClipboardEvent<HTMLInputElement>) => {
+    const text = ev.clipboardData.getData('text');
+    text
+      .split('')
+      .slice(0, CODE_LENGTH)
+      .forEach((char, index) => setValue(`otp_attempt.${index}`, char));
+  };
```
clementprdhomme, line 110: Really nice of you to have implemented this! 👏

frontend/pages/sign-in/code.tsx
```diff
@@ -0,0 +1,192 @@
...
+    const currentInputIndex = Number(name.split('.')[1]);
+    let nextInput = currentInputIndex + 1;
+    if (!value || currentInputIndex === 4) return;
+    setFocus(`otp_attempt.${nextInput}`);
+  };
+
+  const handleSignIn = useCallback(
+    (data: SignIn) => {
+      signIn.mutate(data, {
+        onSuccess: () => {
```
clementprdhomme, line 96: On the `pages/signin-in/index.tsx` file, we also have a case where the user accepts an invitation to an account. This is missing from this page. Could you add this? https://github.com/Vizzuality/heco-invest/blob/63111388539ae3c2c818425626933b75dcac1f32/frontend/pages/sign-in/index.tsx#L106-L112

frontend/pages/sign-in/code.tsx
```diff
@@ -0,0 +1,192 @@
...
+    handleSubmit,
+    setFocus,
+    setValue,
+  } = useForm<SignInCodeForm>({
+    shouldUseNativeValidation: true,
+    reValidateMode: 'onChange',
+  });
+
+  // Redirect if user is signed in
+  useEffect(() => {
```
clementprdhomme, line 64: At the moment, anybody can access this page, even if they don't come from the sign in page.

frontend/pages/sign-in/code.tsx
```diff
@@ -0,0 +1,192 @@
...
+  const onSubmit: SubmitHandler<SignInCodeForm> = (code) => {
+    const otp_attempt = Object.values(code.otp_attempt).join('');
+    handleSignIn({ ...sessionData, otp_attempt });
+  };
+
+  return (
+    <div className="flex flex-col items-center justify-center w-full h-screen max-w-xl m-auto">
+      <div>
+        <div className="mb-6 text-center">
+          <p className="mb-6 font-semibold">HeCo invest</p>
```
clementprdhomme, line 121:
```suggestion
          <p className="mb-6 font-semibold">HeCo Invest</p>
```

frontend/pages/sign-in/code.tsx
```diff
@@ -0,0 +1,192 @@
...
+            <FormattedMessage defaultMessage="Insert your code" id="Nn0ig7" />
+          </h1>
+        </div>
+
+        <form onSubmit={handleSubmit(onSubmit)} noValidate>
+          <div className="w-full max-w-md text-center">
+            <fieldset>
+              <legend className="mb-6 text-gray-800">
+                <FormattedMessage
+                  defaultMessage="To help keep your account secure, we sent a security code to your email. "
```
clementprdhomme, line 132: There's an extra unnecessary space at the end of the string.

frontend/pages/sign-in/code.tsx
```diff
@@ -0,0 +1,192 @@
...
+                  id="/vHsVH"
+                />
+              </legend>
+              <div className="flex justify-center gap-2.5">
+                {code.map((digit) => (
+                  <Input
+                    key={digit}
+                    type="text"
+                    name={`otp_attempt.${digit}`}
+                    registerOptions={{
```
clementprdhomme, line 142: Maybe you could add something like this as well?
```ts
{
  pattern: '[0-9]',
}
```

Note that `pattern` has a different type when set inside the `registerOptions` prop.

frontend/pages/sign-in/code.tsx
```diff
@@ -0,0 +1,192 @@
...
+    // Set focus at the first input box
+    setFocus('otp_attempt.0');
+  }, [setFocus]);
+
+  const handleChangeCode = (e: ChangeEvent<HTMLInputElement>) => {
+    const { name, value } = e.currentTarget;
+    const currentInputIndex = Number(name.split('.')[1]);
+    let nextInput = currentInputIndex + 1;
+    if (!value || currentInputIndex === 4) return;
+    setFocus(`otp_attempt.${nextInput}`);
```
clementprdhomme, line 90: Really nice!

frontend/pages/sign-in/code.tsx
```diff
@@ -0,0 +1,192 @@
...
+                    }}
+                    aria-describedby="code-error"
+                    register={register}
+                    className="text-center"
+                    onPaste={handlePasteCode}
+                  />
+                ))}
+              </div>
+            </fieldset>
+            <ErrorMessage
```
clementprdhomme, line 155: This error is never shown, even if I submit the form without filling the inputs. Can you check why?

frontend/pages/sign-in/code.tsx
```diff
@@ -0,0 +1,192 @@
...
+                  <Input
+                    key={digit}
+                    type="text"
+                    name={`otp_attempt.${digit}`}
+                    registerOptions={{
+                      onChange: handleChangeCode,
+                      maxLength: 1,
+                      required: true,
+                    }}
+                    aria-describedby="code-error"
```
clementprdhomme, line 147: The `<fieldset />`'s `legend` is not enough to make these inputs accessible. I think that if you add this prop, it's way clearer when using a screen reader:
```suggestion
                    aria-describedby="code-error"
                    aria-label={`Two factor authentication code, digit ${digit + 1}`}
```

frontend/store/session/index.ts
```diff
@@ -0,0 +1,28 @@
...
+import type { RootState } from 'store';
+import { SignIn } from 'types/sign-in';
+
+// Define the initial state using that type
+const initialState: SignIn = {
+  email: undefined,
+  password: undefined,
+};
+
+export const sessionSlice = createSlice({
```
clementprdhomme, line 13: If you do decide to keep using Redux, I would suggest that you add a comment above this line to explain what this slice is used for (and on which pages).

clementprdhomme (commented):
Great job @barbara-chaves on taking and implementing this complex task.

You told me that you couldn't test the whole flow because the back-end implementation wasn't merged. It actually is: https://github.com/Vizzuality/heco-invest/pull/72. You need to enable 2FA in your user settings and then you'll be able to test the whole flow (which I haven't done because there are some changes to make first).

frontend/pages/sign-in/code.tsx
```diff
@@ -0,0 +1,217 @@
...
+import NakedLayout, { NakedLayoutProps } from 'layouts/naked';
+import { PageComponent } from 'types';
+import { SignIn, SignInCodeForm } from 'types/sign-in';
+
+import { useSignIn } from 'services/authentication/authService';
+import { useAcceptInvitation, useInvitedUser } from 'services/invitation/invitationService';
+
+export const getServerSideProps = withLocalizedRequests(async (context) => {
+  // Property 'query' does not exist on type 'GetStaticPropsContext<ParsedUrlQuery, PreviewData>'
+  /** @ts-ignore */
```
clementprdhomme, line 29: `withLocalizedRequests` accepts a generic type:

```suggestion
export const getServerSideProps = withLocalizedRequests<GetServerSideProps>(async (context) => {
```

frontend/pages/sign-in/code.tsx
```diff
@@ -0,0 +1,217 @@
...
+          } else {
+            replace(Paths.Dashboard);
+          }
+        },
+      });
+    },
+    [signIn, invitedUser, acceptInvitation, query.invitation_token, replace]
+  );
+
+  const handlePasteCode = (ev: ClipboardEvent<HTMLInputElement>) => {
```
clementprdhomme, line 113: I think we're missing one instruction in this callback. There's a case where the user is not able to proceed after pasting the code:

1. Write a letter in the first input

A validation error is displayed.

2. Delete that letter
3. Paste the code in the first input

The user is not able to submit the form. They have to manually type the first digit again.

clementprdhomme (approved):

PR #624, opened 2022-10-06
Add validations folder to i18n:extract script

barbara-chaves:
This PR adds the 'validations' folder to the 'i18n:extract' script

## Testing instructions

The validation texts on files inside the validations folder should be included on the 'lang/transifex/zu.json' file

## Tracking

[1194](https://vizzuality.atlassian.net/browse/LET-1194)

### DIFF
--- a/frontend/lang/transifex/zu.json
+++ b/frontend/lang/transifex/zu.json
@@ -20,6 +20,9 @@
   "+USuwC": {
     "string": "Funding available per project"
   },
+  "+UvbBR": {
+    "string": "You need to confirm the password."
+  },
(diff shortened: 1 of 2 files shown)


clementprdhomme (changes requested):
I think you have overlooked other folders that should be included in the list. If you haven't seen it, the task says:
> Make sure to include all the top folders that may contain source strings.

If I have a quick glance at the code, I can see source strings are located inside the `schemas` folder. Can you check if other folders need to be included?

clementprdhomme (approved):

PR #634, opened 2022-10-11
Attempted fix for "Discarded TranslateJob due to a ActiveJob::DeserializationError"

agnessa:
I checked the logs for the project mentioned in the task description, which ended up not being translated. The error I found in the log was `Discarded TranslateJob due to a ActiveJob::DeserializationError`, which I think might have to do with passing an AR object as argument to the background job. While in principle that should work, I am not really sure how CloudTasks interacts with those arguments, and so to be on the safe side I think it is worth passing objects to translate by id and see if this error goes away (there are multiple instances in the logs, related to different types of translated objects, so there's nothing specifically wrong about projects I think).

## Testing instructions

I tested that translation jobs still run locally, so at least this is not worse than it was, but really it needs to be vetted on staging, as I never came across this error locally previously.

## Tracking

https://vizzuality.atlassian.net/browse/LET-1172?atlOrigin=eyJpIjoiNzBkNmRiN2QyZDkzNDE4YTllZTYxZGVkYjhhNmNjMjMiLCJwIjoiaiJ9

### DIFF
--- a/backend/app/jobs/translate_job.rb
+++ b/backend/app/jobs/translate_job.rb
@@ -1,7 +1,9 @@
 class TranslateJob < ApplicationJob
   queue_as ENV["CLOUD_TASKS_TEST_QUEUE_NAME"].to_sym
+  discard_on ActiveRecord::RecordNotFound
 
-  def perform(resource, changed_attrs: nil)
+  def perform(resource_id, resource_class, changed_attrs: nil)
+    resource = resource_class.constantize.find(resource_id)
     Translations::Translate.new(resource, changed_attrs: changed_attrs).call
   end
 end
--- a/backend/app/models/concerns/translatable.rb
+++ b/backend/app/models/concerns/translatable.rb
@@ -24,6 +24,6 @@ module Translatable
       res << attr if public_send "saved_change_to_#{attr}_#{language}?"
     end
 
-    TranslateJob.perform_later self, changed_attrs: changed_attrs if changed_attrs.present?
+    TranslateJob.perform_later id, self.class.name, changed_attrs: changed_attrs if changed_attrs.present?
   end
 end
--- a/backend/app/models/concerns/translatable.rb
+++ b/backend/app/models/concerns/translatable.rb
@@ -4,7 +4,7 @@ module Translatable
   included do
     attr_accessor :skip_translation
 
-    after_save :translate_attributes, unless: :skip_translation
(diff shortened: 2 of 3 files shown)


martintomas (approved):
I don't have any issue with your solution, but I actually think that you are trying to solve what deserialisation already solves for you. I see that there is already `discard_on ActiveJob::DeserializationError` at parent and I think it is correct. If I remember correctly, AR records are serialized by saving klass and id at queue as param and during start of job, deserialisation loads this object back from db based on saved klass and id. This is basically same stuff which you are doing with this tweak. I believe that `ActiveJob::DeserializationError` is raised when:
 - object was deleted in meantime --> we don't care to be honest
 - object was never created --> it would be definitely better to put `after_commit` here https://github.com/Vizzuality/heco-invest/blob/develop/backend/app/models/concerns/translatable.rb#L7 instead of `after_save` so we deal with rollback

Whatever, I am fine with this, but don't believe that `ActiveJob::DeserializationError` was responsible for missing description.

martintomas:
I have checked https://staging.hecoinvest.org/backend/api/v1/background_job_events?page%5Bsize%5D=70&filter%5Bcreated_at_min%5D=2022-09-26T09%3A29%3A46.389Z&filter%5Bcreated_at_max%5D=2022-09-26T10%3A29%3A46.389Z (need to be logged as Admin to see results) and from it is clear that this job (I guess it was one responsible for translation) was correctly enqueued and completed --> in second case, it is visible from log that object was correctly deserialised, because it contains attributes of object saved inside log.

agnessa:
> I have checked https://staging.hecoinvest.org/backend/api/v1/background_job_events?page%5Bsize%5D=70&filter%5Bcreated_at_min%5D=2022-09-26T09%3A29%3A46.389Z&filter%5Bcreated_at_max%5D=2022-09-26T10%3A29%3A46.389Z (need to be logged as Admin to see results) and from it is clear that this job (I guess it was one responsible for translation) was correctly enqueued and completed --> in second case, it is visible from log that object was correctly deserialised, because it contains attributes of object saved inside log.

Interesting! Still, there was an error on this job, and I suppose the logic is if a job is discarded like that, it still ends up as "completed", even if it ended up doing nothing.

Screenshot 2022-10-11 at 17 32 39

martintomas:
Yes, discarding error will for sure complete job without crash. Hmm, from your logs it truly looks like problem is deserialization and cloudtasker....

PR #640, opened 2022-10-12
Modified dynamic search scope to match on partial terms

agnessa:
Change to `pg_search` configuration to make it search by term prefixes.

## Testing instructions

Searching by word prefixes, e.g. if translatable fields contain "project", "p" should work, but not "r"
Applies to:
- projects, open calls, project developers, investors in API list endpoints (including projects map endpoint),
- same entities as above + users & admins in the backoffice.
(i.e. everywhere the `Filterer` service is used, and in all custom pg_search_scopes)

## Tracking

https://vizzuality.atlassian.net/browse/LET-1216
https://vizzuality.atlassian.net/browse/LET-1196


backend/spec/fixtures/snapshots/api/v1/user.json
```diff
@@ -1,26 +1,26 @@
...
+        "analyzed": false,
         "small": null,
         "medium": null,
-        "original": null,
-        "analyzed": false
-      },
-      "ui_language": "en",
-      "account_language": null,
-      "otp_required_for_login": false
+        "original": null
```
martintomas, line 22: I don't think it was necessary to change all these json. Could we throw away changes which are not needed? Or keep it now, but try to prevent it future? :)

backend/spec/requests/api/v1/projects_spec.rb
```diff
@@ -90,14 +91,25 @@
...
+        context "when filtered by project name" do
           let("filter[full_text]") { @project.name }
 
           it "contains only correct records" do
             expect(response_json["data"].pluck("id")).to eq([@project.id])
           end
         end
 
+        context "when filtered by partial project name" do
+          before(:each) do
```
martintomas, line 103: this is detail, but the most of the code at this repo puts `let` before `before`. Could we stick to this? :)

backend/spec/services/api/filterer_spec.rb
```diff
@@ -41,13 +41,46 @@
...
 
       context "when filtered by full_text param" do
-        let(:filters) { {full_text: "TEST", language: :en} }
-        let!(:correct_project) { create :project, name_en: "TEST" }
-        let!(:different_language_project) { create :project, name_es: "TEST" }
-        let!(:different_text_project) { create :project, name_en: "DIFFERENT" }
+        let(:filters) { {full_text: "banana", language: :en} }
+        let!(:correct_project_en) { create :project, name_en: "yellow banana", language: :en }
+        let!(:correct_project_es) {
+          create :project, name_es: "plátano amarillo", name_en: "yellow banana", language: :es
```
martintomas, line 47: You removed one use case from test with this change -- It should verify that language of filter matters. It means that `yellow banana` gets ignored if it is saved at different language (for example `name_es`) and searching is done at another language (for example `en`). Right now this test verifies that `yellow banana` needs to be there, but relation between which language filter uses and at which attribute it is saved at is lost.

backend/spec/system/backoffice/open_calls_spec.rb
```diff
@@ -50,6 +50,17 @@
         end
       end
 
+      context "when searching by partial name" do
+        it "shows only found open calls" do
+          expect(page).to have_text(open_call.name[0..2])
+          open_calls.each { |o| expect(page).to have_text(o.name) }
+          fill_in :q_name_or_investor_account_name_or_country_name_or_department_name_or_municipality_name_i_cont, with: "Ultr"
```
martintomas, line 57: Here is still `_cont` which I am not 100% but it is not only prefix searching --> should also return results if text is found at middle of word...

martintomas (approved):
Wow so many changes! :)

I don't mind boom of tests, even though I would used a bit different approach --> test all use cases at model unit tests (`search `, `dynamic_search `) and keep just one simple test at request and system tests which ensures that it is connected (ofc you could mock these methods completely and just check that they are called, but found this approach easier).

Whatever, I think it is okay ;)

backend/spec/services/api/filterer_spec.rb
agnessa: OK, I'll restore that

backend/spec/fixtures/snapshots/api/v1/user.json
agnessa: OK, I'll try to use the CONSERVATIVE snaphost update, in general specs were giving me a lot of troubles (and taking forever to run) and I guess that's why I gave up, replacing snaphots spared me at least some troubles :)

backend/spec/requests/api/v1/projects_spec.rb
agnessa: yes, I try to stick to the convention, but in this case how would you use `let` for stubbing? when there's no result to be memoized, to me `before` feels more natural

backend/spec/system/backoffice/open_calls_spec.rb
agnessa: you're right, but I think in ransack there's no equivalent of "any word starts with" matcher, so I ended up ignoring this

backend/spec/requests/api/v1/projects_spec.rb
martintomas: I am fine with using before @agnessa :), but tests at this project contain first `let` and then `before` blocks so in this case it should look like:
```
let(:filters) { {full_text: "ba", language: :en} }
let!(:correct_project_en) { create :project, name_en: "yellow banana", language: :en }
let!(:correct_project_es) {
  create :project, name_es: "plátano amarillo", name_en: "yellow banana", language: :es
}
let!(:incorrect_project) { create :project, name_en: "red apple", language: :en }

before do
  allow(Project).to receive(:translatable_attributes).and_return([:name])
end
```
but you know, it is detail. For me it is easier to read when scanning code when `let`, `before`, etc. are always at same places and there is ideally extra space between these two sections.

backend/spec/fixtures/snapshots/api/v1/user.json
martintomas: Yeah, it is slow :), which I think is quite common when using factory bot. We can either use `before_all` tweak which Tomasz uses sometimes, but it has its own cons --> when used, you need to put all variables into it which can get quite messy for files with a lot of tests. When it comes to speed, I personally prefer fixtures which are so much faster (and minitest).

backend/spec/system/backoffice/open_calls_spec.rb
martintomas: Hmm, checked https://activerecord-hackery.github.io/ransack/getting-started/search-matches/ and there are `*_start*` matchers, so maybe these would be usable in this case. It would just require to split full text sentence to words inside controller.

martintomas: Okay, this one will not work, because text at db is still un-split. Really the easiest to ignore it now :D. We could theoretically use `pg_search` instead of `ransacker` if we give up on translation fallback, but that is bigger refactoring  and probably overkill now

backend/spec/requests/api/v1/projects_spec.rb
agnessa: ah, that's what you meant! 🤦‍♀️ ☕

backend/spec/fixtures/snapshots/api/v1/user.json
agnessa: OK, I managed to reduce the number of changes to snapshots 😅

PR #642, opened 2022-10-13
Let 1157 fe tweak the project and open call public page layout

barbara-chaves:
This PR updates the open call and project public pages design

### Open calls public page

- Reduce the space between sections
- Remove “Location" title and just keep the country and state and align them horizontally
- Remove the “Funding available per project” since we have this info before there’s no need to repeat it here

Header’s white card
-  Reduce the size of text so it fits properly all labels (ticket size and instrument values) to text-xl instead of text-2xl. The -    white box should always stay with fixed width, otherwise it will hide text.
- Fix also space between the buttons, these should be fixed

**Designs:** 

[Open call with few content](https://www.figma.com/file/KC49uMR9t1YARmFR0Sr6aN/HeCo---Work-in-progress?node-id=8517%3A107785)
[Open call default](https://www.figma.com/file/KC49uMR9t1YARmFR0Sr6aN/HeCo---Work-in-progress?node-id=8517%3A107457)
Design: [here](https://www.figma.com/file/3epSYh4KQZBCyh4Y5OzmHw/HeCo---Design-System?node-id=2864%3A3446)

### Projects public page

- Reduce the size of category label to text-sm
- Add space between these two sections
- Increase the size of the left column

Header’s white card

- Reduce the size of text so it fits properly all labels (ticket size and instrument values) to text-xl instead of text-2xl. The white box should always stay with fixed width, otherwise it will hide text.
- Fix also space between the buttons, these should be fixed

Design: [here](https://www.figma.com/file/3epSYh4KQZBCyh4Y5OzmHw/HeCo---Design-System?node-id=2864%3A3446)

## Tracking

[LET-1157](https://vizzuality.atlassian.net/browse/LET-1157)
[LET-1162](https://vizzuality.atlassian.net/browse/LET-1162)

### DIFF
--- a/frontend/containers/open-call-page/header/component.tsx
+++ b/frontend/containers/open-call-page/header/component.tsx
@@ -110,8 +110,8 @@ export const OpenCallHeader: FC<OpenCallHeaderProps> = ({ openCall, instrumentTy
               <p>{openCall.description}</p>
             </div>
             <div className="mb-10 flex flex-col justify-start sm:mr-4 p-6 bg-white drop-shadow-xl sm:mb-[-70%] h-full sm:w-[395px] sm:translate-y-[-70%] sm:max-w-1/3 rounded-2xl mt-8 sm:mt-0">
-              <>
-                <div className="flex flex-col gap-8 pl-2 sm:gap-11 sm:flex-row sm:justify-between">
+              <div className="px-2">
+                <div className="flex flex-col gap-8 sm:gap-11 sm:flex-row sm:justify-between">
                   <div className="flex flex-col items-center justify-end gap-2 sm:items-start">
                     <span
                       aria-labelledby="open-call-value"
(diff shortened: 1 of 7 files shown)


frontend/containers/open-call-page/header/component.tsx
```diff
@@ -151,10 +151,16 @@ export const OpenCallHeader: FC<OpenCallHeaderProps> = ({ openCall, instrumentTy
                 <hr className="mt-6 mb-8" />
                 <div className="flex flex-col justify-between gap-8 pl-2 sm:flex-row sm:gap-11">
```
clementprdhomme, line 152: When the circle is shown, the alignment and spacing is not exactly like the design. Can you have a look?
<table>
<thead>
	<td><b>Designs</b>
	<td><b>Reality</b>
<tr>
	<td>
The deadline and circle are aligned in the middle
	<td>
The deadline and circle are aligned at the bottom

</table>

clementprdhomme (commented):
The pages look way better now!

Do you think we could take this opportunity to change the colour of the social media icons as well? It's not in the criteria but they are not very legible at the moment.

On the designs, the icons are green on a light grey background

clementprdhomme (approved):

PR #644, opened 2022-10-13
Fix a deployment failure due to the Transifex CLI installation

clementprdhomme:
Since #635 has been merged, all the deployments to staging have been failing ([example](https://console.cloud.google.com/cloud-build/builds;region=global/1c795806-2a2a-47b6-b8ed-11d0847559d6;step=0?project=heco-342801)).

At the present, when deployed, the Transifex CLI is downloaded from GitHub and decompressed at the root of the repository. [The archive](https://github.com/transifex/cli/releases/latest/download/tx-linux-amd64.tar.gz) contains a few files that may overwrite inconsequential files in our source code (such as `README.md`).

Nevertheless, it seems that a change introduced in the CLI installation script between v1.0.3 and v1.3.1 has also changed the behaviour of how the archive is decompressed: https://github.com/transifex/cli/commit/e86c5ede66b4f1069c876729f75ba37fc882ae30. This change prevents the CLI from being installed and thus prevents the deployment from proceeding.

This PR aims to download and decompress the CLI archive in a different folder.

## Testing instructions

When merged, the GitHub Actions run smoothly and the build on Google Cloud as well.

## Tracking

[LET-1195](https://vizzuality.atlassian.net/browse/LET-1195) (reopened for this issue).

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -40,7 +40,11 @@ RUN yarn install
 
 COPY . /usr/src/app
 
+# Install Transifex CLI
+WORKDIR /usr/tx-cli
 RUN curl -o- https://raw.githubusercontent.com/transifex/cli/v1.0.3/install.sh | bash
+WORKDIR /usr/src/app
+
 ENV PATH "$PATH:/usr/src/app"
 
 EXPOSE 3000
--- a/backend/lib/tasks/transifex.rake
+++ b/backend/lib/tasks/transifex.rake
@@ -2,7 +2,7 @@ namespace :transifex do
   desc "Pulling transifex translations and post processing"
   task :pull do
     puts "transifex_pull: Running TX client to pull translations"
-    IO.popen("tx pull -f -a") { |p| p.each { |line| puts line } }
+    IO.popen("/usr/tx-cli/tx pull -f -a") { |p| p.each { |line| puts line } }
     puts "transifex_pull: Post-processing"
     Dir[Rails.root.join("config/locales/{en,es,pt}.yml")].each do |file|
       puts "transifex_pull: Cleaning up #{file} from empty strings"
--- a/.github/workflows/backend-push-translations.yml
+++ b/.github/workflows/backend-push-translations.yml
@@ -18,12 +18,17 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+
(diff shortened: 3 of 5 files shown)


backend/lib/tasks/transifex.rake
```diff
@@ -2,7 +2,7 @@ namespace :transifex do
   desc "Pulling transifex translations and post processing"
   task :pull do
     puts "transifex_pull: Running TX client to pull translations"
-    IO.popen("tx pull -f -a") { |p| p.each { |line| puts line } }
+    IO.popen("/usr/tx-cli/tx pull -f -a") { |p| p.each { |line| puts line } }
```
clementprdhomme, line 5: I'm quite unsure about this change.

backend/Dockerfile
```diff
@@ -40,7 +40,11 @@ RUN yarn install
 
 COPY . /usr/src/app
 
-RUN curl -o- https://raw.githubusercontent.com/transifex/cli/v1.3.1/install.sh | bash
+# Install Transifex CLI
+WORKDIR /usr/tx-cli
+RUN curl -o- https://raw.githubusercontent.com/transifex/cli/v1.0.3/install.sh | bash
+WORKDIR /usr/src/app
+
```
tiagojsag, line 47:
```suggestion
# Install Transifex CLI
RUN mkdir /usr/tx-cli && cd /usr/tx-cli && curl -o- https://raw.githubusercontent.com/transifex/cli/v1.0.3/install.sh | bash
```

backend/lib/tasks/transifex.rake
tiagojsag: this will create a system-wide dependency, as well as some evil breaks (if, for any reason, anyone is running this on windows, it now stopped working)

tiagojsag:
I see the problem, and the solution is somewhere along this path, but it needs some polishing. I need to take a closer look at this solution, but thanks for pushing it this far, it will surely help

clementprdhomme:
@tiagojsag I've temporarily reverted the changes from #635 in #650 so that we can deploy to staging/production while we work on a long-term fix.

tiagojsag:
superseeded by https://github.com/Vizzuality/heco-invest/pull/672

PR #645, opened 2022-10-13
change useBreakPoint function

barbara-chaves:
This PR adds the **mobile** text to the beta version disclaimer

## Testing instructions

On mobile (< 768px wide), on the Dashboard, the beta disclaimer should say:
«HeCo Invest is currently on Beta version. We are still testing and making improvements and for that reason some features may not be fully functional. You are seeing a mobile version of the platform which has limited functionality, please use the desktop version to see the complete version»

## Tracking

[LET-1222](https://vizzuality.atlassian.net/browse/LET-1222)

### DIFF
--- a/frontend/containers/layouts/beta-version-disclaimer/component.tsx
+++ b/frontend/containers/layouts/beta-version-disclaimer/component.tsx
@@ -24,7 +24,7 @@ const BetaVersionDisclaimer: FC<BetaVersionDisclaimerProps> = ({ className }) =>
     localStorage.setItem('beta-version-disclaimer-closed', 'true');
   };
 
-  const isNotMobile = useBreakpoint(true)('md');
+  const isMobile = !useBreakpoint(true)('md');
 
   return (
     <div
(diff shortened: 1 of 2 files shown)


clementprdhomme (commented):
I think your solution is clever unfortunately there is a bug in `useBreakpoint` that causes the mobile message to flash on desktop:

https://user-images.githubusercontent.com/6073968/195786224-e71d5c06-f827-4733-adcd-80e30c88da7c.mp4

It seems that if `window` is not defined, `useBreakpoint` should behave as if the largest breakpoint was met, as you can read in this comment:

https://github.com/Vizzuality/heco-invest/blob/eff3e8876b0f9b66470ac37a087a29bd1123c811/frontend/hooks/use-breakpoint.ts#L37-L38

Unfortunately, the variable that holds the current breakpoint is not updated until inside the `useEffect`, which means that the initial render will not use the largest breakpoint.

https://github.com/Vizzuality/heco-invest/blob/eff3e8876b0f9b66470ac37a087a29bd1123c811/frontend/hooks/use-breakpoint.ts#L62-L69

We need that the variable's initial value is the largest breakpoint instead of `null`:

https://github.com/Vizzuality/heco-invest/blob/eff3e8876b0f9b66470ac37a087a29bd1123c811/frontend/hooks/use-breakpoint.ts#L11

There is also a small bug in the `computeBreakpoint` function. If `window` is not defined, it should not return the default breakpoint, but set `currentBreakpoint` (via `setCurrentBreakpoint`) instead. Could you make this change as well?

clementprdhomme (approved):

PR #654, opened 2022-10-18
Update project developer page to mobile version

barbara-chaves:
This PR updates the project developer public page layout for mobiles

** I added to this PR the fix of the investor page main section layout width. It's the same as the PD  

## Testing instructions

[Design](https://www.figma.com/file/KC49uMR9t1YARmFR0Sr6aN/HeCo---Work-in-progress?node-id=8546%3A108190)

## Tracking

[LET-1230](https://vizzuality.atlassian.net/browse/LET-1230)


clementprdhomme (commented):
In addition to the inline comments:
- The space between the mission and topics/sectors should be smaller (only 32px total) on mobile;
- If the space permits on mobile, the total of projects and projects waiting funding should be displayed side-by-side (this is never the case at the moment);
- On mobile, there is too much space between the investor's website and the `<hr />` element;
- Up until 640px, the breadcrumbs should be aligned with the header to have a consistency with the investor page;
- Up until 640px, there should be less space between the breadcrumbs and the green section to have a consistency with the investor page;
- On mobile, the slider's arrows of the projects section may be behind the project cards.
The slider's right arrow is behind the project card

clementprdhomme (approved):

PR #658, opened 2022-10-19
Stress tests JMeter configuration and report

agnessa:
## Testing instructions

I used JMeter 5.5 on a mac to run the tests. Results in the final report were copied from the "summary report" for the different tests.

## Tracking

https://vizzuality.atlassian.net/browse/LET-639?atlOrigin=eyJpIjoiOGVkMDE2OGU4OTU1NGZiYjljYTRjZTdlZjZmZjM3ZjAiLCJwIjoiaiJ9


tiagojsag:
So I've decided to move the load testing folder to the root of the project (it's not just a BE thing) and also remove the report and the different jmx files. IMO there's really no point in having multiple files in the repo, since they will all be the same, with the only difference being the semi-arbitrary number of users.

PR #664, opened 2022-10-20
Fix date translation

barbara-chaves:
This PR fixes the open call date translation

## Testing instructions

- Deadline on Search results 
- Deadline on Open call page 
- PD Dashboard - Open call applications → Application date
- Investor dashboard - Open call applications → Application date

## Tracking

[LET-1226](https://vizzuality.atlassian.net/browse/LET-1226)

### DIFF
--- a/frontend/pages/_app.tsx
+++ b/frontend/pages/_app.tsx
@@ -22,6 +22,9 @@ import { LayoutStaticProp } from 'types';
 
 import 'styles/globals.css';
 
+require('dayjs/locale/es');
+require('dayjs/locale/pt');
+
 // Polyfills
 import 'url-search-params-polyfill';
 
--- a/frontend/pages/_app.tsx
+++ b/frontend/pages/_app.tsx
@@ -10,6 +10,7 @@ import { useRouter } from 'next/router';
 import { I18nProvider } from '@react-aria/i18n';
 import { OverlayProvider } from '@react-aria/overlays';
 import { SSRProvider } from '@react-aria/ssr';
+import dayjs from 'dayjs';
 import { Hydrate } from 'react-query/hydration';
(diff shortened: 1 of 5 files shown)


frontend/pages/_app.tsx
```diff
@@ -22,6 +22,9 @@ import { LayoutStaticProp } from 'types';
 
 import 'styles/globals.css';
 
+require('dayjs/locale/es');
+require('dayjs/locale/pt');
```
clementprdhomme, line 26: I think we could make the code more flexible if we use the locales defined in `locales.config.json`:
```suggestion
import localesConfig from 'locales.config.json';

// Enable dayjs to display translated dates in all the application's locales
localesConfig.locales.forEach(({ locale }) => require(`dayjs/locale/${locale}.js`));
```

clementprdhomme (commented):
There are other places where dates are not displayed in the correct locale:
- https://github.com/Vizzuality/heco-invest/blob/113935607671d8cd17546fc75dc3fcaa5eb02a64/frontend/pages/dashboard/open-call/[openCallId]/application/[applicationId].tsx#L152
- https://github.com/Vizzuality/heco-invest/blob/113935607671d8cd17546fc75dc3fcaa5eb02a64/frontend/pages/dashboard/open-call-applications/[id]/index.tsx#L141

I think a more robust way to make sure we display dates correctly is to [change the dayjs' locale globally](https://day.js.org/docs/en/i18n/changing-locale). This can be achieved by adding this to `pages/_app.tsx`:
```ts
dayjs.locale(locale);
```

You could then remove all the `.locale(locale)` that you added in the PR 😉.

clementprdhomme (approved):

PR #673, opened 2022-10-25
LET-1234 Back Office - Improvements to the account approval/rejection process for accounts with invalid pictures uploaded

agnessa:
Some changes to the logic of rejecting / approving accounts in the backoffice, most importantly introduces conditional validations to the account object when rejecting. Not all validations can be made conditional, since some match constraints on the db; the rest of them should not be checked when rejecting an account either from the list or the form in the backoffice.

Validation errors should be displayed when approving an invalid object, i.e. they should be visible on the status tab.

## Testing instructions
Requires to manufacture an invalid object, e.g. directly in db.

In the list (investors / pd's):
- it should be possible to reject an invalid object
- it should not be possible to approve and an error message should be shown

In the form: (investor / pd):
- it should be indicated that picture is mandatory (asterisk)
- on the status tab, it should be possible to reject an invalid object
- on the status tab, it should not be possible to approve an invalid object and validation messaged should be shows

## Tracking

https://vizzuality.atlassian.net/browse/LET-1265
https://vizzuality.atlassian.net/browse/LET-1266
https://vizzuality.atlassian.net/browse/LET-1265

### DIFF
--- a/backend/spec/models/account_spec.rb
+++ b/backend/spec/models/account_spec.rb
@@ -81,6 +81,51 @@ RSpec.describe Account, type: :model do
     end
   end
 
+  describe "approve!" do
+    subject { create(:account, :unapproved) }
+    context "when no validation errors" do
+      it "should change status to approved" do
+        subject.approved!
+        expect(subject.review_status).to eq("approved")
+      end
+    end
+
+    context "when validation errors" do
+      before(:each) do
+        subject.assign_attributes about_en: nil, about_es: nil, about_pt: nil
+        subject.save(validate: false)
+      end
+
+      it "should not change status to approved" do
+        expect {
+          subject.approved!
+        }.to raise_error(ActiveRecord::RecordInvalid)
+      end
+    end
+  end
+
+  describe "reject!" do
+    subject { create(:account, :unapproved) }
+    context "when no validation errors" do
+      it "should change status to rejected" do
+        subject.rejected!
+        expect(subject.review_status).to eq("rejected")
+      end
+    end
+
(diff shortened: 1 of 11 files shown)


backend/spec/models/account_spec.rb
```diff
@@ -81,6 +81,49 @@
...
+        expect(subject.review_status).to eq("rejected")
+      end
+    end
+
+    context "when validation errors" do
+      before(:each) do
+        subject.assign_attributes about_en: nil, about_es: nil, about_pt: nil
+        subject.save(validate: false)
+      end
+      it "should change status to rejected" do
```
martintomas, line 120: sometimes you put extra space between `before ` and `it` block and sometimes you don't :). Definitely prefer option with extra space, again, easier to read ;)

backend/spec/system/backoffice/investors_spec.rb
```diff
@@ -97,6 +97,18 @@
...
       end
+
+      context "when validation errors" do
+        before { approved_investor.account.update_attribute(:picture, nil) }
+
+        it "doesn't change the status" do
+          within_row(unapproved_investor.name) do
+            expect(page).to have_text(ReviewStatus.find("unapproved").name)
+            click_on t("backoffice.common.reject")
+            expect(page).not_to have_text(ReviewStatus.find("approved").name)
```
martintomas, line 108: maybe it would be also worthy to check that error message is displayed to user in this case

martintomas (approved):
Looks good, just details which you can ignore or take into account :)

PR #674, opened 2022-10-25
LET-1268 Added updated_at to project and open call serializers

agnessa:
Needed on the FE.

## Testing instructions

since all project / open call endpoints use the same serializers, it should be enough to check e.g. the list endpoints

## Tracking

https://vizzuality.atlassian.net/browse/LET-1268


martintomas (approved):
Normally, I would say that you also need to rebuild rswag doc so examples there represent reality, but this is such minor change that I don't think it is required. But I would recommend to add this new attribute to entities at `swagger_helper`...

PR #675, opened 2022-10-26
LET-1209 created / updated timestamps in project and open call form

agnessa:
Timestamps were added in the project and open call forms in backoffice.

Screenshot 2022-10-26 at 08 53 25

## Tracking

https://vizzuality.atlassian.net/browse/LET-1209

### DIFF
--- a/backend/app/views/backoffice/open_calls/edit.html.erb
+++ b/backend/app/views/backoffice/open_calls/edit.html.erb
@@ -20,6 +20,10 @@
     <% end %>
   </aside>
   <div class="w-full h-full bg-white rounded-xl p-6">
+    <div class="mb-3 flex justify-end gap-3 text-s text-gray-600">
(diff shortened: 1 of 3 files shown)


backend/app/views/backoffice/open_calls/edit.html.erb
```diff
@@ -20,6 +20,10 @@
     <% end %>
   </aside>
   <div class="w-full h-full bg-white rounded-xl p-6">
+    <div class="mb-3 flex justify-end gap-3 text-s text-gray-600">
+      <span>Created: <%= I18n.l @open_call.created_at.to_date %></span>
```
martintomas, line 24: Not sure if `Created:` and `Updated:` text is so international that it does not need to be translated, but maybe it should be moved to `I18n`.

martintomas (approved):

backend/app/views/backoffice/open_calls/edit.html.erb
agnessa: of course 🙈  - thanks!

PR #681, opened 2022-10-28
Basic http auth for Rails app - backoffice only

agnessa:
## Testing instructions

If the env vars HTTP_AUTH_PASSWORD and HTTP_AUTH_USERNAME are set, you should be prompted for a login on any backoffice page, but not API, as that will be secured by http auth on the FE.

Note: the reason why I used a `before_action` rather than `http_basic_authenticate_with` is because of the usage of env vars I found it impossible to test otherwise, i.e. regardless if I used stubbing on ENV or the approach of updating ENV for the duration of the test, the ENV vars would not always be reloaded in the context of the controller class action and I gave up.

## Tracking

https://vizzuality.atlassian.net/browse/LET-1277?atlOrigin=eyJpIjoiN2E0MTdhOTg4MmJkNDI3Mjk1YTA3NTM3YTIzZGQ2ODUiLCJwIjoiaiJ9

### DIFF
--- a/backend/app/controllers/backoffice/admins/passwords_controller.rb
+++ b/backend/app/controllers/backoffice/admins/passwords_controller.rb
@@ -1,3 +1,4 @@
 class Backoffice::Admins::PasswordsController < Devise::PasswordsController
+  include Backoffice::HttpAuth
   include Backoffice::Localization
 end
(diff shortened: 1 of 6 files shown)


martintomas (approved):
Looks good, I like it ;)

PR #684, opened 2022-11-03
Improve the responsiveness of the public pages and Discover

clementprdhomme:
This PR improves the responsiveness of:

- **Breadcrumbs**: on mobile, they are always displayed on one line only and with an ellipsis if the text is too long
- **SDGs**: on mobile, they are always displayed in a 3-column layout
- **Discover**:
	- the spinning loader is now fully visible on mobile and its background is aligned with the cards
	- the general layouts have been updated to fix alignment issues

## Testing instructions

Check the responsiveness of the public pages and Discover. Make sure the scrollbars do not break the layout on Discover.

## Tracking

[LET-1287](https://vizzuality.atlassian.net/browse/LET-1287)

This PR makes changes to the Discover page that were not tracked by the task.

### DIFF
--- a/frontend/containers/breadcrumbs/component.tsx
+++ b/frontend/containers/breadcrumbs/component.tsx
@@ -104,7 +104,7 @@ export const Breadcrumbs: FC<BreadcrumbsProps> = ({
     >
       <ol
         className={cx({
-          'text-sm ': true,
+          'text-sm flex': true,
           'text-gray-400': theme === 'dark',
           'text-white': theme === 'light',
         })}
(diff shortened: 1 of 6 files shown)


SARodrigues (approved):
LGTM and it meets [LET-1287](https://vizzuality.atlassian.net/browse/LET-1287)'s requirements. 

Just a small note. While on mobile the loader's background is aligned with the cards, it looks like on desktop it became misaligned: 

Screenshot 2022-11-08 at 12 20 59

clementprdhomme:
You're right, thank you! I think I've addressed the issue now.

PR #685, opened 2022-11-03
Fix the visibility of the Discover's filters panel on Chrome (Android)

clementprdhomme:
This PR fixes an issue where the action buttons of the filters panel (Discover) wouldn't be fully visible on Chrome on Android.

## Testing instructions

- Make sure that the search bar, filters panel and active filters panel look good on all screen sizes.
- Make sure that the filters panel is not cut off at the bottom on Chrome on Android.

## Tracking

[LET-1288](https://vizzuality.atlassian.net/browse/LET-1288)


SARodrigues (approved):
Tested on Chrome/Android, LGTM!

PR #688, opened 2022-11-07
Basic auth for rswag ui

agnessa:
This took way longer than I ever hoped and took me places I never wanted to go to. Basically I believe Rswag::Ui is not flexible enough to configure it to work in our scenario, which has to do with `RAILS_RELATIVE_URL_ROOT`.  Eventually I resorted to a monkey patch. I tried to explain the reasoning in the comment. Just to recap my order of actions:

- first I just added `basic_auth_enabled` and `basic_auth_credentials` to our initialiser and expected it to work as advertised. Nothing happened - basic auth would not be triggered.
- then I digged into why, and realised that the method comparing the base path of the current request with the base path of the swagger url ends up checking whether `"api-docs" == "swagger.yaml"`
- then I digged into why, and I realised that is because we're using a relative url for the `swagger_endpoint`, which doesn't have the "api-docs" part at all
- trying to add "/api-docs" to that url - basic auth magically starts working, but everything else stops working, i.e. it cannot find the swagger file because it is now expecting a full path starting from "/backend"
- so I add "/backend" to that url - everything is working again, except for basic auth, which now checks whether `"api-docs" == "backend"` 😅 
- so I monkey patched it ☠️ 

## Tracking

https://vizzuality.atlassian.net/browse/LET-245?atlOrigin=eyJpIjoiNzhlZTI1MjgzMDg2NGM0NTk3ZjIzYjY5Y2JkNGQwODIiLCJwIjoiaiJ9

### DIFF
--- a/backend/config/initializers/rswag_ui.rb
+++ b/backend/config/initializers/rswag_ui.rb
@@ -1,5 +1,51 @@
 # frozen_string_literal: true
 
 Rswag::Ui.configure do |c|
-  c.swagger_endpoint "v1/swagger.yaml", "API V1 Docs"
+  c.swagger_endpoint "#{ENV["RAILS_RELATIVE_URL_ROOT"]}/api-docs/v1/swagger.yaml", "API V1 Docs"
+  c.basic_auth_enabled = (ENV["HTTP_AUTH_USERNAME"].present? && ENV["HTTP_AUTH_PASSWORD"].present?)
+  c.basic_auth_credentials ENV["HTTP_AUTH_USERNAME"], ENV["HTTP_AUTH_PASSWORD"]
+end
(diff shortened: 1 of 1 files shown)


martintomas (approved):
Wow, sounds like thorny journey :D. I am okay with this solution. Tbh when I was thinking about securing Rswag UI, my initial idea was to use devise authenticate at routes (https://www.rubydoc.info/github/plataformatec/devise/ActionDispatch%2FRouting%2FMapper:authenticate) and enable documentation for admins for example (or do whatever check you need there), but this solution is maybe better regarding basic auth

agnessa:
Indeed devise auth would probably have been easier!

PR #696, opened 2022-11-10
Manually keep track of changed attributes, which get wiped out in aft…

agnessa:
It's the "content not translated" bug again 🐛 , except it morphed 🐉 .

After we switched "after_save" to "after_commit", projects content gets translated OK, but account / project developer content - not at all. I finally managed to reproduce this locally and now my new working hypothesis is that our method for checking if there's anything that needs translating is not effective in an `after_commit`, because dirty state gets overwritten by multiple saves (e.g. there's a timestamp touch on account after saving project developer) and it ends up not enqueuing the job because "nothing changed".

This solution uses an instance variable to keep track of changed attributes. By the way this happens when creating / updating via controller, so model / service level specs didn't catch this - I added a request spec.

## Tracking

https://vizzuality.atlassian.net/browse/LET-1172?atlOrigin=eyJpIjoiN2JkNzUzZDQ3NDQ4NGMxMWI1NWEyODQ5MDhkOGQ4YTYiLCJwIjoiaiJ9

### DIFF
--- a/backend/app/models/concerns/translatable.rb
+++ b/backend/app/models/concerns/translatable.rb
@@ -4,6 +4,8 @@ module Translatable
   included do
     attr_accessor :skip_translation
 
+    after_save :store_changed_attributes
+    after_commit :clear_changed_attributes # this will execute last
     after_commit :translate_attributes, on: [:create, :update], unless: :skip_translation
   end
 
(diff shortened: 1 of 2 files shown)


martintomas (approved):
Great! Sounds as Rails bug to be honest, because I would expect that `saved_change_to_` would be able to keep what changed across whole saving session, but amazing that you have discovered this, had no idea about this behaviour :)

PR #705, opened 2022-11-16
LET-1302 fix issue with concurent executions of migrations on api dep…

tiagojsag:
This adds an extra step to the api deployment. It first deploys a version with a single instance and not traffic. Once that's done, it deploys the "standard" number of instances while accepting traffic. hopefully it will work, because I can't think of alternatives


agnessa (approved):
OK, let's see :) Thank you!

PR #719, opened 2023-01-05
fix: Reset impact attributes when project geometry changes

martintomas:
This PR fixes bug that allowed to set `impact_calculated` to false (impact is not calculated), but there were actually values at impact attributes. We were not reseting impact attributes between individual recalculations so if impact was calculated at past, it could contain some values even though `impact_calculated` was false. In normal situation, background job which calculates impact should change really soon `impact_calculated` back to true and this situation could go unnoticed, but for whatever reason it does not happened in this case. I have actually spend quite a lot of time to figuring out why impact calculation job failed or was never enqueued, but have not managed to find some kind of error -- not sure if google error notification still works, but there are no errors and to find something at logs without idea when exactly this happened is close to impossible.

Anyway, with this fix, everytime when impact needs to be recalculated, it also resets impact attributes so if there are values from past, they get removed.

## Testing instructions

Change geometry or project impacts and verify that project impact attributes gets reseted first (set to null) and after impact is computed, new values gets saved in it.

## Tracking

https://vizzuality.atlassian.net/browse/LET-1309


agnessa (approved):
great, the fix will ensure we have no inconsistency

PR #723, opened 2023-01-13
FE - Improve search results page performance

barbara-chaves:
This PR minimises the number of fetches and the size of fetched data on discover pages

- The first load fetches the complete data from just the current page and just one item from the other pages;
- Projects included entities fields minimised to just the used fields;
- Sorting value just change when the page can use the value;
- Filter values just change when the page can use the value;

## Tracking

https://vizzuality.atlassian.net/browse/LET-1314

### DIFF
--- a/frontend/layouts/discover-page/helpers.ts
+++ b/frontend/layouts/discover-page/helpers.ts
@@ -1,10 +1,21 @@
+import { useMemo } from 'react';
+
 import { useIntl } from 'react-intl';
 
-import { compact } from 'lodash-es';
+import { useRouter } from 'next/router';
+
+import { compact, pickBy } from 'lodash-es';
+
+import { useQueryParams } from 'helpers/pages';
 
 import { SortingOptionType } from 'components/sorting-buttons';
 import { SortingOptionKey, SortingOrderType } from 'components/sorting-buttons/types';
-import { Queries } from 'enums';
+import { Paths, Queries } from 'enums';
+
+import { useInvestorsList } from 'services/investors/investorsService';
+import { useOpenCallsList } from 'services/open-call/open-call-service';
+import { useProjectDevelopersList } from 'services/project-developers/projectDevelopersService';
+import { useProjectsList } from 'services/projects/projectService';
 
 export type SortingByTargetType = Queries.Project;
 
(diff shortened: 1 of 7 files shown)


frontend/layouts/discover-page/helpers.ts
```diff
@@ -47,3 +58,123 @@ export const useSortingByOptions = (target?: SortingByTargetType): SortingOption
...
+  const pathFilter = pickBy(filters, (_, key) => pathFilters[path].includes(key));
+
+  const params: { [key: string]: string | string[] } = { page, search, ...pathFilter };
+
+  if (pathname !== path) {
+    return {
+      ...params,
+      fields: ['name'],
+      perPage: 1,
+      'page[size]': 1,
```
clementprdhomme, line 113: Is this parameter necessary?

frontend/layouts/discover-page/helpers.ts
```diff
@@ -47,3 +58,123 @@ export const useSortingByOptions = (target?: SortingByTargetType): SortingOption
...
+    };
+  }
+
+  // Use sorting value only when applies to path to avoid unnecessary fetches
+  if (path === Paths.Projects || sorting?.includes('name') || sorting?.includes('created_at')) {
+    params.sorting = sorting;
+  }
```
clementprdhomme, line 120: I fail to see why this condition is needed. Can't we always send the `sorting` parameter?

When the user changes tab, the sorting options are reset, so we'll never send a sorting option that doesn't make sense for the current tab.

frontend/layouts/discover-page/helpers.ts
```diff
@@ -47,3 +58,123 @@ export const useSortingByOptions = (target?: SortingByTargetType): SortingOption
...
+      includes: ['project_developer', 'involved_project_developers'],
+      'fields[project_developer]': 'picture,name,contact_phone,contact_email',
+      'fields[involved_project_developers]': 'picture,name,contact_phone,contact_email',
+    };
+  }
+  if (pathname === Paths.Investors || pathname === Paths.ProjectDevelopers) {
+    return { ...params, perPage: 9 };
+  }
+  if (pathname === Paths.OpenCalls) {
+    return { ...params, includes: ['investor'] };
```
clementprdhomme, line 134: Could we limit the amount of data we load from the investor as you did for the PD and involved PD when loading projects?

frontend/pages/discover/projects.tsx
```diff
@@ -32,7 +32,7 @@ import { Project as ProjectType } from 'types/project';
 
 import { getProject } from 'services/projects/projectService';
 
-export const getServerSideProps = withLocalizedRequests(async ({ locale }) => {
+export const getStaticProps = withLocalizedRequests(async ({ locale }) => {
```
clementprdhomme, line 35: If you apply a filter and reload the page, you will see that we will first render the page with all the projects (the request to fetch the projects is sent without the filter), and then it is rendered with the filter applied (another request with the filter is sent).

This is because `useQueryParams` uses `useRouter` internally and the [Next.js' documentation](https://nextjs.org/docs/api-reference/next/router#router-object) says this:
> `query`: `Object` - The query string parsed to an object, including [dynamic route](https://nextjs.org/docs/routing/dynamic-routes) parameters. It will be an empty object during prerendering if the page doesn't use [Server-side Rendering](https://nextjs.org/docs/basic-features/data-fetching/get-server-side-props). Defaults to `{}`

To avoid `query` being an empty object, we want to keep `getServerSideProps` in this file.

frontend/layouts/discover-page/helpers.ts
```diff
@@ -47,3 +58,123 @@ export const useSortingByOptions = (target?: SortingByTargetType): SortingOption
...
 };
+
+const pathFilters = {
+  [Paths.Projects]: [
+    'filter[category]',
+    'filter[sdg]',
+    'filter[instrument_type]',
+    'filter[ticket_size]',
+    'filter[impact]',
+    'filter[only_verified]',
```
clementprdhomme, line 69:
```suggestion
    // 'filter[only_verified]', VERIFICATION FILTERS: HIDDEN
```

frontend/layouts/discover-page/helpers.ts
```diff
@@ -47,3 +58,123 @@ export const useSortingByOptions = (target?: SortingByTargetType): SortingOption
...
+    'filter[instrument_type]',
+    'filter[ticket_size]',
+    'filter[impact]',
+    'filter[full_text]',
+  ],
+  [Paths.OpenCalls]: [
+    'filter[sdg]',
+    'filter[instrument_type]',
+    'filter[ticket_size]',
+    'filter[only_verified]',
```
clementprdhomme, line 91:
```suggestion
    // 'filter[only_verified]', VERIFICATION FILTERS: HIDDEN
```

frontend/layouts/discover-page/helpers.ts
```diff
@@ -47,3 +58,123 @@ export const useSortingByOptions = (target?: SortingByTargetType): SortingOption
...
+
+const getPathParams = (
+  path: string,
+  pathname: string,
+  queryParams: { [key: string]: string | string[] }
+) => {
+  const { page, sorting, search, ...filters } = queryParams;
+
+  // Get only applicable filters for each path to avoid unnecessary fetches
+  const pathFilter = pickBy(filters, (_, key) => pathFilters[path].includes(key));
```
clementprdhomme, line 104: I'm not really a fan of the list of filters which apply to each tab above and that is used by this line.

I understand that with it, when the user applies a filter that is not available for a specific tab, we avoid a request to fetch again the number of results for this tab.

That said, we're making the front-end code aware of which filters exists and for which tab. Previously, only the back-end contained that information.This means that if tomorrow the BE enables an existing filter for tab that had no use for it, we will have to update the FE as well.

In my opinion, it would be better to make the FE as dumb as possible so as to simplify the maintenance of the code.

Please let me know if you disagree. I might have forgotten other cases in which this piece of code saves requests.

frontend/layouts/discover-page/helpers.ts
barbara-chaves: The sorting options are reset when the tab changes, but when a tab's parameters change, like sorting, it triggers fetching all contents of tabs with the current params.

currently on staging: 
Screenshot 2023-01-16 at 09 40 04

frontend/layouts/discover-page/helpers.ts
barbara-chaves: I agree with you. The requests avoided here are not delaying the render of the main content, so we can keep them for the sake of maintenance the FE code.

frontend/layouts/discover-page/helpers.ts
barbara-chaves: The last one ('page[size]': 1)?

PR #726, opened 2023-01-16
feat: Enqueue Klab jobs when projects geometry is changed and KLAB is enabled

martintomas:
This PR allows to enqueue Klab jobs when project geometry or impact areas changes. To do so, new env variable called `KLAB_ENABLED` needs to be set to true.

I am also adding all new required columns to Project model (please see migrations) and basic structure of Klab jobs -- these are more or less dummy now and I will code rest of the stuff when I start working on appropriate tickets.

## Testing instructions

Set KLAB_ENABLED to true and verify that `Klab::SubmitContextsJob` and `Klab::CalculateImpactsJob` are enqueued.

## Tracking

https://vizzuality.atlassian.net/browse/LET-1351
https://vizzuality.atlassian.net/browse/LET-1356


backend/spec/fixtures/vcr_cassettes/calculate_project_impact_score.json
```diff
@@ -0,0 +1,1688 @@
+{
```
martintomas, line 1: I have used this file to replace `backend/spec/fixtures/vcr_cassettes/export.json` which had more then 17.000 lines -- this one is 10x shorter, which is probably caused by decreasing resolution of `aries.heco.locations.colombia_continental`, which makes BC3 response a lot faster...

agnessa (approved):
looks great, but we need a new env var. if it's OK to merge this while the env var is not yet available in staging, I'll add it in a separate PR, otherwise I'd add it here - let me know!

martintomas:
> looks great, but we need a new env var. if it's OK to merge this while the env var is not yet available in staging, I'll add it in a separate PR, otherwise I'd add it here - let me know!

No need for new variable now @agnessa. `KLAB_ENABLED ` is required to turn feature on so if omit it for now, it will work same way as before. I also plan to merge all stuff for BC3 feature for some time into `klab-integration` branch so it will not interfere with staging. After we have some reasonable part of BC3 functionality done, I will push it to Stage so it can be tested at once.

PR #744, opened 2023-01-20
feat: Finalisation of klab calculate_impacts job

martintomas:
Final implementation of `Klab::CalculateImpactsJob` job which is responsible to calculate impacts based on demands obtained from k.LAB API. It can take some time to get all demands from k.LAB, therefore job checks if all demands are already available or not and waits till they are not (or maximal limit of tries run out). Default setup is now to wait almost 2h...

I have also realized that `Klab::BuildContextString` can return `nil` for some impact levels, because intersection of centroid with some location type does not have to always return some location/geometry which could be send to k.LAB. If such situation happens, demands are skipped and final impacts for such impact level are gonna be nil (our original implementation also returned nil in case there were not intersection between centroid and locations so I have tried to keep consistency here).

## Testing instructions

Pick some random project and:
 - reset all impact attributes at selected project
 - make sure that all `#{impact_level}_demands_calculated` attributes are set to `true`
 - set all demands attributes at project model to some numbers
 - enqueue `Klab::CalculateImpactsJob` and verify that after job ends, impacts are not empty at selected project anymore (impact should reflect demands which you have selected and calculation formula handled by `Impacts:: CalculateForProject ` service).

## Tracking

https://vizzuality.atlassian.net/browse/LET-1355


agnessa (approved):
OK, I ran this locally, looks good

PR #750, opened 2023-01-25
feat: Adding service for building k.LAB context string

martintomas:
Adding new service which helps to build context string send to k.LAB API via submit context API endpoint. I have tried to update Readme with new basic information how k.LAB string is constructed. Deeper implementation detail then can be found at `Klab::BuildContextString` service itself.

## Testing instructions

Find some project which have geometry (geojson) that you know how final k.LAB context string should look like. Build context string with `Klab::BuildContextString` and compare returned result with value that you have expected.

## Tracking

https://vizzuality.atlassian.net/browse/LET-1352

### DIFF
--- a/backend/app/jobs/klab/submit_contexts_job.rb
+++ b/backend/app/jobs/klab/submit_contexts_job.rb
@@ -22,7 +22,7 @@ module Klab
         context_string = Klab::BuildContextString.new(project, impact_level).call
         next skip! impact_level if context_string.blank?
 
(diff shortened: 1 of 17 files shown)


agnessa (approved):
I find the testing instructions intimidating, but the logic looks sound :)

martintomas:
> I find the testing instructions intimidating, but the logic looks sound :)

Haha, sorry for this, maybe I could mentioned there some geojson which could be used during testing

PR #768, opened 2023-03-21
LET-1369: Use observable attribute of KLAB artifact to figure out demand origin

martintomas:
Refactoring how KLAB artifact response is paired with correct requested observable. Instead of order in which observables were requested, attribute `observable` which is part of artifact response is used.

## Testing instructions

There is multiple automatic tests which ensures that logic still works (they use vcr mocks). To test it in real, `Klab::Run.new.call` can be used ;).

## Tracking

https://vizzuality.atlassian.net/browse/LET-1369


agnessa:
I updated the ARIES connection in staging, so this can be tested

agnessa (approved):

PR #769, opened 2023-03-29
Enable klab in production

agnessa:
needs running terraform apply to take effect


martintomas (approved):
I think you can enable it @agnessa even now. If I remember correctly, production is still almost not used and if klab credentials are same as at Stage, we should be fine even before Clement releases last patch -- based on info from Ferdinando, our "old" implementation should still work till they don't start doing computation in parallel.

PR #785, opened 2023-11-23
LET-1383 [BE] Change tourism-and-recreation for sustainable-tourism

yulia-bel:
This PR is BE part for replacing Tourism and recreation for Sustainable tourism (LET-1376) and includes following updates:

- Update category model and replace tourism-and-recreation with sustainable-tourism
- Migration file that will replace in existing data all the category values of tourism-and-recreation for sustainable-tourism (in tables projects, investors and project_developers)
- Update of the tests and snapshots accordingly
- Update of the swagger documentation

## Testing instructions

Actual tests are updated and cover the implemented changes (since no new functionality has been introduced).
The results of running new Migration in local database:

_before migration_

Screenshot from 2023-11-23 11-12-47
Screenshot from 2023-11-23 11-12-12
Screenshot from 2023-11-23 11-11-30

_after migration_

Screenshot from 2023-11-23 11-46-41
Screenshot from 2023-11-23 11-46-17
Screenshot from 2023-11-23 11-45-52

## Tracking

[1383](https://vizzuality.atlassian.net/browse/LET-1383)


martintomas (approved):
Looks good to me ;)

agnessa (approved):
looks good, just take into account that there are references to the old slug in the FE code. unless agreed otherwise, perhaps we could update the FE code as well not to break this?

martintomas:
> looks good, just take into account that there are references to the old slug in the FE code. unless agreed otherwise, perhaps we could update the FE code as well not to break this?

Good point, I hoped that they don't use it hardcoded anywhere 🤔 . From my point of view, it should be enough to just replace it everywhere on FE

backend/db/migrate/20231123093623_change_tourism_and_recreation_to_sustainable_tourism.rb
```diff
@@ -1,12 +1,17 @@
 class ChangeTourismAndRecreationToSustainableTourism < ActiveRecord::Migration[7.0]
   def up
     Project.where(category: "tourism-and-recreation").update_all(category: "sustainable-tourism")
+
     [Investor, ProjectDeveloper].each do |model|
-      model.find_each do |record|
+      model.where("categories @> ?", "{tourism-and-recreation}").find_each do |record|
```
martintomas, line 6: Almost there :). It should be something like this:
```
new_categories_for_all_records = model.where("categories @> ?", "{tourism-and-recreation}").pluck(:id, :categories).map do |id, categories|
  new_categories = categories.map do |category|
    category == "tourism-and-recreation" ? "sustainable-tourism" : category
  end
  {id: id, categories: new_categories}
end
model.upsert_all new_categories_for_all_records, unique_by: :id
```

Problem is that you are triggering `upsert_all` inside cycle so it gets called once per every record. It needs to go outside so you collect all data inside cycle and then outside of it call just once `upsert_all` per model.

barbara-chaves (approved):

PR #803, opened 2023-11-29
fix: admin mailer view format + update tests

yulia-bel:
Fix for previous admin notifications for new accounts creation:
- Mailer view for new project developer notification renamed to have correct format
- admin_mailer_spec.rb extended to cover project_developer_created and investor_created mailer views

## Testing instructions

admin_mailer_spec.rb extended to cover project_developer_created and investor_created mailer views

## Tracking

Link to the task(s), if any

### DIFF
--- a/backend/spec/mailers/admin_mailer_spec.rb
+++ b/backend/spec/mailers/admin_mailer_spec.rb
@@ -4,6 +4,8 @@ RSpec.describe AdminMailer, type: :mailer do
   include ActionView::Helpers::UrlHelper
 
   let(:admin) { create :admin }
+  let(:project_developer) { create :project_developer }
+  let(:investor) { create :investor }
 
   describe ".first_time_login_instructions" do
     let(:mail) { AdminMailer.first_time_login_instructions admin }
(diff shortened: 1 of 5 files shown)


agnessa (approved):
:clap:

backend/spec/mailers/admin_mailer_spec.rb
```diff
@@ -20,4 +22,34 @@
...
+    let(:mail) { AdminMailer.project_developer_created admin, project_developer }
+
+    it "renders the headers" do
+      expect(mail.subject).to eq(I18n.t("admin_mailer.project_developer_created.subject"))
+      expect(mail.to).to eq([admin.email])
+    end
+
+    it "renders the body" do
+      expect(mail.body.encoded).to match(I18n.t("admin_mailer.greetings", full_name: admin.full_name))
+      expect(mail.body.encoded).to match(I18n.t("admin_mailer.project_developer_created.content_html", new_pd_url: ""))
```
martintomas, line 36: I would recommend to test also whole url inside mailer and not just `new_pd_url: ""`. You should be able to use here `link_to('New Investor', backoffice_investor_url(investor))` -- only part which is missing to make this work is I think url helper (`backoffice_investor_url`) which is added automatically to request tests here: https://github.com/Vizzuality/heco-invest/blob/develop/backend/spec/rails_helper.rb#L41. You can modify this line and add these helpers also to mailer tests which should give you all methods you need.

martintomas (approved):

